PR #3, opened 2020-11-25
Add label properties description to spec

joshmoore:
Describe the format for label properties and provide an example.

Migrated from: https://github.com/ome/omero-ms-zarr/pull/79

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -244,6 +244,12 @@ should ignore all except the _last_ entry.
 Some implementations may represent overlapping labels by using a specially assigned
 value, for example the highest integer available in the pixel range.
 
+The `properties` key defines a list of JSON objects which also describes the unique 
+label values. Each entry in the list MUST contain the key "label-value" with the 
+pixel value for that label. Additionally, an arbitrary number of key-value pairs 
+MAY be present for each label value denoting associated metadata. Not all label 
+values must share the same key-value pairs within the properties list.
+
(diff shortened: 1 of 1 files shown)


joshmoore:
For review: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fngff.openmicroscopy.org%2F0.1%2F&doc2=http%3A%2F%2Fapi.csswg.org%2Fbikeshed%2F%3Furl%3Dhttps%3A%2F%2Fraw.githubusercontent.com%2Fome%2Fngff%2F89756cf9ccb4d7698f549931bc50a100ac6208c0%2Flatest%2Findex.bs

Note: review URL uses 0.1 rather than latest.

cc: @DragaDoncila

joshmoore:
Spec tested via:
 * https://github.com/ome/ome-zarr-py/pull/62
 * https://github.com/ome/ome-zarr-py/pull/63
 * etc.

Merging for inclusion in an 0.1.4 tag.

@DragaDoncila, FYI, our general feeling while working through the last few PRs before the I2K conference was that it wasn't possible to get everything out to the community in time for comment. ASAP we'll open an issue on image.sc, reference you and this work along with the rest of the HCS work, and basically say that if there are any breaking suggestions, we'll implement those in an upcoming 0.2. The only downside may be needing to re-generate data. Hope that's ok with you. Thanks again!

PR #16, opened 2020-11-25
Clarify "Status" section

joshmoore:
- Latest now links to 0.1 (could be made clear)
 - 0.1 now says "this version" (requires HTML rather than markdown)
 - Both still explain the upgrade policy (needs more detail)

### DIFF
--- a/0.1/index.bs
+++ b/0.1/index.bs
@@ -2,7 +2,7 @@
 Title: Next-generation file formats (NGFF)
 Shortname: ome-ngff
 Level: 1
-Status: w3c/ED
+Status: w3c/CG-FINAL
 Group: ome
 URL: https://ngff.openmicroscopy.org/0.1/
 Repository: https://github.com/ome/ngff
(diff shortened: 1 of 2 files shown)


0.1/index.bs
```diff
@@ -2,7 +2,7 @@
 Title: Next-generation file formats (NGFF)
 Shortname: ome-ngff
 Level: 1
-Status: w3c/ED
+Status: w3c/CG-FINAL
```
joshmoore, line 5: The list of permissible values is available under https://tabatkins.github.io/bikeshed/#metadata

```
ED: "Editor’s Draft"
WD: "W3C Working Draft"
FPWD: "W3C First Public Working Draft"
LCWD: "W3C Last Call Working Draft"
CR: "W3C Candidate Recommendation"
PR: "W3C Proposed Recommendation"
REC: "W3C Recommendation"
PER: "W3C Proposed Edited Recommendation"
WG-NOTE: "W3C Working Group Note"
IG-NOTE: "W3C Interest Group Note"
MO: "W3C Member-only Draft"
UD: "Unofficial Proposal Draft"
CG-DRAFT: "Draft Community Group Report"
CG-FINAL: "Final Community Group Report"
```

"REC", "CR", and "PR" all require other metadata (list of implementations) that we _may_ want to provide, but I don't think we're there yet. This renders as:

Screen Shot 2020-11-25 at 17 03 23

sbesson (approved):
Thanks for this. I think this metadata mapping faithfully captures the state of both versions of the specification.

PR #24, opened 2020-12-14
Clarify plate and well specifications for sparse plates

melissalinkert:
Starting point for discussion.  The main scenarios to clarify are plates that are missing entire row(s)/column(s), and wells with a field in some (but not all) of the defined acquisitions.

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -366,7 +366,8 @@ custom attributes of the plate group under the `plate` key.
   <dd>A list of JSON objects defining the columns of the plate. Each column
       object defines the properties of the column at the index of the object
       in the list. If not empty, it MUST contain a `name` key specifying the
-      column name.</dd>
+      column name. Each column in the physical plate MUST be defined, even
+      if no wells in the column are defined.</dd>
   <dt><strong>field_count</strong></dt>
   <dd>An integer defining the maximum number of fields per view across all
       wells.</dd>
@@ -376,7 +377,8 @@ custom attributes of the plate group under the `plate` key.
   <dd>A list of JSON objects defining the rows of the plate. Each row object
       defines the properties of the row at the index of the object in the
       list. If not empty, it MUST contain a `name` key specifying the row
-      name.</dd>
+      name. Each row in the physical plate MUST be defined, even if no wells
+      in the row are defined.</dd>
   <dt><strong>version</strong></dt>
   <dd>A string defining the version of the specification.</dd>
   <dt><strong>wells</strong></dt>
@@ -384,8 +386,8 @@ custom attributes of the plate group under the `plate` key.
       MUST contain a `path` key identifying the path to the well subgroup.</dd>
 </dl>
 
-For example the following JSON object defines a plate with two acquisition and
-6 wells (2 rows and 3 columns), containing up 2 fields of view per acquistion.
+For example the following JSON object defines a plate with two acquisitions and
+6 wells (2 rows and 3 columns), containing up to 2 fields of view per acquisition.
 
 ```json
 "plate": {
@@ -448,11 +450,102 @@ For example the following JSON object defines a plate with two acquisition and
   }
 ```
 
+The following JSON object defines a sparse plate with one acquisition and
+2 wells in a 96 well plate, containing one field of view per acquisition.
+
+```json
+"plate": {
+    "acquisitions": [
+        {
+            "id": 1,
+            "maximumfieldcount": 1,
+            "name": "single acquisition",
+            "starttime": 1343731272000
+        },
+    ],
+    "columns": [
+        {
+            "name": "1"
+        },
+        {
+            "name": "2"
+        },
+        {
+            "name": "3"
+        },
+        {
+            "name": "4"
+        },
+        {
+            "name": "5"
+        },
+        {
+            "name": "6"
+        },
+        {
+            "name": "7"
+        },
+        {
+            "name": "8"
+        },
+        {
+            "name": "9"
+        },
+        {
+            "name": "10"
+        },
+        {
+            "name": "11"
+        },
+        {
+            "name": "12"
+        }
+    ],
+    "field_count": 1,
+    "name": "sparse test",
+    "rows": [
+        {
+            "name": "A"
+        },
+        {
+            "name": "B"
+        },
+        {
+            "name": "C"
+        },
+        {
+            "name": "D"
+        },
+        {
+            "name": "E"
+        },
+        {
+            "name": "F"
+        },
+        {
+            "name": "G"
+        },
+        {
+            "name": "H"
+        }
+    ],
+    "version": "0.1",
+    "wells": [
+        {
+            "path": "C/5"
+        },
+        {
+            "path": "D/7"
+        }
+    ]
+  }
+```
+
 "well" metadata {#well-md}
 --------------------------
 
 For high-content screening datasets, the metadata about all fields of views
-under a given  well can be found under the "well" key in the attributes of the
+under a given well can be found under the "well" key in the attributes of the
 well group.
 
 <dl>
@@ -460,7 +553,7 @@ well group.
   <dd>A list of JSON objects defining the fields of views for a given well.
       Each object MUST contain a `path` key identifying the path to the
       field of view. If multiple acquisitions were performed in the plate, it  
-      SHOULD contain an `acquisition` key identifying the id of the 
+      MUST contain an `acquisition` key identifying the id of the 
       acquisition which must match one of acquisition JSON objects defined in
       the plate metadata.</dd>
   <dt><strong>version</strong></dt>
@@ -468,7 +561,7 @@ well group.
 </dl>
 
 For example the following JSON object defines a well with four fields of
-views. The first two fields of view were part of the first acquisition while
+view. The first two fields of view were part of the first acquisition while
 the last two fields of view were part of the second acquisition.
 
 ```json
@@ -493,6 +586,26 @@ the last two fields of view were part of the second acquisition.
     ],
     "version": "0.1"
   }
+```
+
+The following JSON object defines a well with two fields of view in a plate with
+four acquisitions. The first field is part of the first acquisition, and the second
+field is part of the last acquisition.
+
+```json
+"well": {
+    "images": [
+        {
+            "acquisition": 0,
+            "path": "0"
+        },
+        {
+            "acquisition": 3,
+            "path": "1"
+        }
+    ],
+    "version": "0.1"
+}
 ```
 
 Implementations {#implementations}
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -156,7 +156,7 @@ High-content screening {#hcs-layout}
 ------------------------------------
 
 The following specification defines the hierarchy for a high-content screening
-dataset. Three groups must be defined above the images:
+dataset. Three groups MUST be defined above the images:
 
 -   the group above the images defines the well and MUST implement the
     [well specification](#well-md). All images contained in a well are fields
@@ -165,6 +165,9 @@ dataset. Three groups must be defined above the images:
 -   the group above the well row defines an entire plate i.e. a two-dimensional
     collection of wells organized in rows and columns. It MUST implement the
     [plate specification](#plate-md)
+
+A well row group MUST NOT be present if there are no images in the well row.
+A well group MUST NOT be present if there are no images in the well.
 
 
 ```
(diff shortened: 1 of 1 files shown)


will-moore:
Looks good. 👍

sbesson:
The suggested changes also read fine to me and are inline with the decisions made for the first version of the HCS specification. From my side, this commit could also be ported directly to the `0.1/index.bts` specification as well.

As discussed recently, as we start  applying the OME-Zarr HCS specification to more real-world HCS use cases especially  sparse plates, we might need to review and reconsider how we handle these the specification. This can be captured and discussed as a separate issue.

melissalinkert:
0c28690 expands on the sparse plate handling to explicitly identify the row and column for each well.  https://github.com/glencoesoftware/bioformats2raw/pull/91 is a corresponding proposed implementation.

Both are based on discussion with @kkoz and @chris-allan.  In the sparse plate example where only C5 and D7 are acquired, a human reading the JSON can clearly see that ```C/5``` means C5 and ```D/7``` means D7, but the only way to automatically calculate that is to split the well path on ```/``` and match each token against ```rows``` and ```columns```.

Happy to split 0c28690 into a separate issue if that's easier to discuss.

will-moore:
There is a proposal to simplify the specifications of "collections" #31.
I assume that this will replace the existing HCS spec with something more generic.
Currently it looks like we are nearing some consensus on the overall structure of the data.
But haven't yet decided on any specific keywords for adding e.g. HCS metadata.
I'll try and come up with a suggestion, although it may initially not include plate-acquisition info.

latest/index.bs
```diff
@@ -376,16 +377,20 @@ custom attributes of the plate group under the `plate` key.
...
-      name.</dd>
+      name. Each row in the physical plate MUST be defined, even if no wells
+      in the row are defined.</dd>
   <dt><strong>version</strong></dt>
   <dd>A string defining the version of the specification.</dd>
   <dt><strong>wells</strong></dt>
   <dd>A list of JSON objects defining the wells of the plate. Each well object 
-      MUST contain a `path` key identifying the path to the well subgroup.</dd>
+      MUST contain a `path` key identifying the path to the well subgroup.
+      Each well object MUST contain both a `row_index` key identifying the index into
```
sbesson, line 387: Regarding https://github.com/ome/ngff/pull/24#issuecomment-801297872, are there cases where it is not possible to recompute these indexes based on the knowledge of the individual wells `path` as well as the `rows` and `names` dictionaries? If recomputing is always possible (but at the cost of the consumer), my primary consideration is whether the recommendation for these  new fields should be `SHOULD` rather than `MUST`.

For real-world examples, I can definitely see how `row_index/column_index` makes sense in terms of optimizing some of the queries. In addition to testing this with sparse plates, it will be useful to also generate representative plate with many wells (384 at least) to check there is no performance impact with the extra JSON metadata.

chris-allan: In order for these indexes to be forward or reverse computable, `path` would need to be much more explicitly defined than it is now:

> A list of JSON objects defining the fields of views for a given well. Each object MUST contain a path key identifying the path to the field of view. If multiple acquisitions were performed in the plate, it SHOULD contain an acquisition key identifying the id of the acquisition which must match one of acquisition JSON objects defined in the plate metadata.

Furthermore, the `wells` array would need have be null or similar padding in order for those indexes to make sense.

Neither of these things are ideal obviously. I don't think there's a way to not have these things be `MUST` if we want to guarantee that lookups can happen based on physical plate characteristics.

melissalinkert:
5a1ddc7 is based on https://github.com/glencoesoftware/bioformats2raw/pull/119 and discussion with @chris-allan earlier today, in preparation for discussion with @sbesson tomorrow. The proposed changes around well path in particular are still up for debate.

latest/index.bs
```diff
@@ -375,17 +379,24 @@ custom attributes of the plate group under the `plate` key.
...
   <dt><strong>wells</strong></dt>
   <dd>A list of JSON objects defining the wells of the plate. Each well object 
-      MUST contain a `path` key identifying the path to the well subgroup.</dd>
+      MUST contain a `path` key identifying the path to the well subgroup.
+      The `path` MUST consist of a `name` in the `rows` list, a file separator (`/`),
+      and a `name` from the `columns` list, in that order. The `path` MUST NOT contain
+      additional leading or trailing directories.
+      Each well object MUST contain both a `row_index` key identifying the index into
+      the `rows` list and a `column_index` key indentifying the index into
+      the `columns` list. `row_index` and `column_index` MUST be 0-based.</dd>
```
will-moore, line 395: I realise that https://github.com/ome/ngff/pull/70 has been added after this PR was opened, but the decision there means these new attributes should now be named `rowIndex` and `columnIndex`.

melissalinkert: Should be fixed in 7c2536a.

melissalinkert:
Following discussion with @sbesson and @chris-allan, 3c31c14 relaxes the "no empty groups" statement to address https://github.com/ome/ngff/pull/24#commitcomment-61294088. There are also some clarifications to the row and column naming, intended to be consistent with https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#NamingConvention.

sbesson:
## Overview 

The specification changes proposed in this PR closely reflect several  choices made in `omero-cli-zarr` when implementing the first version of the HCS metadata. Effectively, this migrates implementation details at the specification level clearing several ambiguities when dealing with sparse plates. The advantages are:
1. reduce the divergence between writing implementations
2. clarify the expectations for consumer when dealing with NGFF datasets implementing the HCS specification

## RFC and Community call

This PR has now passed several rounds of internal review and is reaching the state where community feedback would be useful before integrating formally in an upcoming version of the specification. Given the [latest announcement of the next NGFF call](https://forum.image.sc/t/next-call-on-next-gen-bioimaging-data-tools-2022-01-27/60885), I would propose to set the week of 2022-01-24 as the deadline for public comments. Ideally, we can review the state of this proposal, reach an agreement and decide on the timeline for getting these changes released as part of this community call.

## Specific comments

### Empty Zarr groups

A former version of the proposal forbade the existence of Zarr groups for wells and well rows containing no images. Following the feedback from https://github.com/ome/ngff/pull/24#commitcomment-61294088, the latest version now reduces this as a recommendation. I can think of rationales backing both specification. Importantly, the biggest decision factor might be at the level of the consumer library:
- assuming a strict proposal (MUST NOT), a library can either use the existence of Zarr group or the `wells` metadata to determine whether wells are populated with images
- assuming the more lenient proposal (SHOULD NOT), a library cannot rely on the exisence of Zarr groups. Instead, the `wells` field acts as the single source of truth for whether wells are populated.

### Rows/columns names

The new requirements regarding the content of the `rows` and `columns` arrays  allow to communicate a representation of the physical plate layout independently of whether wells are populated or not.

Constraints have been added to the `name` definition of rows and columns. This should make broken scenarios like duplicate row/column names invalid as per the specification. Additionally, these constraints support the ubiquitous convention in the High-Content Screening domain of using letters/numbers for rows/columns e.g. row A, column 2 while still catering for some flexibility in the naming of rows/columns.

### Wells indices

A major change in this proposal is that each `well` element now requires three keys: a `path` *AND* a `rowIndex` *AND* a `columnIndex`. The first element is unchanged compared to the previous specs and specifies the path to the Zarr group. The two indices allow to link this group to the associated row/column in the `plate` metadata.

The  examples in the specification page as well as the `bioformats2raw` and `omero-cli-zarr` implementations use systematic naming conventions where the path to the well is derived from the names of the corresponding row and the column e.g. the well corresponding to row `A`(columnIndex: `0`) and column `2` (columnIndex: `1`) is located in path `A/2`. This representation has obviously readability advantageous but this behavior is *not* enforced by the current proposal i.e. libraries should assume that the `path` to individual wells is independent of the row/column names.

As discussed in https://github.com/ome/ngff/pull/24#discussion_r597699471, an alternate proposal would be to force a mapping between the path to the Zarr group of the well and the names of the row and column associated with this well. Under such a proposal, it would become superfluous to require both the `path` and `rowIndex/columnIndex` attributes in the `wells` array as one could be recomputed from the other. 
Probably the biggest trade-offs up for discussion here are:
- flexibility e.g. support for well paths of type `0/0`
- size of the plate metadata e.g. for 1536 wells plate
- performance i.e. cost of the name <-> index lookup

### Well metadata

At the well group level, for multi-acquisition plates, the `acquisition` key is now a mandatory key. An alternative would be to define some default behavior if this field is absent e.g. the first element of the top-level `acquisitions` array. Multi-acquisition plates are the exception rather the norm but even in these scenarios, this change sounds completely reasonable to me

## Samples and implementations

The specification includes a few examples of metadata for sparse HCS data that complements the existing examples of dense plates. As for every release, representative real-world HCS examples should be generated covering as many features as possible. 

In terms of implementation, https://github.com/glencoesoftware/bioformats2raw/pull/119 contains the implementation of these changes for `bioformats2raw`, changes are expected to `omero-cli-zarr` to support the new attributes. Possibly consuming libraries like `vizarr` could be updated to benefit from the index lookup.

sbesson:
Another comment while looking at validation this morning is that the the current specification does not define the level of requirement for the keys under `plate` and `well`.

The initial JSON schema introduced in https://github.com/ome/ngff/pull/76/files#diff-2e387106f2f394aca19236f21c170f70b94c7931f60bfc2d8f6549941105e0cfR105 defines `version`, `rows`, `columns` and `wells` as required for `plate`. This would leave `acquisitions`, `field_count` and `name` as optional. This is largely in-line with the spirit of the changes here with possibly a discussion around `version` as this key is marked as recommended in the other specifications. That being said, I am personally in favor of enforcing `version` as a requirement everywhere in the mid-term.

For `well`, I assume `images` is required and `version` is either required or recommended, aligning with the decision regarding plates.

latest/index.bs
```diff
@@ -425,17 +433,30 @@ custom attributes of the plate group under the `plate` key.
...
+      and `aA`).</dd>
   <dt><strong>version</strong></dt>
   <dd>A string defining the version of the specification.</dd>
   <dt><strong>wells</strong></dt>
   <dd>A list of JSON objects defining the wells of the plate. Each well object
-      MUST contain a `path` key identifying the path to the well subgroup.</dd>
+      MUST contain a `path` key identifying the path to the well subgroup.
+      The `path` MUST consist of a `name` in the `rows` list, a file separator (`/`),
+      and a `name` from the `columns` list, in that order. The `path` MUST NOT contain
+      additional leading or trailing directories.
```
sbesson, line 450: It just occurs to me that this reads as if the  names for the well row Zarr group and the well Zarr group must be defined in the rows and columns lists. 

My understanding is that the intent was not to enforce this requirement and this was one of the basis for making column/row indices mandatory.

will-moore: Currently `ome-zarr-py` doesn't even read the `path` value for each Well, but assumes the path is `row/column` for every Well:

https://github.com/ome/ome-zarr-py/blob/13b84281732a06aa714648a86d070c6d9e7a40c9/ome_zarr/reader.py#L537

It's assumed the plate is not sparse, but we read the tiles in a try/catch that handles missing Wells.

melissalinkert: Probably worth a short discussion. Row and column names are already explicitly required here (added in https://github.com/ome/ngff/pull/24/commits/5a1ddc7d57d3c7f96e1d84efdb7a86e701666de6). I don't know how we could allow the names to be optional, but still enforce ```row/column``` as the well path.

will-moore: In vizarr I do the opposite: start with `path` and work out the row and column from that:
https://github.com/hms-dbmi/vizarr/blob/3f2149727923eb38340d8f6590b20ccf93eebc9f/src/ome.ts#L169

In both cases, I am managing without the new `rowIndex` and `columnIndex` because of this rule (which wasn't in the spec before)? With this PR, you could imagine that the `path` could be anything (if you look it up for every Well), allowing you to store the Well images in a flexibly way, using whatever `path` you want.

I'm not too sure of why we need `rowIndex` and `columnIndex` to be defined (instead of calculating them), but with them, I think a file would be invalid if the `path` wasn't equal to `rows[rowIndex]/columns[colIndex]` because you wouldn't know whether to use that or the provided `path`?

sbesson: It looks like I misremembered the set of decisions made in https://github.com/ome/ngff/pull/24#issuecomment-990082917. https://github.com/ome/ome-zarr-py/pull/159 was adjusted to match the specification as defined and additional checks are added to validate the content of well paths against `rows`, `rowIndex`, `columns`, `columnIndex`

sbesson (approved):
After reviewing the discussions from the 6th OME-NGFF community call, no objection/amendment was made for this proposal. The various HCS aware OME-Zarr implementations have been updated to support the new proposed layout. Merging in preparation of the upcoming 0.4 specification announcement.

Issue #35, opened 2021-02-28
Data dimensionality and axes metadata


constantinpape:
In last weeks meeting the question of data dimensionality came up again (in the morning it was raised by @jni, and I think it came up in the afternoon as well).
Currently, the spec demands that all data is 5 dimensional (I think with axis order TCXYZ, but I am not quite sure).

Do we want to lift the restriction and allow data of lower dimensionality? In this case, we would add metadata in `multiscales` to describe the axes (e.g. `"axes": ["x", "y", "z"]`).

Note that this is also important for the transformation spec #28, where we need to clarify which axes a transformation applies to.

Independent of the decisions, we should add a field that describes the physical units of the axes, e.g. `"units": ["micrometer", "micrometer", "micrometer"]`.

jni:
> (I think with axis order TCXYZ, but I am not quite sure).

TCZYX. ;)

> Do we want to lift the restriction and allow data of lower dimensionality?

Yes please! And in fact the axis names should be *whatever*, I we should not be limited to subsets of "TCZYX". eg could be ["lat", "lon"] or ["left-right", "superior-inferior", "anterior-posterior"].

joshmoore:
@jni: what behavior would you expect for an array with no x or y?

tischi:
Based on how this discussion evolved: https://github.com/ome/ngff/issues/28 I guess the axis names may be part of the specification of the transformation from data space to physical space, is it?

tischi:
> what behavior would you expect for an array with no x or y?

@joshmoore  What do you mean by "behavior", maybe "how it would be rendered in a viewer"?

d-v-b:
> @jni: what behavior would you expect for an array with no x or y?

In my opinion, a generic image viewer should have no intrinsic opinion about the particular axis names of the data it displays. If the user has 2D data with axes labelled X and B, then the viewer should display the data (with a default, but overrideable, mapping from data coordinates to viewer coordinates) as an image with one axis labelled "X" and the other axis labelled "B". If the data axis labelled "X" happens to be mapped to a display axis also called "X", then that is just a happy coincidence. A general-purpose data visualization tool should not assign any "meaning" to an axis name like "X" or "T". A more specialized tool might have an opinion about axis names, though.

tischi:
> then the viewer should display the data (with a default, but overrideable, mapping from data coordinates to viewer coordinates)

[...]

Issue #28, opened 2021-02-23
Transformation Specification


constantinpape:
This is a feature request towards v0.2 of the multi-scale image spec:
Support (affine) transformations for  images to apply these on the fly (e.g. to map images to a shared coordinate system).

So far, I have come around two different implementations that contain a proposal for this:

**[Open Organelle](https://openorganelle.janelia.org/)**
```json
{
"datasets": [
      {
        "path": "s0",
        "transform": {
          "axes": ["z", "y", "x"],
          "scale": [1, 1, 1],
          "translate": [0, 0, 0],
          "units": ["nm", "nm", "nm"]
         }
       }
   ]
} 
```
The transformation metadata is also duplicated in the individual datasets.

**[I2K Tutorial](https://github.com/tischi/i2k-2020-s3-zarr-workshop)**

Same as the open organelle proposal, but the transformation is defined for on the dataset level, not for each individual scale:
```
{
"datasets": [...],
"transform": {...}
}
```
The transformation metadata is not duplicated.

(I have left out the nested `multiscale` etc. in both cases to make this more readable.)

I would be happy to develop this into a PR to the spec, but it would be good to have some feedback first:
- Is there a competing approach we should take into account? Any general critique?
- Should the transformation be defined per dataset or per scale level (I think per scale level actually makes more sense).
- Do we add `rotation` and `shear` to `transformation`, so that a full affine trafo can be specified? 

@joshmoore @frauzufall @kephale @axtimwalde @d-v-b @tischi @jni
(please tag anyone else who might be interested in this)

----

Original text from #12:

Originally discussed as part of the [multiscales specification](https://github.com/zarr-developers/zarr-specs/issues/50), scale metadata related to each volume in a multiscale image needs storing for the proper interpretation of the data.

see also https://github.com/ome/omero-iviewer/pull/359 in that each level should be strictly smaller than the last.

jni:
Note that there is a scale "discussion" ;) in #12.

Some considerations:
- "axes" appears to be the same as "dimension_order", or at least intimately related, as discussed during the meeting. This is probably not the right place for this information?
- In napari, we're now finding that many users want to specify scale/translate in one place, and an affine transformation *in addition* to this. I thought coming in to this issue that it was mostly about the latter bit...?
[...]

Issue #29, opened 2021-02-23
Hierarchical chunk storage


constantinpape:
Currently, the ome zarr specification demands that the chunks are stored in a single directory.
This causes issues on the file system for large images / volumes with many chunks.

N5 stores chunks in a hierarchy natively and zarr also supports this with the [NestedDirectoryStore](https://github.com/zarr-developers/zarr-python/blob/760bf190468ad867a0fc733fb10db1362616df3d/zarr/storage.py#L1179) (although this is currently not working according to @joshmoore).

joshmoore:
Here's what I've been working on:
 * n5 supports nested (of course) but zarr+s3 is missing: https://github.com/saalfeldlab/n5-zarr/pull/5
 * jzarr doesn't support nested: https://github.com/bcdev/jzarr/pull/17
 * NestedDirectoryStore supports zarr+nesting but the newer FSStore did not: https://github.com/zarr-developers/zarr-python/pull/699

joshmoore:
Status update:
 * bioformats2raw is moving towards 0.3.0. There will already be some breaking layout changes.
 * zarr-python has the FSStore fixe merged to the mainline. The 2.7.0 release will contain it.
 * The jzarr work will be taken over by @SabineEmbacher but is moving forward.

With all of this in mind, I'd propose that we go ahead and bump to OME-Zarr version to 0.2.0 stating in the spec that nested storage is a MUST. A client can then detect whether the fileset is 0.1.0 or 0.2.0 from the metadata and make a choice as to the chunk index separator.

cc: @manzt (since I've failed to engage on the JS front)

joshmoore:
see: https://github.com/zarr-developers/zarr-python/pull/715

joshmoore:
Feels like most implementations have now implemented this new strategy. Likely a solid next step would be to add v0.1 and v0.2 implementations to https://github.com/ome/ome_zarr_test_suite

PR #39, opened 2021-03-13
Lift 5d requirement for images and move multiscales description into spec

constantinpape:
As discussed in #35 this lifts the restriction for images to be 5-dimensional.

Also, I have added a description of the multiscales metadata based on https://forum.image.sc/t/multiscale-arrays-v0-1/37930.
This will allow us to keep track of changes to the metadata better.

Note that I have added the field "axes" in order to specify the names of axes in the dataset; this becomes necessary due to 
images not having fixed dimension any more. Note that there is still an ongoing discussion about the allowed names for axes (and also about allowing more than 5 dimensions) in #35 and #28. I have not addressed these issues yet, so for now the names are restricted to "t", "c", "x", "y", "z". Also, this is a SHOULD requirement to not break compatibility with 0.1.

@joshmoore 
- I am still refering to the version with 0.1, should it be bumped due to the changes w.r.t. number of dims?
- Do I need to [render the changes locally](https://w3c-ccg.github.io/bikeshed_instructions.html) or is there some CI that also renders PRs? (Would be very convenient :))

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -212,16 +212,16 @@ found under the "multiscales" key in the group-level metadata.
 "multiscales" MUST contain the field "datasets", which is a list of dictionaries describing
 the arrays storing the individual resolution levels. Each dictionary in "datasets" MUST
 contain the field "path", whose value contains the path to the array for this resolution relative
-to the root zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
+to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
+
+It MUST contain the field "axes", which is a list of dimension names of the axes.
+The values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
+The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
 
 It SHOULD contain the field "name".
 
 It SHOULD contain the field "version", which indicates the version of the
-multiscale metadata of this image (current version is 0.1).
-
-It SHOULD contain the field "axes", which is a list of dimension names of the axes.
-If given, the values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
-The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
+multiscale metadata of this image (current version is 0.2).
 
 It SHOULD contain the field "type", which gives the type of downscaling method used to generate the multiscale image pyramid.
 It SHOULD contain the field "metadata", which contains a dictionary with additional information about the downscaling method.
@@ -230,7 +230,7 @@ It SHOULD contain the field "metadata", which contains a dictionary with additio
 {
     "multiscales": [
         {
-            "version": "0.1",
+            "version": "0.2",
             "name": "example",
             "datasets": [
                 {"path": "0"},
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -124,8 +124,8 @@ multiple levels of resolutions and optionally associated labels.
     ├── n                     # The name of the array is arbitrary with the ordering defined by
     │   │                     # by the "multiscales" metadata, but is often a sequence starting at 0.
     │   │
-    │   ├── .zarray           # All image arrays are 5-dimensional
-    │   │                     # with dimension order (t, c, z, y, x).
+    │   ├── .zarray           # All image arrays are up to 5-dimensional with dimension order (t, c, z, y, x).
+    │   │                     # An image without time dimension has axes (c, z, y, x) etc.
     │   │
     │   ├── 0.0.0.0.0         # Chunks are stored with the flat directory layout.
     │   │   ...               # Each dotted component of the chunk file represents
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -207,8 +207,52 @@ keys as specified below for discovering certain types of data, especially images
 
 Metadata about the multiple resolution representations of the image can be
 found under the "multiscales" key in the group-level metadata.
-The specification for the multiscale (i.e. "resolution") metadata is provided
-in [zarr-specs#50](https://github.com/zarr-developers/zarr-specs/issues/50).
+"multiscales" contains a list of dictionaries where each entry describes a multiscale image.
+
+"multiscales" MUST contain the field "datasets", which is a list of dictionaries describing
+the arrays storing the individual resolution levels. Each dictionary in "datasets" MUST
+contain the field "path", whose value contains the path to the array for this resolution relative
+to the root zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
+
+It SHOULD contain the field "name".
+
+It SHOULD contain the field "version", which indicates the version of the
+multiscale metadata of this image (current version is 0.1).
+
+It SHOULD contain the field "axes", which is a list of dimension names of the axes.
+If given, the values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
+The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
+
+It SHOULD contain the field "type", which gives the type of downscaling method used to generate the multiscale image pyramid.
+It SHOULD contain the field "metadata", which contains a dictionary with additional information about the downscaling method.
+
+```json
+{
+    "multiscales": [
+        {
+            "version": "0.1",
+            "name": "example",
+            "datasets": [
+                {"path": "0"},
+                {"path": "1"},
+                {"path": "2"}
+            ],
+            "axes": [
+                "t", "c", "z", "y", "z"
+            ],
+            "type": "gaussian",
+            "metadata": {                                       # the fields in metadata depend on the downscaling implementation
+                "method": "skimage.transform.pyramid_gaussian", # here, the paramters passed to the skimage function are given
+                "version": "0.16.1",
+                "args": "[true]",
+                "kwargs": {"multichannel": true}
+            }
+        }
+    ]
+}
+```
+
+
 If only one multiscale is provided, use it. Otherwise, the user can choose by
 name, using the first multiscale as a fallback:
 
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -209,9 +209,9 @@ Metadata about the multiple resolution representations of the image can be
 found under the "multiscales" key in the group-level metadata.
 "multiscales" contains a list of dictionaries where each entry describes a multiscale image.
 
-"multiscales" MUST contain the field "datasets", which is a list of dictionaries describing
-the arrays storing the individual resolution levels. Each dictionary in "datasets" MUST
-contain the field "path", whose value contains the path to the array for this resolution relative
+Each "multiscales" dictionary MUST contain the field "datasets", which is a list of dictionaries describing
+the arrays storing the individual resolution levels.
+Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 
 It MUST contain the field "axes", which is a list of dimension names of the axes.
@@ -238,7 +238,7 @@ It SHOULD contain the field "metadata", which contains a dictionary with additio
                 {"path": "2"}
             ],
             "axes": [
-                "t", "c", "z", "y", "z"
+                "t", "c", "z", "y", "x"
             ],
             "type": "gaussian",
             "metadata": {                                       # the fields in metadata depend on the downscaling implementation


joshmoore:
> * is there some CI that also renders PRs? (Would be very convenient :))

I haven't build it into the GHA workflow(s) yet (*wink, wink*) but:

 * http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/ome/ngff/9fae8451eba2e3dbec45fe22c4a74bc66750831e/latest/index.bs produces the HTML that merging this PR would generate
 * while https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fngff.openmicroscopy.org%2Flatest%2F&doc2=http%3A%2F%2Fapi.csswg.org%2Fbikeshed%2F%3Furl%3Dhttps%3A%2F%2Fraw.githubusercontent.com%2Fome%2Fngff%2F9fae8451eba2e3dbec45fe22c4a74bc66750831e%2Flatest%2Findex.bs even produces a diff of the existing latest/index.html to the output of this PR

latest/index.bs
```diff
@@ -207,8 +207,52 @@ keys as specified below for discovering certain types of data, especially images
...
+the arrays storing the individual resolution levels. Each dictionary in "datasets" MUST
+contain the field "path", whose value contains the path to the array for this resolution relative
+to the root zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
+
+It SHOULD contain the field "name".
+
+It SHOULD contain the field "version", which indicates the version of the
+multiscale metadata of this image (current version is 0.1).
+
+It SHOULD contain the field "axes", which is a list of dimension names of the axes.
```
sbesson, line 222: What is the default behavior if `axes` is not specified? In particular should the consumer have an expectation with regards to the number of underlying dimensions or try to detect it?

Alternatively, the lifting of the 5D requirement could lead to increasing the requirement for `axes` at a `MUST` level.

latest/index.bs
```diff
@@ -207,8 +207,52 @@ keys as specified below for discovering certain types of data, especially images
...
+            ],
+            "axes": [
+                "t", "c", "z", "y", "z"
+            ],
+            "type": "gaussian",
+            "metadata": {                                       # the fields in metadata depend on the downscaling implementation
+                "method": "skimage.transform.pyramid_gaussian", # here, the paramters passed to the skimage function are given
+                "version": "0.16.1",
+                "args": "[true]",
+                "kwargs": {"multichannel": true}
```
sbesson, line 248: Side note probably outside the original scope of this PR, this example was copied from https://github.com/zarr-developers/zarr-specs/issues/50 and reads very Python-specific in a specification that is expected to be increasingly language-agnostic.

latest/index.bs
```diff
@@ -207,8 +207,52 @@ keys as specified below for discovering certain types of data, especially images
 
 Metadata about the multiple resolution representations of the image can be
 found under the "multiscales" key in the group-level metadata.
-The specification for the multiscale (i.e. "resolution") metadata is provided
-in [zarr-specs#50](https://github.com/zarr-developers/zarr-specs/issues/50).
+"multiscales" contains a list of dictionaries where each entry describes a multiscale image.
```
sbesson, line 210: :+1: for embedding the multiscales convention explicitly in this specification from my side

sbesson (commented):
Thanks @constantinpape.  A few initial questions largely revolving around the contract for the `axes` metadata.

latest/index.bs
constantinpape: Yes, maybe it's better to change `axes` to `MUST`. I went for SHOULD to stay compatible with 0.1; but if hierarchical storage becomes `MUST` (see #29) 0.1 will be deprecated anyway and we can also promote `axes` to `MUST`.

latest/index.bs
constantinpape: Yes, I was also wondering about this when I copied from the example.

latest/index.bs
constantinpape: Any thoughts @joshmoore?

constantinpape:
> Yes. You can bump to an 0.2.x but if we go through with [#29 (comment)](https://github.com/ome/ngff/issues/29#issuecomment-797420293) it'll likely become an 0.3.x.

Ok, I will bump to 0.2 then. And looking forward to having hierarchies :).


> I haven't build it into the GHA workflow(s) yet (_wink, wink_) but:
> 
>     * http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/ome/ngff/9fae8451eba2e3dbec45fe22c4a74bc66750831e/latest/index.bs produces the HTML that merging this PR would generate
> 
>     * while https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fngff.openmicroscopy.org%2Flatest%2F&doc2=http%3A%2F%2Fapi.csswg.org%2Fbikeshed%2F%3Furl%3Dhttps%3A%2F%2Fraw.githubusercontent.com%2Fome%2Fngff%2F9fae8451eba2e3dbec45fe22c4a74bc66750831e%2Flatest%2Findex.bs even produces a diff of the existing latest/index.html to the output of this PR

Cool, I will check it out!

latest/index.bs
joshmoore: I'd leave the python example _somewhere_, but we can also add a Java one, e.g. from bioformats2raw.

latest/index.bs
```diff
@@ -207,8 +207,52 @@ keys as specified below for discovering certain types of data, especially images
...
+
+"multiscales" MUST contain the field "datasets", which is a list of dictionaries describing
+the arrays storing the individual resolution levels. Each dictionary in "datasets" MUST
+contain the field "path", whose value contains the path to the array for this resolution relative
+to the root zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
```
joshmoore, line 215: I would have thought it's relative to the current zarr group (containing the multiscales) as opposed to the root zarr group.

latest/index.bs
```diff
@@ -124,8 +124,8 @@ multiple levels of resolutions and optionally associated labels.
     ├── n                     # The name of the array is arbitrary with the ordering defined by
     │   │                     # by the "multiscales" metadata, but is often a sequence starting at 0.
     │   │
-    │   ├── .zarray           # All image arrays are 5-dimensional
-    │   │                     # with dimension order (t, c, z, y, x).
+    │   ├── .zarray           # All image arrays are up to 5-dimensional with dimension order (t, c, z, y, x).
+    │   │                     # An image without time dimension has axes (c, z, y, x) etc.
```
joshmoore, line 128: Is a 3D array interpreted as `(z, y, x)`?

joshmoore (commented):
Few thoughts on top of Seb's comments.

latest/index.bs
joshmoore: No immediate objection to breaking which this loosening of the 5D/strict-ordering set up, but it might be nice as a part of this effort to plan out the next 1 or 2 large changes based on #35 and see which are _also_ going to be breaking.

latest/index.bs
constantinpape: Makes sense. I will change it.

latest/index.bs
constantinpape: If we promote "axes" to MUST, then the interpretation always depends on the metadata, i.e. on the values in "axes".

latest/index.bs
constantinpape: Ok, we can also leave it as it is; I don't have any strong feelings either way here.

latest/index.bs
constantinpape: > No immediate objection to breaking now along with this loosening of the 5D/strict-ordering version, but it might be nice as a part of this effort to plan out the next 1 or 2 large changes based on #35 and see which are _also_ going to be breaking.

Good point. I will try to summarise the current discussion in there later so we can see what else down the line might be breaking.

constantinpape: I changed to MUST now, also summarised potential changes: https://github.com/ome/ngff/issues/35#issuecomment-800662190. As far as I can see none of this would be breaking with the current proposal for 0.2.

latest/index.bs
```diff
@@ -207,8 +207,52 @@ keys as specified below for discovering certain types of data, especially images
 
 Metadata about the multiple resolution representations of the image can be
 found under the "multiscales" key in the group-level metadata.
-The specification for the multiscale (i.e. "resolution") metadata is provided
-in [zarr-specs#50](https://github.com/zarr-developers/zarr-specs/issues/50).
+"multiscales" contains a list of dictionaries where each entry describes a multiscale image.
+
+"multiscales" MUST contain the field "datasets", which is a list of dictionaries describing
```
will-moore, line 212: It is a bit confusing to have `"multiscales" contains a list of dictionaries...` followed by `"multiscales" MUST contain the field "datasets"...`
I think the second sentence needs to say `Each "multiscales" dictionary MUST contain the field "datasets"...`

latest/index.bs
```diff
@@ -207,8 +207,52 @@ keys as specified below for discovering certain types of data, especially images
...
+        {
+            "version": "0.2",
+            "name": "example",
+            "datasets": [
+                {"path": "0"},
+                {"path": "1"},
+                {"path": "2"}
+            ],
+            "axes": [
+                "t", "c", "z", "y", "z"
```
will-moore, line 241: You have 2 "z" in this list. Needs "x"

latest/index.bs
constantinpape: Good point.

jni:
Hi all,

Sorry to go around in circles, but I'd prefer relaxing MUST to SHOULD here. In the absence of axis labels, consumers could process things as:

- 5D: backwards-compatible behaviour, TCZYX
- Anything else: all spatial dimensions unless time or channel axes are specified *by the consumer*.

Of course, the latter behaviour is asking for trouble, but that's what SHOULD means.

My two cents. Of course, since backwards compatibility is no object given the versioning scheme, I am not too stressed about this choice. :blush: Thanks @constantinpape for jumping in!

constantinpape:
@jni  I would still opt for MUST, because not having "axes" is inviting undefined behaviour and we will also require to have this
in future iterations of the spec, see discussion in #35. So requiring it now is also more future proof.

>     * Anything else: all spatial dimensions unless time or channel axes are specified _by the consumer_.
> 
> 
> Of course, the latter behaviour is asking for trouble, but that's what SHOULD means.

Consumer software can still treat this metadata as a hint and over-ride it or allow the user to override; or even dismiss it completely.

joshmoore:
FWIW xarray's use of zarr *requires* an _ARRAY_DIMENSIONS attribute: http://xarray.pydata.org/en/stable/internals.html#zarr-encoding-specification 

see also: https://github.com/pydata/xarray/issues/4118#issuecomment-801240559 cc: @rabernat

jni:
@constantinpape 

> Consumer software can still treat this metadata as a hint and over-ride it or allow the user to override; or even dismiss it completely.

I'm not worried about consumers, but writers/converters. Many image formats don't have well-defined axis names, so if I want to convert some TIFF or .npy to NGFFv0.2, I now have to make up some axes. Which I guess is fine, zyx is a reasonable guess as I point out above, but it feels yuckier at write time (I'm codifying my guess to disk) than at read time (my guess is ephemeral until I close the software).

d-v-b:
this is a problem for xarray, which it solves by using default names `dim_0`, `dim_1`, etc. Those names seem a little more generic than x,y,z

constantinpape:
> this is a problem for xarray, which it solves by using default names `dim_0`, `dim_1`, etc. Those names seem a little more generic than x,y,z

But the xarray scope is also a more general than NGFF. I think x, y, z, c, y will cover most of the use case AND these changes are just an intermediate step until we can agree on a more general solution (that may also involve allowing more than 5d), see discussion in #35.
We need to make sure though that what we decide here does not require breaking changes in a more general solution, see https://github.com/ome/ngff/issues/35#issuecomment-800662190.

joshmoore:
@tlambert03 just pinged me elsewhere that this is a general improvement in terms of readability (e.g. "how does one store an image with only one resolution?" Answer: "as a multiscales of array length = 1")

joshmoore:
Are there any other high-level points, or is the discussion now primarily around whether this is a MUST or a SHOULD?

If so, I'm inclined to bump this to `0.3` (due to #40) and get it in as "latest" while we start on implementations.

constantinpape:
> Are there any other high-level points, or is the discussion now primarily around whether this is a MUST or a SHOULD?

Yes, I think this is the only open point of discussion.



> If so, I'm inclined to bump this to `0.3` (due to #40) and get it in as "latest" while we start on implementations.

:+1:

constantinpape:
Let's revitalize this PR. I think there are only 2 things to do before this can be merged:
- Do we classify `axes` as MUST or SHOULD?
    - As mentioned above, I think the changes here are only a first step towards #35. So I don't have a very strong preference here and SHOULD would also be fine with me.
- How do we coordinate with #40? Either of the options below is fine with me.
    - Merge this PR before it
    - Merge #40 and  then merge this one, fixing potential merge conflicts.
    - Rebase this PR on #40, merge it and then merge into main

joshmoore:
> * Do we classify `axes` as MUST or SHOULD?

I'll just add a data point to this discussion for now: https://twitter.com/shoyer/status/1379877702254034946

> * How do we coordinate with #40?

@constantinpape, if you can prepare a commit which is the information from zarr-developers/zarr-specs#50 without the specific bits of this PR, then I'm happy to cherry-pick and get 0.2.0 out the door. Then you can move this PR to 0.3.0.

constantinpape:
Thanks for the reviews everyone.
I will close this PR to not get confused; we will first incorporate the wording changes into 0.2 and then work on adding `axes`.

joshmoore:
This PR has been migrated to https://github.com/ome/ngff/pull/46

Please join the conversation there. It would be good to get a first version of axes naming in place.

PR #44, opened 2021-04-26
Add description of multiscales metadata 0.2

constantinpape:
This was part of #39

Issue #45, opened 2021-04-26
Spec development process


constantinpape:
I am a bit confused by the different versions of the spec that are stored in this repository:
We have the specs for the individual versions, right now `0.1` and `0.2` and we have `latest`.
Currently, `0.2` is further ahead then `latest`, because it contains the changes to the chunk layout from #40, which are not in `latest` yet.

In the future, I would propose a more structured dev process:
All changes are made against `latest` and only when we draft a new release, we copy `latest` to the folder for the corresponding version and add this with a separate commit (which only copies the spec and does not introduce any further changes).

cc @joshmoore

joshmoore:
Agreed, @constantinpape. This was my mistake because I was thinking you "owned" latest while I was making a non-breaking change. Something that hadn't been done before. The changes that just got merged to `0.2` will need backporting to `latest`. I can take care of that if you are not already in the middle of it.

P.S. It might be though that these static copies aren't ideal anyway and we want something less prone to conflicts.

constantinpape:
> Agreed, @constantinpape. This was my mistake because I was thinking you "owned" latest while I was making a non-breaking change. Something that hadn't been done before.

Sure, no worries. Luckily the changes are still very manageable, so it's not problem right now to sync this manually.

> The changes that just got merged to `0.2` will need backporting to `latest`. I can take care of that if you are not already in the middle of it.

I am in the middle of it already, see #46. I just need to sync the latest review changes later.

> P.S. It might be though that these static copies aren't ideal anyway and we want something less prone to conflicts.

I think the static copies are fine, if we keep development strictly limited to `latest`.

constantinpape:
Ok, after looking at the review comments in #46: I guess it would be best to introduce
some placeholder for the version number in `latest` and then write a script that makes the copy and replaces the placeholder with the correct version.

joshmoore:
Hmm... on call, but I don't see an immedaite way to replace a variable. Files can be included if that would help: https://tabatkins.github.io/bikeshed/#including

constantinpape:
> Hmm... on call, but I don't see an immedaite way to replace a variable.

Ok, I will have a closer look at bikeshed too and see if I can find something.
If there is nothing in there, we could also raise an issue about it.
Alternatively, it would be fairly easy to write some script for this ourselves (but we would need to introduce our own bikeshed flavor)

> Files can be included if that would help: https://tabatkins.github.io/bikeshed/#including

Thanks for the pointer.

[...]

PR #46, opened 2021-04-26
Add axes field to multiscale metadata

constantinpape:
Add the `axes` field to `multiscales` and update `latest` to v0.2. Pending #44 and release of 0.2 (that's why I marked it as a draft).

The discussion whether `axes` is mandatory (MUST) or encouraged (SHOULD) is still pending. cc @joshmoore @sbesson @jni

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -2,7 +2,8 @@
 Title: Next-generation file formats (NGFF)
 Shortname: ome-ngff
 Level: 1
-Status: w3c/CG-FINAL
+Status: LS-COMMIT
+Status: w3c/ED
 Group: ome
 URL: https://ngff.openmicroscopy.org/latest/
 Repository: https://github.com/ome/ngff
@@ -17,11 +18,11 @@ Editor: Sébastien Besson, Open Microscopy Environment (OME) https://www.openmic
 Abstract: This document contains next-generation file format (NGFF)
 Abstract: specifications for storing bioimaging data in the cloud.
 Abstract: All specifications are submitted to the https://image.sc community for review.
-Status Text: This is the 0.3 release of this specification. Migration scripts
-Status Text: will be provided between numbered versions. Data written with the latest version
+Status Text: The current released version of this specification is
+Status Text: <a href="../0.3/index.html">0.3</a>. Migration scripts
+Status Text: will be provided between numbered versions. Data written with these latest changes
 Status Text: (an "editor's draft") will not necessarily be supported.
 </pre>
-
 
 Introduction {#intro}
 =====================
@@ -597,6 +598,11 @@ Version History {#history}
       <td>Description</td>
     </tr>
   </thead>
+  <tr>
+    <td>0.3.0</td>
+    <td>2021-08-24</td>
+    <td>Add axes field to multiscale metadata     </td>
+  </tr>
   <tr>
     <td>0.2.0</td>
     <td>2021-03-29</td>
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -116,7 +116,8 @@ multiple levels of resolutions and optionally associated labels.
     │
     ├── .zgroup               # Each image is a Zarr group, or a folder, of other groups and arrays.
     ├── .zattrs               # Group level attributes are stored in the .zattrs file and include
-    │                         #  "multiscales" and "omero" below)
+    │                         # "multiscales" and "omero" (see below). In addition, the group level attributes
+    │                         # must also contain "_ARRAY_DIMENSIONS" if this group directly contains multi-scale arrays.
     │
     ├── 0                     # Each multiscale level is stored as a separate Zarr array,
     │   ...                   # which is a folder containing chunk files which compose the array.
@@ -218,6 +219,8 @@ to the current zarr group. The "path"s MUST be ordered from largest (i.e. highes
 It MUST contain the field "axes", which is a list of dimension names of the axes.
 The values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
 The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
+In addition, the "axes" values MUST be repeated in the field "_ARRAY_DIMENSIONS" of the top-level group for the arrays containing
+the multiscale data. This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
 
 It SHOULD contain the field "name".
 
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -15,6 +15,7 @@ Boilerplate: style-darkmode off
 Markup Shorthands: markdown yes
 Editor: Josh Moore, Open Microscopy Environment (OME) https://www.openmicroscopy.org
 Editor: Sébastien Besson, Open Microscopy Environment (OME) https://www.openmicroscopy.org
+Editor: Constantin Pape, European Molecular Biology Laboratory (EMBL) https://www.embl.org/sites/heidelberg/
 Abstract: This document contains next-generation file format (NGFF)
 Abstract: specifications for storing bioimaging data in the cloud.
 Abstract: All specifications are submitted to the https://image.sc community for review.
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -123,7 +123,7 @@ multiple levels of resolutions and optionally associated labels.
     ├── n                     # The name of the array is arbitrary with the ordering defined by
     │   │                     # by the "multiscales" metadata, but is often a sequence starting at 0.
     │   │
-    │   ├── .zarray           # All image arrays are 5-dimensional
+    │   ├── .zarray           # All image arrays must be up to 5-dimensional
     │   │                     # with dimension order (t, c, z, y, x).
     │   │
     │   └─ t                  # Chunks are stored with the nested directory layout.
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -219,8 +219,9 @@ to the current zarr group. The "path"s MUST be ordered from largest (i.e. highes
 It MUST contain the field "axes", which is a list of dimension names of the axes.
 The values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
 The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
-In addition, the "axes" values MUST be repeated in the field "_ARRAY_DIMENSIONS" of the top-level group for the arrays containing
-the multiscale data. This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
+In addition, the "axes" values MUST be repeated in the field "_ARRAY_DIMENSIONS" of all scale groups
+(i.e. groups containing arrays with the multiscale data).
+This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
 
 It SHOULD contain the field "name".
 
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -215,6 +215,10 @@ the arrays storing the individual resolution levels.
 Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 
+It MUST contain the field "axes", which is a list of dimension names of the axes.
+The values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
+The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
+
 It SHOULD contain the field "name".
 
 It SHOULD contain the field "version", which indicates the version of the
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -4,7 +4,7 @@ Shortname: ome-ngff
 Level: 1
 Status: w3c/CG-FINAL
 Group: ome
-URL: https://ngff.openmicroscopy.org/0.2/
+URL: https://ngff.openmicroscopy.org/latest/
 Repository: https://github.com/ome/ngff
 Issue Tracking: Forums https://forum.image.sc/tag/ome-ngff
 Logo: http://www.openmicroscopy.org/img/logos/ome-logomark.svg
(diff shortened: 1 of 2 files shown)


latest/index.bs
```diff
@@ -2,10 +2,9 @@
 Title: Next-generation file formats (NGFF)
 Shortname: ome-ngff
 Level: 1
-Status: LS-COMMIT
-Status: w3c/ED
+Status: w3c/CG-FINAL
 Group: ome
-URL: https://ngff.openmicroscopy.org/latest/
+URL: https://ngff.openmicroscopy.org/0.2/
```
joshmoore, line 7: These will need reverting. As a part of https://github.com/ome/ngff/issues/45, this is one of the annoyances of the current static copies.

latest/index.bs
```diff
@@ -18,12 +17,12 @@ Editor: Sébastien Besson, Open Microscopy Environment (OME) https://www.openmic
 Abstract: This document contains next-generation file format (NGFF)
 Abstract: specifications for storing bioimaging data in the cloud.
 Abstract: All specifications are submitted to the https://image.sc community for review.
-Status Text: The current released version of this specification is
-Status Text: <a href="../0.1/index.html">0.1</a>. Migration scripts
-Status Text: will be provided between numbered versions. Data written with these latest changes
+Status Text: This is the 0.2 release of this specification. Migration scripts
```
joshmoore, line 20: Ditto

latest/index.bs
```diff
@@ -127,10 +126,11 @@ multiple levels of resolutions and optionally associated labels.
     │   ├── .zarray           # All image arrays are 5-dimensional
     │   │                     # with dimension order (t, c, z, y, x).
     │   │
-    │   ├── 0.0.0.0.0         # Chunks are stored with the flat directory layout.
-    │   │   ...               # Each dotted component of the chunk file represents
-    │   └── t.c.z.y.x         # a "chunk coordinate", where the maximum coordinate
-    │                         # will be `dimension_size / chunk_size`.
+    │   └─ t                  # Chunks are stored with the nested directory layout.
```
joshmoore, line 129: :+1: for being a backport of https://github.com/ome/ngff/pull/40

latest/index.bs
```diff
@@ -235,7 +276,7 @@ can be found under the "omero" key in the group-level metadata:
 ```json
 "id": 1,                              # ID in OMERO
 "name": "example.tif",                # Name as shown in the UI
-"version": "0.1",                     # Current version
+"version": "0.2",                     # Current version
```
joshmoore, line 279: Will this be bumped to 0.3?

joshmoore (commented):
Few thoughts (sorry if you're still in the process of updating)

latest/index.bs
constantinpape: Good point.
I think we should introduce some placeholder for this in the future (https://github.com/ome/ngff/issues/45#issuecomment-826686522), but let's first clean this up manually here.

constantinpape:
I synced the latest changes to 0.2 now and updated all version numbers (I could find) to 0.3 now.

constantinpape:
When migrating #39 I forgot to update the dimensionality requirements, i.e. allowing for image data with less than 5 dimensions.
I added this in the last commit.
@joshmoore let me know if you think we should formulate this more explicitly.

thewtex (commented):
👍  @constantinpape explicit dimensions are helpful to make the dimension identity self-evident and make the format easy to understand and use.

In the interest of compatibility with xarray, we could update the way the axes are encoded with:

``` 
        "0/example/.zattrs": {
            "_ARRAY_DIMENSIONS": [
                "t",
                "c",
                "z",
                "y",
                "x", 
           ],
```

per https://xarray.pydata.org/en/stable/internals.html#zarr-encoding-specification

latest/index.bs
```diff
@@ -124,13 +123,14 @@ multiple levels of resolutions and optionally associated labels.
...
-    │   ├── .zarray           # All image arrays are 5-dimensional
+    │   ├── .zarray           # All image arrays must be up to 5-dimensional
     │   │                     # with dimension order (t, c, z, y, x).
     │   │
-    │   ├── 0.0.0.0.0         # Chunks are stored with the flat directory layout.
-    │   │   ...               # Each dotted component of the chunk file represents
-    │   └── t.c.z.y.x         # a "chunk coordinate", where the maximum coordinate
-    │                         # will be `dimension_size / chunk_size`.
+    │   └─ t                  # Chunks are stored with the nested directory layout.
+    │      └─ c               # All but the last chunk element are stored as directories.
```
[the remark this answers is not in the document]
joshmoore: Hi @thewtex. Yes & no. 

## Short-term

The switch in v0.2 was to let us start actually creating these datasets before all implementations start supporting https://github.com/zarr-developers/zarr-python/pull/715, i.e. `ome-zarr-*` implementations for v0.2 can "blindly" assert the use of "/" as the dimension separator. _Python_ implementations should be in the (enviable) position that it doesn't matter. If you put the metadata in place, it should work for either assuming your zarr-python version is >=2.8.0 (note: ongoing bug fixes) Does that track with what you are seeing or am I missing something on the xarray front?

## Longer-term

Once all of this has propagated through the ecosystem I think this spec shouldn't really care about which array layout is used. So a v0.3 or later can drop the requirement all together. That, however, does not get you want you want in the general case with IPFS. Do you foresee it really being that bad? (I ask because I *know* that the flat directory is *unusable* at some scales.)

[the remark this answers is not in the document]
thewtex: @joshmoore ah, thanks for the information! I was not aware of https://github.com/zarr-developers/zarr-python/pull/715 . 

On the xarray front, they are not tied to a specific zarr spec, as far as I know, so it should cause an issue. But, it seems that we should do the same here, i.e. both layouts, `.` or `/`, would be acceptable, as you suggest, correct?

> I ask because I know that the flat directory is unusable at some scales.

I can see how this would be the case. In which store types do you experience this?

For IPFS, each entity is identified with a cryptographic-hash. Updating a chunk in a file means updating its identifier, but it also means updating the identifiers for all parent directories in the directory tree; so, deeply nested trees are not desirable.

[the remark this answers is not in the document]
joshmoore: > both layouts, `.` or `/`, would be acceptable, as you suggest, correct?

That's the plan.

> In which store types do you experience this?

DirectoryStore & FSStore

> so, deeply nested trees are not desirable.

Understood. Thanks for the info.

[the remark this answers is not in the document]
constantinpape: What do you think is the best way to move forward here, @thewtex @joshmoore?
I would rather not change this now (especially since this change was done in a previous PR already and the changes here are just about syncing the different versions of the spec). Also, like @joshmoore said, not supporting hiearchical storage for the chunks can be a big issue for large datasets on the file-system.

So I would propose that we leave this as it is for now and eventually change the spec once most implementations are agnostic to the actual separator.

If really necessary and there are some implementations that require a flat hierarchy, this could then be added as an extra implementation specific requirement. (I.e. these implementations would not be able to read all ngff formats, but only a subset.)

constantinpape:
Sorry, it took me a while to catch up with this. Holidays and then a fairly busy week otherwise.
I have left a comment for the hierarchical chunk storage.

@thewtex 
> In the interest of compatibility with xarray, we could update the way the axes are encoded with:
> 
> ```
>         "0/example/.zattrs": {
>             "_ARRAY_DIMENSIONS": [
>                 "t",
>                 "c",
>                 "z",
>                 "y",
>                 "x", 
>            ],
> ```

I am not so sure that this will work for the OME NGFF format. Potentially there could be multiple datasets which have different `axes` stored in a single zarr container.
How is this case handled by `xarray` right now?

latest/index.bs
```diff
@@ -218,6 +219,8 @@ to the current zarr group. The "path"s MUST be ordered from largest (i.e. highes
 It MUST contain the field "axes", which is a list of dimension names of the axes.
 The values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
 The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
+In addition, the "axes" values MUST be repeated in the field "_ARRAY_DIMENSIONS" of the top-level group for the arrays containing
+the multiscale data. This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
```
constantinpape, line 223: @thewtex If I understand the xarray encoding correctly this addition should ensure that we are compatible.

latest/index.bs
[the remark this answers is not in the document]
joshmoore: Yes, I'd propose we leave 0.2 as hierarchical, focus 0.3 on axes, and then we can try to remove the 0.2 restriction ASAP.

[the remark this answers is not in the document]
thewtex: 👍

latest/index.bs
thewtex: @constantinpape yes, you are right!

One tweak: *top-level group* -> *all scale groups*. This ensures xarray compatibility and that each scale can be treated independently.

constantinpape:
@joshmoore I think this is good to be merged, but I would suggest we make a 0.2.0 release in the main repository before that.

latest/index.bs
```diff
@@ -117,20 +116,22 @@ multiple levels of resolutions and optionally associated labels.
...
+    │                         # "multiscales" and "omero" (see below). In addition, the group level attributes
+    │                         # must also contain "_ARRAY_DIMENSIONS" if this group directly contains multi-scale arrays.
     │
     ├── 0                     # Each multiscale level is stored as a separate Zarr array,
     │   ...                   # which is a folder containing chunk files which compose the array.
     ├── n                     # The name of the array is arbitrary with the ordering defined by
     │   │                     # by the "multiscales" metadata, but is often a sequence starting at 0.
     │   │
-    │   ├── .zarray           # All image arrays are 5-dimensional
+    │   ├── .zarray           # All image arrays must be up to 5-dimensional
```
JacksonMaxfield, line 127: @joshmoore shouldn't this be:

`must be 5+ dimensional`? OME-Tiff has the 8D / n-D spec and it would be a shame to lose that :wink:

latest/index.bs
```diff
@@ -207,8 +208,56 @@ keys as specified below for discovering certain types of data, especially images
...
+            "version": "0.3",
+            "name": "example",
+            "datasets": [
+                {"path": "0"},
+                {"path": "1"},
+                {"path": "2"}
+            ],
+            "axes": [
+                "t", "c", "z", "y", "x"
+            ],
```
JacksonMaxfield, line 248: If we do support n-dimensional, (OME 8D spec / OME n-D spec) then this simply expands to include the dimension shorthands.

latest/index.bs
constantinpape: The goal of this PR is to allow for less(!) than 5D, so that it's not necessary to introduce singleton dimensions for 2D/3D/4D data.
There is an ongoing discussion about allowing for more than 5D, see https://github.com/ome/ngff/issues/35. But that is out of scope for the current PR.

joshmoore:
@constantinpape, a few points before I disappear on holidays:
 * https://github.com/ome/ngff/pull/50 will conflict but hopefully get us into a better place.
 * in general, :+1: for the strategy of this PR to make changes to `latest/`
 * we need a process and/or script for migrating latest's index.bs to a given version location
 * you'll need to change https://github.com/ome/ngff/blob/main/.github/workflows/publish.yml#L18 to `uses: joshmoore/spec-prod@0_3` until someone with better JS skills than me makes that repo behave (I haven't checked if the upstream version does what we want yet)

latest/index.bs
```diff
@@ -2,8 +2,7 @@
 Title: Next-generation file formats (NGFF)
 Shortname: ome-ngff
 Level: 1
-Status: LS-COMMIT
-Status: w3c/ED
+Status: w3c/CG-FINAL
```
joshmoore, line 5: you likely want to revert these top-matter items

latest/index.bs
```diff
@@ -18,12 +17,12 @@ Editor: Sébastien Besson, Open Microscopy Environment (OME) https://www.openmic
 Abstract: This document contains next-generation file format (NGFF)
 Abstract: specifications for storing bioimaging data in the cloud.
 Abstract: All specifications are submitted to the https://image.sc community for review.
-Status Text: The current released version of this specification is
-Status Text: <a href="../0.1/index.html">0.1</a>. Migration scripts
```
joshmoore, line 22: This text should stay as is but point to the `0.3` that you will create when `latest/index.bs` is done.

latest/index.bs
```diff
@@ -551,6 +597,11 @@ Version History {#history}
       <td>Description</td>
     </tr>
   </thead>
+  <tr>
```
joshmoore, line 600: Plus add in your own changelog.

joshmoore (changes requested):
Few last minute items to go on top of https://github.com/ome/ngff/pull/50

joshmoore:
I've dealt with merge conflicts and handled the "publication steps" (such as they are) if anyone would like to take a look.

joshmoore:
If no objections, I'll merge this and move forward with https://github.com/orgs/ome/projects/17?add_cards_query=is%3Aopen

will-moore:
Sorry, just a thought....

Do we want limit the last two axes to be `["y", "x"]` (at least 2D)?
As I understand it, the idea of `v0.3` was to allow less than 5D, but as it stands, these are all valid:

```
["c"],
["x", "y"],  # inverted
["c", "y", "t", "x"],
```

But do we want all clients that support the spec to also support all of those?

will-moore:
Ah - sorry, I missed the existing caveat `# with dimension order (t, c, z, y, x).` which hasn't been changed by this PR.
It's just that this isn't mentioned in the list of `axes` rules in the text.
And there's still no 2D minimum - but maybe that's OK since we're all dealing with Images.
So I don't think this is a blocker.
Apologies for the noise!

sbesson (approved):
This new version of the specification has already been tested in various read and write scenarios across different implementations including omero-cli-zarr, vizarr, MoBIE. The specification changes look good so I am in favor of getting this published and completing the various tasks associated with OME-NGFF 0.3.

jni:
Just spinning back up after my hiking-trip-turned-lockdown-staycation and I'm so delighted to find this merged! 🎉 Thank you SO much to everyone for your patience and perseverance with this! ❤️‍🔥 (◀️ just discovered that emoji and it's perfect!)

Issue #31, opened 2021-02-24
Collections Specification


DragaDoncila:
**What is an image collection?**
A collection of images is a semantic grouping of two or more associated ome-ngff images and/or image-labels. 

This definition could include
- Images which do not share a physical coordinate space e.g. training dataset of images containing bees
- Images which share a physical coordinate space and whose storage specification must support sufficient metadata to determine this positioning e.g. high-content screening plates and wells
- A hierarchy of image groups of arbitrary depth which may or may not share physical coordinates
- Other things…?

**What workflows should it support?**
The specification should support implementations being able to traverse the image collection and, where relevant, map the associated metadata to the physical coordinate space for loading these images.

Ideally, the specification should provide sufficient information at each level of a hierarchical grouping to allow for the loading of both the entire collection, and the loading of an arbitrary level of the hierarchy. This can be important when wanting to share/view partial datasets or update only small parts of the entire collection.

Where labels or other related data is provided (e.g. meshes, points…), the specification should support being able to associate any member of the image collection with its associated labels, regardless of the level in the hierarchy.

The OME-NGFF spec is close to supporting this functionality with the HCS specification which allows the positioning of wells into rows and plates. The main drawbacks of this specification are
- It is too specific to be easily used for images which ARE physically associated but are not HCS acquisitions
- It may be difficult to understand for researchers who are not working with HCS images but nevertheless wish to store their collection in OME-NGFF format
- It does not support an arbitrary depth of groupings
- It does not support collections which are not physically associated

**What should it be called?**
- Dataset - this term is already used in various places so may not be the best choice
- Collection - a general enough term which is currently mostly unused
- Hierarchical definition - there is a case for this specification being a hierarchy of specifications, with each one defining a more tightly bound collection e.g.
    - Bag - associated images with no metadata
    - Stack - associated images which overlap in physical space
    - Panorama - associated images which stitch together in physical space

Ideally, the names used in the base specification would be general enough to support a broad variety of use cases and tailored use cases could be demonstrated using examples in the documentation.

**Reference specifications**
[BDV XML Files](https://imagej.net/BigDataViewer.html#Multi-view_data_converted_to_XML.2FHDF5)
[SVG](https://www.w3.org/TR/SVG2/coords.html)
[TrakEM2](https://imagej.net/TrakEM2)
[Napari Plugin for image-label collections](https://github.com/DragaDoncila/napari-compressed-labels-io)
[mobie grid view of many sources](https://github.com/mobie/mobie/issues/20)

[...]

PR #57, opened 2021-09-07
Extend the axes fields in multiscales metadata

constantinpape:
## Proposal summary

This proposal adds "axes" metadata to describe the axes / dimensions of a physical coordinate data space and "transformation" metadata.
Both axes and transformation metadata are then used in "multiscales" to describe the physical coordinate system of the data and the mapping from data coordinates to physical coordinates (for each scale).

Note that we only introduce simple "scale" and "translation" transformations here, but intend to expand the set of supported transformations in the future. See #67 for initial discussions.

The main motivation of the proposed changes is to enable expressing "simple" spatial coordinates for a multiscale image that can be interpreted by all viewers (i.e. also viewers without support for more complex transformations like standard Fiji). 
As a second design goal, the transformation spec outlined here should be extensible to more complex transformations (see again discussions in #67) to be able to later express these in the same spec.

Some example data for ome.zarr files using this proposal are available here: https://github.com/ome/ome-ngff-prototypes.

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -219,11 +219,14 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 
 It MUST contain the field "axes", which is a list of dimension names of the axes.
-The values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
-The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
-In addition, the "axes" values MUST be repeated in the field "_ARRAY_DIMENSIONS" of all scale groups
-(i.e. groups containing arrays with the multiscale data).
+The values provide axes labels and MUST be unique, the number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
+In addition, the "axes" values MUST be repeated in the field "_ARRAY_DIMENSIONS" of all scale groups (i.e. groups containing arrays with the multiscale data).
 This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
+
+It MUST contain the field "axes_types", which is a list that describes the semantic type of each axis. It MUST contain only the elements "space", "time" and "channel" and MUST have the same length as "axes".
+These types are hints TODO
+
+It MUST contain the field "units" TODO
 
 It SHOULD contain the field "name".
 
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -216,13 +216,13 @@ keys as specified below for discovering certain types of data, especially images
 If part of [[#multiscale-md]], the length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
 
 
-"transformation" metadata {#trafo-md}
+"transformations" metadata {#trafo-md}
 -------------------------------------
 
-Describes a transformation, e.g. to transform the discrete data space of an array to the physical space.
-It is a dictionary, which MUST contain the field "type".
+"transformations" describes a series of transformations, e.g. to map discrete data space of an array to the corresponding physical space.
+It is a list of dictionaries. Each entry describes a single transformation and MUST contain the field "type".
 The value of "type" MUST be one of the elements of the `type` column in the table below.
-Additional fields are defined by the column `fields`.
+Additional fields for the entry depend on "type" and are defined by the column `fields`.
 
 | type          | fields | description |
 |-              |-       |-            |
@@ -233,7 +233,7 @@ Additional fields are defined by the column `fields`.
 In addition, the field "axisIndices" MAY be given to specify the subset of axes that the transformation is applied to, leaving other axes unchanged. If not given, the transformation is applied to all axes. The length of "axisIndices" MUST be equal to the dimensionality of the transformation. If "axisIndices" are not given, the dimensionality of the transformation MUST be equal to the number of dimensions of the space that the transformation is applied to.
 If given, "axisIndices" MUST be given in increasing order.
 
-If transformations are stored in a list, e.g. as part of [[#multiscale-md]] metadata, they are always applied sequentally and in order.
+The transformations in the list are applied sequentally and in order.
 
 
 "multiscales" metadata {#multiscale-md}
@@ -253,17 +253,16 @@ the arrays storing the individual resolution levels.
 Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 All arrays MUST have the same number of dimensions and MUST NOT have more than 5 dimensions.
-Each dictionary MAY contain the field "transformations", which contains a list of [[#trafo-md]] that specifies the transformation from the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
-The transformations MUST only be of type `identity`, `translation` or `scale`.
-The list MUST contain at most one `scale` transformation per axis taht specifies the size in physical units.
-It also MUST contain at most one `translation` per axis that specifies the offset in physical units.
-If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates.
-The transformations in the list are applied sequentially and in order. If not given, the identity transformation is assumed.
+Each dictionary MAY contain the field "transformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
+The transformations are defined according to [[#trafo-md]]. In addition, the transformation types MUST only be `identity`, `translation` or `scale`.
+They MUST contain at most one `scale` transformation per axis that specifies the pixel size in physical units.
+It also MUST contain at most one `translation` per axis that specifies the offset from the origin in physical units.
+If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates. If "transformations" is not given, the identity transformation is assumed.
 The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while
 being compatible with the general transformation spec.
 
-It MAY contain the field "transformations" containing a list of [[#trafo-md]], describing transformations that are applied to each resolution level. 
-The transformations MUST follow the same rules about allowed values and order of `scale` and `translation` as in "datasets".
+It MAY contain the field "transformations", describing transformations that are applied to each resolution level.
+The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:transformations".
 These transformations are applied after the per resolution level transformations specified in "datasets". They can for example be used to specify the `scale` for a dimension that is the same for all resolutions.
 
 It SHOULD contain the field "name".
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -227,7 +227,7 @@ The value of "type" MUST be one of the elements of the `type` column in the tabl
 Additional fields for the entry depend on "type" and are defined by the column `fields`.
 
 | type          | fields | description |
-|-              |-       |-            |
+| ------------- | ------ |------------ |
 | `identity`    |        | identity transformation, is the default transformation and is typically not explicitly defined |
 | `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 | `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -210,7 +210,7 @@ keys as specified below for discovering certain types of data, especially images
 
 "axes" describes the dimensions of a physical coordinate space. It is a list of dictionaries, where each dictionary describes an dimension (axis) and:
 - MUST contain the field "name" that gives the name for this dimension. The values MUST be unique across all "name" fields.
-- SHOULD contain the field "type" to specify the type of this dimension. The value SHOULD be one of "space", "channel" or "time".
+- SHOULD contain the field "type" to specify the type of this dimension. The value SHOULD be one of "space", "channel" or "time". If "type" is not given, it is assumed to be "null", i.e. unkown or not represented by the spec yet.
 - SHOULD contain the field "unit" to specify the physical unit of this dimension. The value SHOULD be a valid unit according to UDUNITS-2.
 
 If part of [[#multiscale-md]], the length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -239,20 +239,22 @@ The transformations in the list are applied sequentally and in order.
 "multiscales" metadata {#multiscale-md}
 ---------------------------------------
 
-Metadata about the multiple resolution representations of the image can be
-found under the "multiscales" key in the group-level metadata.
+Metadata about an image can be found under the "multiscales" key in the group-level metadata. Here, image refers to 2 to 5 dimensional data representing image or volumetric data with optional time or channel axes. It is stored in a multiple resolution representation.
 
 "multiscales" contains a list of dictionaries where each entry describes a multiscale image.
 
 Each dictionary MUST contain the field "axes", see [[#axes-md]].
-The values of the "name" field must be repeated in the field "_ARRAY_DIMENSIONS" of all scale groups (i.e. groups containing arrays with the multiscale data).
+The length of "axes" must be between 2 and 5 and MUST be equal to the dimensionality of the zarr arrays storing the image data (see "datasets:path").
+The "axes" MUST contain 2 or 3 entries of "type:space" and MAY contain one additional entry of "type:time" and MAY contain one additional entry of "type:channel" or a null / custom type.
+The order of the entries MUST correspond to the order of dimensions of the zarr arrays. In addition, the entries MUST be ordered by "type" where the "time" axis must come first (if present), followed by the  "channel" or custom axis (if present) and the axes of type "space".
+The values of the "name" fields must be given as a list in the field "_ARRAY_DIMENSIONS" in the attributes (.zattr) of the zarr arrays.
 This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
+E.g. for "axes: [{"name": "x"}, {"name": "y"}, {"name": z}]", the zarr arrays must contain "{"_ARRAY_DIMENSIONS": ["x", "y", "z"]}" in their attributes.
 
-It MUST contain the field "datasets", which is a list of dictionaries describing
-the arrays storing the individual resolution levels.
+It MUST contain the field "datasets", which is a list of dictionaries describing the arrays storing the individual resolution levels.
 Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
-All arrays MUST have the same number of dimensions and MUST NOT have more than 5 dimensions.
+All arrays MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
 Each dictionary MAY contain the field "transformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
 The transformations are defined according to [[#trafo-md]]. In addition, the transformation types MUST only be `identity`, `translation` or `scale`.
 They MUST contain at most one `scale` transformation per axis that specifies the pixel size in physical units.
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -259,7 +259,7 @@ The transformations MUST only be of type `identity`, `translation`, `scale` or `
 The list MUST contain at most one "scale" transformation per dimensioon, which specifies the size of one physical unit for that dimension.
 The transformations in the list are applied sequentially and in order. If not given, the identity transformation is assumed.
 
-It MAY contain the field "transformations" with a list of [[#trafo-md]], describing transformations that are applied in the same manner to each resolution level, following the same rules as "transformations" in "datasets".
+It MAY contain the field "transformations" with a list of [[#trafo-md]], describing transformations that are applied in to each resolution level, following the same rules as "transformations" in "datasets". These transformations are applied after the per resolution level transformations specified in "datasets".
 
 It SHOULD contain the field "name".
 
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -216,6 +216,27 @@ keys as specified below for discovering certain types of data, especially images
 If part of [[#multiscale-md]], the length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
 
 
+"transformation" metadata {#trafo-md}
+-------------------------------------
+
+Describes a transformation, e.g. to transform the discrete data space of an array to the physical space.
+It is a dictionary, which MUST contain the field "type".
+The value of "type" MUST be one of the elements of the `type` column in the table below.
+Additional fields are defined by the column `fields`.
+
+| type          | fields | description |
+|-              |-       |-            |
+| `identity`    |        | identity transformation, is the default transformation and is typically not explicitly defined |
+| `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
+| `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
+| `rotation`    | one of: `"rotation": List[float], "path:str"` | rotation vector, stored either as a list of floats (`rotation`) pr as binary data at a location in this container (`path`). The length of the vector must either be one for a rotation in two dimensions or three for a rotation in three dimenions. In 3d, the rotation angles are given in the order TODO. |
+
+In addition, the field "axisIndices" MAY be given to specify the subset of axes that the transformation is applied to, leaving other axes unchanged. If not given, the transformation is applied to all axes. The length of "axisIndices" MUST be equal to the dimensionality of the transformation. If "axisIndices" are not given, the dimensionality of the transformation MUST be equal to the number of dimensions of the space that the transformation is applied to.
+If given, "axisIndices" MUST be given in increasing order.
+
+If transformations are stored in a list, e.g. as part of [[#multiscale-md]] metadata, they are always applied sequentally and in order.
+
+
 "multiscales" metadata {#multiscale-md}
 ---------------------------------------
 
@@ -224,15 +245,21 @@ found under the "multiscales" key in the group-level metadata.
 
 "multiscales" contains a list of dictionaries where each entry describes a multiscale image.
 
-Each dictionary contained in the list MUST contain the field "datasets", which is a list of dictionaries describing
+Each dictionary MUST contain the field "axes", see [[#axes-md]].
+The values of the "name" field must be repeated in the field "_ARRAY_DIMENSIONS" of all scale groups (i.e. groups containing arrays with the multiscale data).
+This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
+
+It MUST contain the field "datasets", which is a list of dictionaries describing
 the arrays storing the individual resolution levels.
 Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 All arrays MUST have the same number of dimensions and MUST NOT have more than 5 dimensions.
+Each dictionary MAY contain the field "transformations", which contains a list of [[#trafo-md]] that specifies the transfromation from the data coordinate space to the physical coordinate space (as specified by "axes") for this resolution level.
+The transformations MUST only be of type `identity`, `translation`, `scale` or `rotation`. This restrictions ensures a simple mapping from data space to physical space.
+The list MUST contain at most one "scale" transformation per dimensioon, which specifies the size of one physical unit for that dimension.
+The transformations in the list are applied sequentially and in order. If not given, the identity transformation is assumed.
 
-It MUST contain the field "axes", see [[#axes-md]].
-The values of the "name" field must be repeated in the field "_ARRAY_DIMENSIONS" of all scale groups (i.e. groups containing arrays with the multiscale data).
-This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
+It MAY contain the field "transformations" with a list of [[#trafo-md]], describing transformations that are applied in the same manner to each resolution level, following the same rules as "transformations" in "datasets".
 
 It SHOULD contain the field "name".
 
@@ -243,17 +270,12 @@ It SHOULD contain the field "type", which gives the type of downscaling method u
 
 It SHOULD contain the field "metadata", which contains a dictionary with additional information about the downscaling method.
 
-```json
+```
 {
     "multiscales": [
         {
             "version": "0.3",
             "name": "example",
-            "datasets": [
-                {"path": "0"},
-                {"path": "1"},
-                {"path": "2"}
-            ],
             "axes": [
                 {"name": "t", "type": "time", "unit": "millisecond"},
                 {"name": "c", "type": "channel"},
@@ -261,6 +283,21 @@ It SHOULD contain the field "metadata", which contains a dictionary with additio
                 {"name": "y", "type": "space", "unit": "micrometer"},
                 {"name": "x", "type": "space", "unit": "micrometer"}
             ],
+            "datasets": [
+                {
+                    "path": "0",
+                    "transformations": [{"type": "scale", "scale": [0.5, 0.5, 0.5], "axisIndices": [2, 3, 4]}]  # the voxel size for the first scale level (0.5 micrometer)
+                }
+                {
+                    "path": "1",
+                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 1.0], "axisIndices": [2, 3, 4]}]  # the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
+                },
+                {
+                    "path": "2",
+                    "transformations": [{"type": "scale", "scale": [2.0, 2.0, 2.0], "axisIndices": [2, 3, 4]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
+                }
+            ],
+            "transformations": [{"type": "scale", "scale": [0.1], "axisIndices": [0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
             "type": "gaussian",
             "metadata": {                                       # the fields in metadata depend on the downscaling implementation
                 "method": "skimage.transform.pyramid_gaussian", # here, the paramters passed to the skimage function are given
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -234,11 +234,6 @@ It MUST contain the field "axes", see [[#axes-md]] and the length of the lists i
 The "labels" list must be repeated in the field "_ARRAY_DIMENSIONS" of all scale groups (i.e. groups containing arrays with the multiscale data).
 This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
 
-It MUST contain the field "axes_types", which is a list that describes the semantic type of each axis. It MUST contain only the elements "space", "time" and "channel" and MUST have the same length as "axes".
-These types are hints TODO
-
-It MUST contain the field "units" TODO
-
 It SHOULD contain the field "name".
 
 It SHOULD contain the field "version", which indicates the version of the
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -247,7 +247,7 @@ Metadata about an image can be found under the "multiscales" key in the group-le
 
 "multiscales" contains a list of dictionaries where each entry describes a multiscale image.
 
-Each dictionary MUST contain the field "axes", see [[#axes-md]].
+Each "multiscales" dictionary MUST contain the field "axes", see [[#axes-md]].
 The length of "axes" must be between 2 and 5 and MUST be equal to the dimensionality of the zarr arrays storing the image data (see "datasets:path").
 The "axes" MUST contain 2 or 3 entries of "type:space" and MAY contain one additional entry of "type:time" and MAY contain one additional entry of "type:channel" or a null / custom type.
 The order of the entries MUST correspond to the order of dimensions of the zarr arrays. In addition, the entries MUST be ordered by "type" where the "time" axis must come first (if present), followed by the  "channel" or custom axis (if present) and the axes of type "space".
@@ -256,10 +256,11 @@ The values of the "name" fields must be given as a list in the field "_ARRAY_DIM
 This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
 E.g. for "axes: [{"name": "z"}, {"name": "y"}, {"name": x}]", the zarr arrays must contain "{"_ARRAY_DIMENSIONS": ["z", "y", "x"]}" in their attributes.
 
-It MUST contain the field "datasets", which is a list of dictionaries describing the arrays storing the individual resolution levels.
+Each "multiscales" dictionary MUST contain the field "datasets", which is a list of dictionaries describing the arrays storing the individual resolution levels.
 Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
-All arrays MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
+
+Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
 Each dictionary MAY contain the field "transformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
 The transformations are defined according to [[#trafo-md]]. In addition, the transformation types MUST only be `identity`, `translation` or `scale`.
 They MUST contain at most one `scale` transformation per axis that specifies the pixel size in physical units.
@@ -268,24 +269,20 @@ If both `scale` and `translation` are given `translation` must be listed after `
 The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while
 being compatible with the general transformation spec.
 
-It MAY contain the field "transformations", describing transformations that are applied to each resolution level.
+Each "multiscales" dictionary MAY contain the field "transformations", describing transformations that are applied to each resolution level.
 The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:transformations".
 These transformations are applied after the per resolution level transformations specified in "datasets". They can for example be used to specify the `scale` for a dimension that is the same for all resolutions.
 
-It SHOULD contain the field "name".
+Each "multiscales" dictionary SHOULD contain the field "name". It SHOULD contain the field "version", which indicates the version of the multiscale metadata of this image (current version is 0.4).
 
-It SHOULD contain the field "version", which indicates the version of the
-multiscale metadata of this image (current version is 0.3).
-
-It SHOULD contain the field "type", which gives the type of downscaling method used to generate the multiscale image pyramid.
-
+Each "multiscales" dictionary SHOULD contain the field "type", which gives the type of downscaling method used to generate the multiscale image pyramid.
 It SHOULD contain the field "metadata", which contains a dictionary with additional information about the downscaling method.
 
 ```
 {
     "multiscales": [
         {
-            "version": "0.3",
+            "version": "0.4",
             "name": "example",
             "axes": [
                 {"name": "t", "type": "time", "unit": "millisecond"},
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -229,7 +229,6 @@ Additional fields are defined by the column `fields`.
 | `identity`    |        | identity transformation, is the default transformation and is typically not explicitly defined |
 | `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 | `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
-| `rotation`    | one of: `"rotation": List[float], "path:str"` | rotation vector, stored either as a list of floats (`rotation`) pr as binary data at a location in this container (`path`). The length of the vector must either be one for a rotation in two dimensions or three for a rotation in three dimenions. In 3d, the rotation angles are given in the order TODO. |
 
 In addition, the field "axisIndices" MAY be given to specify the subset of axes that the transformation is applied to, leaving other axes unchanged. If not given, the transformation is applied to all axes. The length of "axisIndices" MUST be equal to the dimensionality of the transformation. If "axisIndices" are not given, the dimensionality of the transformation MUST be equal to the number of dimensions of the space that the transformation is applied to.
 If given, "axisIndices" MUST be given in increasing order.
@@ -254,12 +253,18 @@ the arrays storing the individual resolution levels.
 Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 All arrays MUST have the same number of dimensions and MUST NOT have more than 5 dimensions.
-Each dictionary MAY contain the field "transformations", which contains a list of [[#trafo-md]] that specifies the transfromation from the data coordinate space to the physical coordinate space (as specified by "axes") for this resolution level.
-The transformations MUST only be of type `identity`, `translation`, `scale` or `rotation`. This restrictions ensures a simple mapping from data space to physical space.
-The list MUST contain at most one "scale" transformation per dimensioon, which specifies the size of one physical unit for that dimension.
+Each dictionary MAY contain the field "transformations", which contains a list of [[#trafo-md]] that specifies the transformation from the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
+The transformations MUST only be of type `identity`, `translation` or `scale`.
+The list MUST contain at most one `scale` transformation per axis taht specifies the size in physical units.
+It also MUST contain at most one `translation` per axis that specifies the offset in physical units.
+If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates.
 The transformations in the list are applied sequentially and in order. If not given, the identity transformation is assumed.
+The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while
+being compatible with the general transformation spec.
 
-It MAY contain the field "transformations" with a list of [[#trafo-md]], describing transformations that are applied in to each resolution level, following the same rules as "transformations" in "datasets". These transformations are applied after the per resolution level transformations specified in "datasets".
+It MAY contain the field "transformations" containing a list of [[#trafo-md]], describing transformations that are applied to each resolution level. 
+The transformations MUST follow the same rules about allowed values and order of `scale` and `translation` as in "datasets".
+These transformations are applied after the per resolution level transformations specified in "datasets". They can for example be used to specify the `scale` for a dimension that is the same for all resolutions.
 
 It SHOULD contain the field "name".
 
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -210,7 +210,7 @@ keys as specified below for discovering certain types of data, especially images
 
 Describes axes of a physical coordinate space. It is a dictionary, which MUST contain the fields:
 - "labels": list of strings that specify the name per dimension. The values MUST be unique.
-- "types": list of strings that specify the type per dimension. The values SHOULD be one of "space", "channel", "time" or "". Use "" if none of the other options apply.
+- "types": list of strings that specify the type per dimension. The values SHOULD be one of "space", "channel", "time".
 - "units": list of strings that specify the unit per dimension.
 
 The three lists MUST have the same length.
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -212,8 +212,10 @@ keys as specified below for discovering certain types of data, especially images
 
 "axes" describes the dimensions of a physical coordinate space. It is a list of dictionaries, where each dictionary describes an dimension (axis) and:
 - MUST contain the field "name" that gives the name for this dimension. The values MUST be unique across all "name" fields.
-- SHOULD contain the field "type" to specify the type of this dimension. The value SHOULD be one of "space", "channel" or "time". If "type" is not given, it is assumed to be "null", i.e. unkown or not represented by the spec yet.
 - SHOULD contain the field "unit" to specify the physical unit of this dimension. The value SHOULD be a valid unit according to UDUNITS-2.
+SHOULD contain the field "unit" to specify the physical unit of this dimension. The value SHOULD be one of the following strings, which are valid units according to UDUNITS-2.
+    - Units for "space" axes: 'angstrom', 'attometer', 'centimeter', 'decimeter', 'exameter', 'femtometer', 'foot', 'gigameter', 'hectometer', 'inch', 'kilometer', 'megameter', 'meter', 'micrometer', 'mile', 'millimeter', 'nanometer', 'parsec', 'petameter', 'picometer', 'terameter', 'yard', 'yoctometer', 'yottameter', 'zeptometer', 'zettameter'
+    - Units for "time" axes: 'attosecond', 'centisecond', 'day', 'decisecond', 'exasecond', 'femtosecond', 'gigasecond', 'hectosecond', 'hour', 'kilosecond', 'megasecond', 'microsecond', 'millisecond', 'minute', 'nanosecond', 'petasecond', 'picosecond', 'second', 'terasecond', 'yoctosecond', 'yottasecond', 'zeptosecond', 'zettasecond'
 
 If part of [[#multiscale-md]], the length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
 
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -107,6 +107,8 @@ Images {#image-layout}
 
 The following layout describes the expected Zarr hierarchy for images with
 multiple levels of resolutions and optionally associated labels.
+Note that the number of dimensions is variable between 2 and 5 and that axis names are arbitrary, see [[#multiscale-md]] for details.
+For this example we assume an image with 5 dimensions and axes called `t,c,z,y,x`.
 
 ```
 .                             # Root folder, potentially in S3,
@@ -127,7 +129,7 @@ multiple levels of resolutions and optionally associated labels.
     │   │                     # by the "multiscales" metadata, but is often a sequence starting at 0.
     │   │
     │   ├── .zarray           # All image arrays must be up to 5-dimensional
-    │   │                     # with dimension order (t, c, z, y, x).
+    │   │                     # with the axis of type time before type channel, before spatial axes.
     │   │
     │   └─ t                  # Chunks are stored with the nested directory layout.
     │      └─ c               # All but the last chunk element are stored as directories.
@@ -231,7 +233,7 @@ Additional fields for the entry depend on "type" and are defined by the column `
 | `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 
 In addition, the field "axisIndices" MAY be given to specify the subset of axes that the transformation is applied to, leaving other axes unchanged. If not given, the transformation is applied to all axes. The length of "axisIndices" MUST be equal to the dimensionality of the transformation. If "axisIndices" are not given, the dimensionality of the transformation MUST be equal to the number of dimensions of the space that the transformation is applied to.
-If given, "axisIndices" MUST be given in increasing order.
+If given, "axisIndices" MUST be given in increasing order. It uses zero-based indexing.
 
 The transformations in the list are applied sequentally and in order.
 
@@ -247,9 +249,10 @@ Each dictionary MUST contain the field "axes", see [[#axes-md]].
 The length of "axes" must be between 2 and 5 and MUST be equal to the dimensionality of the zarr arrays storing the image data (see "datasets:path").
 The "axes" MUST contain 2 or 3 entries of "type:space" and MAY contain one additional entry of "type:time" and MAY contain one additional entry of "type:channel" or a null / custom type.
 The order of the entries MUST correspond to the order of dimensions of the zarr arrays. In addition, the entries MUST be ordered by "type" where the "time" axis must come first (if present), followed by the  "channel" or custom axis (if present) and the axes of type "space".
+If there are three spatial axes where two correspond to the image plane ("yx") and images are stacked along the other (anisotropic) axis ("z"), the spatial axes SHOULD be ordered as "zyx".
 The values of the "name" fields must be given as a list in the field "_ARRAY_DIMENSIONS" in the attributes (.zattr) of the zarr arrays.
 This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
-E.g. for "axes: [{"name": "x"}, {"name": "y"}, {"name": z}]", the zarr arrays must contain "{"_ARRAY_DIMENSIONS": ["x", "y", "z"]}" in their attributes.
+E.g. for "axes: [{"name": "z"}, {"name": "y"}, {"name": x}]", the zarr arrays must contain "{"_ARRAY_DIMENSIONS": ["z", "y", "x"]}" in their attributes.
 
 It MUST contain the field "datasets", which is a list of dictionaries describing the arrays storing the individual resolution levels.
 Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -205,6 +205,18 @@ Metadata {#metadata}
 The various `.zattrs` files throughout the above array hierarchy may contain metadata
 keys as specified below for discovering certain types of data, especially images.
 
+"axes" metadata {#axes-md}
+--------------------------
+
+Describes axes of a physical coordinate space. It is a dictionary, which MUST contain the fields:
+- "labels": list of strings that specify the name per dimension. The values MUST be unique.
+- "types": list of strings that specify the type per dimension. The values SHOULD be one of "space", "channel", "time" or "". Use "" if none of the other options apply.
+- "units": list of strings that specify the unit per dimension.
+
+The three lists MUST have the same length.
+If part of [[#multiscale-md]], the length MUST be equal to the array.
+
+
 "multiscales" metadata {#multiscale-md}
 ---------------------------------------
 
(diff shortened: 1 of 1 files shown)


constantinpape:
# First draft for separate axes metadata

Adapted from https://github.com/saalfeldlab/n5-ij/wiki/Transformation-spec-proposal-for-NGFF#proposal-a. 
Find the design choice and my reasoning behind it below; comments / critique are welcome. 
There are still some open questions listed as well where I'd be very happy about some input.

## Design choice

Axes metadata describes the physical coordinate space (= output coordinate space after transformation). It is defined separate from `multiscales` to make it easier to refer to it and reuse in other spec definitions if this becomes necessary.
For similar reasons the transformation spec will also be defined separately, see #63 (WIP).
This design is extensible (new axes_types or transformations can be added) and easily allows lifting the only up-to-5d restriction (these were 2 of the most important points of discussion in the issue / last ngff meeting).

## Open Questions

**Units:** Do we restrict the units or is this free text? If we choose to restrict, what set of units is allowed? Some options that were brought up in #35: SI-Units, https://people.csail.mit.edu/jaffer/MIXF/CMIXF-12.

**Level of requirement:** E.g., must there be units or is it should? Also, do we allow for types that are not in "space", "time", "channel" (right now it's SHOULD, so we do allow other values).

latest/index.bs
```diff
@@ -205,6 +205,18 @@ Metadata {#metadata}
 The various `.zattrs` files throughout the above array hierarchy may contain metadata
 keys as specified below for discovering certain types of data, especially images.
 
+"axes" metadata {#axes-md}
+--------------------------
+
+Describes axes of a physical coordinate space. It is a dictionary, which MUST contain the fields:
+- "labels": list of strings that specify the name per dimension. The values MUST be unique.
+- "types": list of strings that specify the type per dimension. The values SHOULD be one of "space", "channel", "time".
+- "units": list of strings that specify the unit per dimension.
```
will-moore, line 214: What are the enumerated values here? e.g. "micrometer", "MICROMETER", "µm"?
In OMERO we use enums `'ANGSTROM', 'ASTRONOMICALUNIT', 'ATTOMETER', 'CENTIMETER', 'DECAMETER', 'DECIMETER', 'EXAMETER', 'FEMTOMETER', 'FOOT', 'GIGAMETER', 'HECTOMETER', 'INCH', 'KILOMETER', 'LIGHTYEAR', 'LINE', 'MEGAMETER', 'METER', 'MICROMETER', 'MILE', 'MILLIMETER', 'NANOMETER', 'PARSEC', 'PETAMETER', 'PICOMETER', 'PIXEL', 'POINT', 'REFERENCEFRAME', 'TERAMETER', 'THOU', 'YARD', 'YOCTOMETER', 'YOTTAMETER', 'ZEPTOMETER', 'ZETTAMETER'` for length and 

`'ATTOSECOND', 'CENTISECOND', 'DAY', 'DECASECOND', 'DECISECOND', 'EXASECOND', 'FEMTOSECOND', 'GIGASECOND', 'HECTOSECOND', 'HOUR', 'KILOSECOND', 'MEGASECOND', 'MICROSECOND', 'MILLISECOND', 'MINUTE', 'NANOSECOND', 'PETASECOND', 'PICOSECOND', 'SECOND', 'TERASECOND', 'YOCTOSECOND', 'YOTTASECOND', 'ZEPTOSECOND', 'ZETTASECOND'` for time.
Although I don't know what the original source for these was.

constantinpape: Good point, @will-moore. I have listed this under open questions here: https://github.com/ome/ngff/pull/57#issuecomment-932055668.

will-moore:
Are we still restricting the number of dimensions to 5?
Do we still care about the order of dimensions. Currently we have `All image arrays must be up to 5-dimensional with dimension order (t, c, z, y, x).`
Certainly a fair bit of code (vizarr, ome-zarr-py etc) would struggle to support x and y dimensions being not last.

NB: there's a few other places in the example JSON sections that need to be updated.

constantinpape:
> Are we still restricting the number of dimensions to 5?

Yes, I also made this more explicit in b3523c9. 

> Certainly a fair bit of code (vizarr, ome-zarr-py etc) would struggle to support x and y dimensions being not last.

Indeed, we should clarify this more. We can't rely on fixed axes names "xyz" any more (and the demand for this was very clear in the last meeting). But we can add a MUST or SHOULD that the spatial axes should be last.

> NB: there's a few other places in the example JSON sections that need to be updated.

Yes, I will wait with updating until there is a consensus to adopt this proposal and we settled some of the open questions.

satra:
i think units are a must especially once transforms and coordinate systems and other image spaces come into the picture. 

btw, there is a bit of discussion around metadata of transforms in this document (https://docs.google.com/document/d/11gCzXOPUbYyuQx8fErtMO9tnOKC3kTWiL9axWkkILNE/edit) which is part extending the [bids-specification](https://bids-specification.readthedocs.io/en/stable/). 

regarding the actual transformations themselves, itk has a set of good transform representations (i'm sure @thewtex can say a whole lot more :) )

unidesigner:
Thanks @constantinpape for following-up on this. Would it be useful to extract a list of questions from the neuroglancer discussion that may be relevant to discuss here?

Regarding the question about restricting the number of dimension to be maximally 5, and your reply making this more explicit in the docs: Is there really a consensus to do this? Also regarding enforcing explicitly for the last two dimensions to be spatial? I don't think that is generally useful, and would possible limit what can be done in the future with ome-zarr files. Tools that do have specific requirements on the contained data (e.g. requiring at least two or three spatial dimensions, or with more than 5 dimensions etc.) could just fail with an error message in those cases.

constantinpape:
> Would it be useful to extract a list of questions from the neuroglancer discussion that may be relevant to discuss here?

Yes, that would be helpful.

> Regarding the question about restricting the number of dimension to be maximally 5, and your reply making this more explicit in the docs: Is there really a consensus to do this?

I think there is still quite some discussions going on about this restrictions, with probably more support to lift this requirement. However, I would like to avoid opening up this topic for the next release to focus on the axis metadata format. I think lifting the requirement would be very easy in the current proposal (but let me know if you see any potential issues). Same for the axis order discussion.

unidesigner:
Okey. Is there a timeline for when the next release is planned?

If the axes metadata is expressed as lists is a good start to express more or less dimensions. As already discussed a bit, one has to be careful by implying to much from a label name, i.e. assuming too much implicitly, e.g. when writing `x`. As this also very closely relates to the transformation specs, one has to consider what to do, and how to specify, what to do with the non-spatial dimensions and transformations (e.g. imagine a temporal dimension being time-shifted), and not to constrain them to a fixed number of dimensions (e.g.@jbms suggestions of using permuted block-diagonal matrix to specify the affine [in the issue](https://github.com/google/neuroglancer/issues/333). I think it will be crucial to be able to define identifiers for different possible spaces, and use those identifiers in the transformation specification (what is the source space? what is the target space?), so that it becomes possible to make a valid interpretation of what e.g. an `x` axes stands for in any given (transformed) space. This may also already solve some of the axes (re)ordering discussion - imagine specifying a "identity" permutation matrix transformation (e.g. in the format that is used to specify affines), which basically just reorders the axes.

jbms:
While in common cases it seems to make sense to label full coordinate spaces, e.g. where you have a common xyz coordinate space to which multiple volumes are aligned, in general I think it actually makes more sense to label individual dimensions rather than the full coordinate space, and that could be accomplished just by using longer dimension names, and possibly having some way to abbreviate them in viewers.  For example:

fafb.v14.x
fafb.v14.y
fafb.v14.z

For example, you might have a combined EM/functional dataset consisting:

EM alignment v2 with dimensions:
dataset.v2.x
dataset.v2.y
dataset.v2.z

Multi-channel functional recording experiment 1:

dataset.v2.x
dataset.v2.y
dataset.v2.z
dataset.v2.exp1.t
dataset.v2.c

Multi-channel functional recording experiment 2:

dataset.v2.x
dataset.v2.y
dataset.v2.z
dataset.v2.exp2.t
dataset.v2.c

By labeling dimensions we can fully represent the relationship between the three volumes.  If we labeled entire coordinate spaces, we would have to assign a separate coordinate space label to each volume.

bogovicj:
@unidesigner ( Hi Stephan!)

> one has to be careful by implying to much from a label name

yes, agreed.  Label names are possibly arbitrary, but possibly not. They should be allowed to carry information (I quite like @jbms 's example), but not required to. 

> I think it will be crucial to be able to define identifiers for different possible spaces, and use those identifiers in the transformation specification (what is the source space? what is the target space?), so that it becomes possible to make a valid interpretation of what e.g. an x axes stands for in any given (transformed) space. 

This is essentially what we describe in our [transformation "proposal - D"](https://github.com/saalfeldlab/n5-ij/wiki/Transformation-spec-proposal-for-NGFF#proposal-d): giving names to input and output axes, and specifying very explicitly which input axes a given transform applies to and what output axes it generates.  It's quite cumbersome though, and likely "going overboard".

constantinpape:
@unidesigner

> Okey. Is there a timeline for when the next release is planned?

There are two steps for this: getting this PR and #63 merged to incorporate transformations (of course including discussions on what still needs to change with the proposals). Then producing example data and making sure it runs in the reference implementations (currently Fiji via MoBIE, vizarr and napari via the napari-ome-zarr plugin). I can't give you a clear timetable on the first point, because this very much depends on how much discussion is needed. Updating the reference implementations should be relatively fast and be feasible in ~ 1 week (depending of course on the availability of developers / contributors of the individual tools).

> As already discussed a bit, one has to be careful by implying to much from a label name, i.e. assuming too much implicitly, e.g. when writing `x`

In the current proposal I introduced 3 lists in the axes objects: `label` (which specifies the name and is currently arbitrary), `type` (which specifies the axis type and should be one of `space`, `time`, `channel` (note the should, so custom types are allowed but discouraged)). 

> As this also very closely relates to the transformation specs

For the transformations please refer to #63 and have a look at the current proposal and how it relates to what you have written below. (Note that this is based on https://github.com/saalfeldlab/n5-ij/wiki/Transformation-spec-proposal-for-NGFF and inspired by Proposal C but I tried to simplify the proposal a bit; the major differences are listed on top of the PR).

@jbms the current draft allows arbitrary labels and has an additional field `types` to specify the type of the axis.
So introducing more complex naming patterns is certainly possible but not required (and I am not sure aiming for mandatory naming patterns is a good idea at this point; but one could certainly discuss naming conventions, this would merit a separate discussion in another issue though.)

bogovicj:
My opinion on open questions

> Units: Do we restrict the units or is this free text?

units SHOULD be 
* SI or SI-derived (e.g. "radian")  
* or some indicator for arbitrary units (`"null"`, `"N/A"`, `"none"`, `"pixel"`, or similar)

but any string is technically allowed.

> Also, do we allow for types that are not in "space", "time", "channel" (right now it's SHOULD, so we do allow other values).

Agree with SHOULD, I'll probably only ever use these tree, but others should be allowed ( e.g. ["polarization angle"](https://github.com/ome/ngff/issues/35#issuecomment-792235945) )

> E.g., must there be units or is it should

SHOULD.  if units are empty "arbitrary" / "pixel" units are implied

bogovicj:
@jbms 
You could specify voxel size in the way you want with the `scale` transform in the transform spec:
https://github.com/constantinpape/ngff/blob/trafo-spec/latest/index.bs#L231

this is equivalent to giving the diagonal elements of the affine only. does that work for you?

jbms:
I think it may still be desirable to be able to specify the "native" resolution of the data in many cases.  For example, Neuroglancer chooses the units (including multiplier) of the global coordinate space by default to match the native resolution of the data.  The units/multiplier of the global coordinate space affect how coordinates are displayed, the step size for discrete movement commands, and also have implications for numerical precision.  If there is just an affine transform, we could attempt to heuristically extract the units and multiplier from that, but that seems to leave more ambiguity than if the units are specified explicitly.

In general, it seems that if you allow SI prefixes, which are just another form of multiplier, albeit one that is nicer to display in a scale bar, then it would be logical to also allow arbitrary multipliers.

d-v-b:
The way I think of it (and this is highly influenced by my exposure to xarray / netcdf), `unit` defines a property of an axis, where an axis is a vector of values, and therefor the `unit` property is orthogonal to the contents of that axis vector. In this model, `unit: nm` doesn't imply a native resolution of 1 nm -- it just implies that the elements of the axis are to be interpreted as positions in nanometers, which is a pretty clean abstraction, insofar as it ultimately allows irregularly sampled data. 

Since we don't have axes explicitly populated with values in metadata, the `transform` property is needed to define the sampling grid of the data. So from this perspective it's natural to infer the "native" resolution from the transform and not from axis labels. Plus, "4nm" is not a unit :)

jbms:
You have a good point that by using a coordinate array you gain an additional level of indirection and can use it to specify multipliers.  But I would imagine that for regularly spaced data a coordinate array would not normally be specified.

According to the udunits2 library, "4nm" is a unit:
https://www.unidata.ucar.edu/software/udunits/udunits-2.2.28/udunits2lib.html#Syntax

In general while aesthetically "nanometer" is a nicer unit to work with than "4 nanometer", I don't see why fundamentally there should be support for "1e-9 meter" as a unit but not for "4e-9 meter" as a unit.

If the affine transform is not a permutation of a diagonal matrix, then we can't infer scale factors from it, so we would be back to just using whatever units are specified explicitly.  For example, maybe you have aligned functional data to an EM volume, and there is an affine transform involved, but you want to specify the EM native resolution as the explicit units.

It sounds like the spec is merely suggesting a syntax for units, and so "4nm" could still be used, but I think it would help to specify how these units are expected to be used.

bogovicj:
There are two related but separate (in my mind) issues that I feel we're mixing up a bit and will try to dis-entangle, specifically how do we distinguish between:
1. transformations from the discrete-grid to a physical space and 
2. transformations from physical-to-physical space or image-to-image transformations.

### relating to the conversation above 

Here's how I see these mapping in @jbms comment above - please correct / confirm Jeremy.

> I think it may still be desirable to be able to specify the "native" resolution of the data in many cases. 

this is (1), and I agree.

>  For example, maybe you have aligned functional data to an EM volume, and there is an affine transform involved, ...

this refers to (2)

> ...  but you want to specify the EM native resolution as the explicit units.

and this is (1), and I agree.

### is (2) in the spec?
As in Jeremy's example, say I have an EM dataset, and I specify the (1) transform "this image has resolution  `4 x  4 x 40nm`" (`S`) It's aligned to the functional dataset, so I spec the (2) transform affine `A` (EM -> functional).

My viewer now needs to know what space I want my EM to be displayed in before it can do the right thing.
If I want to look in native EM "nm" space, I need the transform `S` only, but if I'm aligning with functional I need
`SA`

Either, the spec covers (1) only, or needs a much more concrete sense of coordinate space / axes as [above](https://github.com/ome/ngff/pull/57#issuecomment-932372276) to be unambiguous - something more like our [proposal D](https://github.com/saalfeldlab/n5-ij/wiki/Transformation-spec-proposal-for-NGFF#proposal-d).  
Are we tackling this now?

jbms:
In regards to your worry, I would say that if you specify that the units of the output space are "4nm", then your transform had better be designed to output coordinates in units of "4nm", not units of "nm", i.e. all of the coefficients in that row of the affine transform should be divided by 4 compared to if the output space is "nm".

That is the same issue that comes up with having an output unit of "um", and a scale factor of 1e-3, vs having an output unit of "nm" and a scale factor of 1.

aeisenbarth:
- Should `labels` be renamed to `names` to avoid varying usage of the term (compared to `labels` images)?

constantinpape:
Thanks for all the discussions and feedback.

@bogovicj 

> My opinion on open questions [...]

I very much agree with your assessments, esp. SHOULD for the units (relevant for all the following discussions).
This means we need to define the recommended unit names. As far as I can see there are 2 options:
- Define a list of names in the spec (see comment by @will-moore above).
- Use a library, e.g. https://www.unidata.ucar.edu/software/udunits/udunits-2.2.28/udunits2lib.html#Syntax or https://people.csail.mit.edu/jaffer/MIXF/CMIXF-12 (which may not be ideal because of cross compatibility)

@bogovicj @d-v-b @jbms 

I personally would also prefer "clean" units, i.e. only having "nanometer" or "nm" in the recommended unit names and not "4 nm".
(But if we keep it at SHOULD, specifying "4nm" is still valid, but not encouraged.)

Regarding the distinction 1) use transformations to specify discrete to physical space vs 2) to specify physical to physical space:
I think that the transformations stored in "multiscales" should only be used for 1) (which also makes it feasible to extract the pixel spacing). For 2), we would then use the collection spec #31, where we would reuse the transformation spec from #63.
Note that the collection spec is still WIP, but we have started some discussions on this with the OME-Team, @tischi, @tpietzsch and @StephanPreibisch. We have a follow up meeting scheduled and will make a better job to announce this and make a summary of the meeting available afterwards. (Also note that this separation would only be best practice / recommended, i.e. it would still be possible, but discouraged, to define a complex transformation in "multiscales".)

@aeisenbarth 

I agree that `names` might be a bit better name here.

axtimwalde:
I am in favor of @jbms proposal to fully adapt UDUNITS-2 if we do not have a good alternative. Any arbitrary subspec would need to be subspecced and I am not sure we want to go down that road. The clean alternative is to get rid of the units and define only the type of axes and a transformation that converts into standard units (m, s, kg(I know), K, ...). We should keep in mind that this transformation, in most scenarios will be used to render scale bars and do measurements, rendering stuff onto screens will typically use other transformations since they aren't 'physical' spaces. The only thing that I dislike about UDUNITS-2 is that it incorporates a very specific subset of transformations, n-d affine transformations and per-axis scaling, translation, and log-scaling. The xarray-style labels work only for discrete data (right?) which may be enough for labeling entries in regular n-d arrays but not for transformations that could bend things.

Other things about style:
1. Why n independent arrays of strings instead of one array of structs? Structs are easier with default values.
1. We had the idea to associate the space definition with transformations, i.e. a transformation specifies that it can consume two spatial axes and a time axis and exports three spatial axes (serial section scanner or light sheet).
1. If the transformation stuff is getting too hairy, may be we leave it for an independent transformation spec, but then we should run with the xarray model (for complicated things) plus UDUNITS-2 for what it covers?

constantinpape:
Thanks for the comments @axtimwalde.

I will start by addressing the comments about transformations:

> The only thing that I dislike about UDUNITS-2 is that it incorporates a very specific subset of transformations, n-d affine transformations and per-axis scaling, translation, and log-scaling.

I am not very familiar with UDUNITS-2 and couldn't find any mention of transformations in the docs / [C-API guide](https://www.unidata.ucar.edu/software/udunits/udunits-2.2.28/udunits2lib.html#Unary).
Could you please point me to what you are refering to? 

> The xarray-style labels work only for discrete data (right?) which may be enough for labeling entries in regular n-d arrays but not for transformations that could bend things.

As far as I understand, xarray only supports discrete spacing for the individual axes, but not for the whole space.
E.g. for a 2D array with size (X,Y) one has X + Y parameters defining the spacing per axis, **not** X*Y for the spacing per point . See `coords` in https://xarray.pydata.org/en/stable/user-guide/data-structures.html#creating-a-dataarray. Please correct me if I am wrong (someone with more xarray experience @joshmoore @thewtex ?)

>     3\. If the transformation stuff is getting too hairy, may be we leave it for an independent transformation spec, but then we should run with the xarray model (for complicated things) plus UDUNITS-2 for what it covers?

Should we first see if #63 is feasible? (And how potential issues with it could be fixed?) I tried to boil down your proposals there, see https://github.com/ome/ngff/pull/63#issuecomment-932194777 for a summary of the differences.
(Also, from my understanding xarray would not give us full flexibility for things like b-splines and I couldn't find the UDUNITS-2 transform definitions, see above.)

>     2\. We had the idea to associate the space definition with transformations, i.e. a transformation specifies that it can consume two spatial axes and a time axis and exports three spatial axes (serial section scanner or light sheet).

Yes, I think that would still be feasible in #63 using `axisIndices`. It does not support reordering of axes though. I am not sure how one would do this in a manner that is portable between e.g. Java and python/numpy that prefer different axes orders.

About the changes in this PR (so only the axes metadata)

> I am in favor of @jbms proposal to fully adapt UDUNITS-2 if we do not have a good alternative. Any arbitrary subspec would need to be subspecced and I am not sure we want to go down that road.

I haven't looked into udunits closer, but at least from python it should be widely available through [cf-uints](https://cf-units.readthedocs.io/en/latest/unit.html). Still having a fixed list might be easier for cross-compatibility.

>     1. Why n independent arrays of strings instead of one array of structs? Structs are easier with default values.

I chose the independent arrays for readability (so rather subjective) and having the independent arrays makes it slightly easier to refer to this in other parts of the spec, e.g. "same values as the `labels` array of `axes`" instead of "same values as the value of `label` in each element of `axes`".
But I fully agree with the point about default values, so I would not mind changing this to array of structs.

axtimwalde:
UDUNITS-2 supports offsets (@), scaling (4nm), quotients (4nm/2s), and log scaling (lg), I confused the quotients with linear combinations earlier, then it would support full affine transformations but that is apparently not the case. Offsets and scaling are subsets of affine transformations, the others are other per axis transformations. So some transformations are baked into the units framework.

constantinpape:
@satra I had a closer look at the example application you posted, but realized that this is a massive dataset already:

Screenshot from 2021-10-15 14-52-15

This is a bit too complex for me to put together a quick example... :D.
Do you think that in principle the strategy laid out here for the transformations in ome.ngff would fit your needs?
Also, feel free to create a PR or issue in https://github.com/constantinpape/ome-ngff-implementations if you want to discuss your use-case and how to map to ome.zarr in more details.

As a plan going forward: I will finalize the examples in https://github.com/constantinpape/ome-ngff-implementations next week and go over the PRs here again to incorporate some of the other changes we discussed and then announce this proposal on image.sc to get more feedback.

thewtex:
> Why n independent arrays of strings instead of one array of structs? Structs are easier with default values.

> I chose the independent arrays for readability (so rather subjective) and having the independent arrays makes it slightly easier to refer to this in other parts of the spec,

I see the readability appeal of multiple arrays. However, to keep the spec simpler, and, more importantly to make it harder to create bugs when working with the spec, the array of structs seems preferable. The constraint

> The three lists MUST have the same length.

Does not have to be explicitly enforced.

thewtex:
> As far as I understand, xarray only supports discrete spacing for the individual axes, but not for the whole space.
E.g. for a 2D array with size (X,Y) one has X + Y parameters defining the spacing per axis, not X*Y for the spacing per point . See coords in https://xarray.pydata.org/en/stable/user-guide/data-structures.html#creating-a-dataarray. Please correct me if I am wrong (someone with more xarray experience @joshmoore @thewtex ?)

@constantinpape yes, you are correct.

For scientific imaging / ngff purposes, the xarray coordinate array capability makes the most sense for specifying times that may be irregularly sampled. Also, absolute times, as opposed to an array of floats, for example, preserve valuable information in its natural form. In the geospatial community, they deal with time dimensions frequently and have resolved many details to handle it properly. We may want to support the date-time array format that xarray (and related tooling) supports.

constantinpape:
Thanks a lot for the feedback @thewtex.

> In addition to the `translation` and `scale` for the `d2p`, a MAY to consider supporting is an additional `rotate` transformation (we would have to add it to #63, it could be specified the same as `affine` but with an orthonormal constraint).

I agree, adding a `rotate` transformation makes sense to add in general; and we can also add it in `d2p` as parsing the pixel size from `scale` is still feasible with `rotate` (and you motivate adding it very well given its utility in medical imaging).
I would prefer to add it as a one / three parameter (2d / 3d) transform though, but we can discuss that in #63.

> > 
> 
> I see the readability appeal of multiple arrays. However, to keep the spec simpler, and, more importantly to make it harder to create bugs when implementing the spec, the array of structs seems preferable. The constraint
> 
> > The three lists MUST have the same length.
> 
> does not have to be explicitly enforced.

Yes, I agree that array of structs is preferable; I will update the proposal.

> For scientific imaging / ngff purposes, the xarray coordinate array capability makes the most sense for specifying times that may be irregularly sampled. Also, absolute times, as opposed to an array of floats, for example, preserve valuable information in its natural form. In the geospatial community, they deal with time dimensions frequently and have resolved many details to handle it properly. We may want to support the date-time array format that xarray (and related tooling) supports.

Thanks for the clarification! I would vote for leaving this up to a later version update to keep the current discussion more focused; but let me know if you think any of the changes discussed here would clash with future changes to include xarray style coordinate arrays.

constantinpape:
@satra @thewtex @unidesigner @bogovicj @axtimwalde
 
After the feedback I got I have decided to  bundle all changes for v0.4 in this PR, see discussion in the last comments and changes in the last 3 commits. I have also updated https://github.com/constantinpape/ome-ngff-implementations accordingly and will host example data for 0.4 soon (our institute s3 is currently under maintenance, I will upload once that is done.)

On my TODOs there are still three open questions for the changes here:
- [ ] `unit` field: how exactly do we define `valid according to UDUNITS2`?
- [ ] `rotation`: define the order of rotation angles for 3d  
- [ ] `axes`: think about fields `from` `to`, see https://github.com/ome/ngff/pull/63#discussion_r729998007. (I like this idea a lot, but we may leave it for a later version to flesh it out more before introducing it).

Once these points are addressed and the example data is uploaded I would announce this proposal on the forum for more feedback.

tischi:
> Do I understand correctly that translation would now be in floating point values of unit (e.g. nm, um etc.) independent of scale?

@unidesigner I had similar questions. I discussed with @constantinpape and apparently the current idea is a list of transformations (`scale`, `translation`, and maybe `rotate`) and the transformations would be applied in the order that they are given in the list. In that scenario, I think the units of `translation` would then depend on whether it would come before or after `scale`. In addition, I think in order to implement a useful rotation one quite often would need to specify two `translation` in order to specify the center of the rotation (e.g., `translate` to center, `rotate`, `translate` back). I also had a discussion with @constantinpape raising my concern that consuming a rotation might not be straightforward for all software. For example, in ImageJ (`ImagePlus`) there is no metadata field for `rotation`, but only for the pixel size (`scale`) and the origin (`translate`), which means that one would have to actually apply the rotation in terms of resampling the data into a new voxel grid in RAM and one would have to decide on an interpolation mode; moreover I was confused what the "voxel size" in the resampled rotated coordinate system would be. 

Long story short, personally, I think, as a first step, for now just adding one `scale` and one `translation` in a fixed order (I think I would prefer first scale and then translation such that translation would be specified in physical coordinates) to the specification would be the easiest way forward.

latest/index.bs
```diff
@@ -205,6 +205,37 @@ Metadata {#metadata}
 The various `.zattrs` files throughout the above array hierarchy may contain metadata
 keys as specified below for discovering certain types of data, especially images.
 
+"axes" metadata {#axes-md}
+--------------------------
+
+"axes" describes the dimensions of a physical coordinate space. It is a list of dictionaries, where each dictionary describes an dimension (axis) and:
+- MUST contain the field "name" that gives the name for this dimension. The values MUST be unique across all "name" fields.
+- SHOULD contain the field "type" to specify the type of this dimension. The value SHOULD be one of "space", "channel" or "time".
+- SHOULD contain the field "unit" to specify the physical unit of this dimension. The value SHOULD be a valid unit according to UDUNITS-2.
```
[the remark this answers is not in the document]
constantinpape: > does this mean we cannot have spatial order `x`,`y`,`z` or if i wanted to call this `i`, `j`, `k` to distinguish from physical space xyz as the mapping by transforms that would not be possible. at this stage i would rather get some metadata in than not, so consider these clarifications being sought.

Sorry, I don't understand this question. For some context: `name` has arbitrary values now, so we cannot impose an order based on it, but instead impose the order based on `type`. I hope this clarifies things; if not please rephrase the question.


> and `must be up to` suggests that any of these components are optional. doesn't this mean that you would still require names to say do `tyx` or `czyx` ?

Again, this is specified by `type` now and we require 2 or 3 axis of spatial type and one optional channel and one optional time axis.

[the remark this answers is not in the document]
will-moore: I still wonder if (or where) we should still state that the convention for ordering spatial axes is `zyx`?
Some viewers (like vizarr) are very much 2D viewers (with Z-slider). And our scaling (in ome-zarr-py) for multiscales is currently operating on XY only (not Z). napari can do orthoganal views, but *currently* it expects the last 2 axes to be `yx`.
So, support for other axes orders is less good and needs work. If you create an image with `xyz` or `xzy` axes, you'll find that it's not handled as well. This may come as a surprise to users unless it's either stated in the spec, or we have some other way of documenting the limitations of various clients/tools?

will-moore:
I'm wondering if "according to UDUNITS-2" is not the most user-friendly way of specifying what's a valid unit?
I started over at https://github.com/ome/omero-cli-zarr/pull/93 to look at writing v0.4 units from OMERO, but didn't even manage to install UDUNITS-2 via conda. I certainly don't want to try and compile the C library.
So I don't even know if I should be writing "nm" or nanometre or nanometer!

Just reading https://github.com/ome/ngff/issues/35#issuecomment-927122435 (from @satra) and https://people.csail.mit.edu/jaffer/MIXF/CMIXF-12 seems a nice reference. Just put a link to that in the spec and you can read directly that "nanometer" is correct (maybe need to state in the spec that it's the full name that we use instead of symbol for units)?

latest/index.bs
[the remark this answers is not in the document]
satra: @constantinpape - this question arose from @will-moore 's comment referring to this sentence

> All image arrays must be up to 5-dimensional with dimension order (t, c, z, y, x).

here it seems to connect dimension with name. if names are not important and only useful for connecting to visualization and interpretation, perhaps this can be restated as since zyx do not mean anything:

`All image arrays must be up to 5-dimensional with dimension order of type (time, channel, spatial, spatial, spatial)`.

i do like the current formulation of disconnecting from names for representation. if the future goal would be to relax the order, perhaps one could add that this restriction could be removed in future versions of the spec.

latest/index.bs
```diff
@@ -205,25 +205,67 @@ Metadata {#metadata}
...
+"transformations" describes a series of transformations, e.g. to map discrete data space of an array to the corresponding physical space.
+It is a list of dictionaries. Each entry describes a single transformation and MUST contain the field "type".
+The value of "type" MUST be one of the elements of the `type` column in the table below.
+Additional fields for the entry depend on "type" and are defined by the column `fields`.
+
+| type          | fields | description |
+|-              |-       |-            |
+| `identity`    |        | identity transformation, is the default transformation and is typically not explicitly defined |
+| `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
+| `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
```
will-moore, line 231: Looking at how this table is rendered at http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/ome/ngff/47620eddae0d04a078505ae7189b7e0e94f5feed/latest/index.bs#trafo-md somethings not right

will-moore: Maybe just needs at least 3 `---` below the header: https://www.markdownguide.org/extended-syntax/

thewtex (approved):
@constantinpape thank you so much for pushing this forward!

This PR LGTM from my perspective. While constrained, we have addressed the needs discussed and have a path forward where additional functionality will fit in nicely. As @satra suggests, more metadata is better than none at all.

I know that some data acquisitions or processing software may not account for `rotation`'s in the data, but I hope that we can add optional support for rotations in a future update as this metadata is essential in medical imaging or emerging large-scale whole-organism microscopy. Support for `rotation`'s will make a much better case for NGFF in these applications. Example datasets that we can use for testing are:

1. The [Allen CCF Mouse Brain Atlas](https://help.brain-map.org/display/mousebrain/API)
2. The [MNI Human Brain Atlas](http://nist.mni.mcgill.ca/mni-ftd-templates/)
3. A brain image downloaded from [OpenNeuro](https://openneuro.org/).

These images are in formats like the BIDS [NIFTI](https://nifti.nimh.nih.gov/), and contain metadata on how the whole brain is oriented relative to the imaging system. This metadata is critical for researchers that want to work with these datasets. We should be able to, for example, load these datasets in napari and observe that they are rendered with the same anatomical orientation.

bjoernthiel:
A mapping transformation would be very much appreciated (at least for non equidistant time points).

latest/index.bs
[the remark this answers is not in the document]
constantinpape: @will-moore I see your point. I will think about a formulation on recommended axis order for axes that map to `zyx` and tag you there so we can discuss this.

@satra 

> All image arrays must be up to 5-dimensional with dimension order (t, c, z, y, x).

Yes, this part of the spec is outdated and I have not updated it yet because it is in the "unstructured part" of the spec and I wanted to get things right in the other part first.
I will update it now.

latest/index.bs
constantinpape: Thanks for checking. I haven't looked at the rendering at all yet; will give it another look.

constantinpape:
@will-moore thanks for raising the concerns about udunits2. I will give this another look now and see if I can get this installed easily (compiling a c-library is certainly not an option...) and then come back to you.

@thewtex thanks for your comments and the example for rotation. I will migrate this into a separate issue later.

@bjoernthiel thanks for bringing this up. We had some discussion about introducing xarray coordinate style transformations, which I think are close to what you are proposing here. I agree that these are desirable, but I suggest we leave this for a later PR in order to keep the discussion here focused. 
Feel free to open a separate issue on this, ideally with some example data.

latest/index.bs
```diff
@@ -205,25 +205,67 @@ Metadata {#metadata}
...
+
+In addition, the field "axisIndices" MAY be given to specify the subset of axes that the transformation is applied to, leaving other axes unchanged. If not given, the transformation is applied to all axes. The length of "axisIndices" MUST be equal to the dimensionality of the transformation. If "axisIndices" are not given, the dimensionality of the transformation MUST be equal to the number of dimensions of the space that the transformation is applied to.
+If given, "axisIndices" MUST be given in increasing order.
```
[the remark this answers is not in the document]
constantinpape: Done

latest/index.bs
constantinpape: @will-moore I have tried to fix this. How do I find the link for the latest deployment to check?

constantinpape:
@will-moore for me installation of `cfunits` was straight forward and checking for validity is very easy:
```python
# install cfunits via conda:
# conda install -c conda-forge cfunits
from cfunits import Units

def check_unit(unit):
  """Check if a unit is valid according to udunits2"""
  return Units(unit).isvalid

if __name__ == "__main__":
  # this is valid
  print(check_unit("m"))
  # this is invalid
  print(check_unit("abc"))
```
Let's discuss this further in https://github.com/ome/omero-cli-zarr/pull/93 and see if we can resolve the issues for you or if this is a bigger issue and whether we indeed want to avoid adding it as a dependency.

bjoernthiel:
Is there any relation between array 'shape' and 'order' (C or F) on one hand side and '_ARRAY_DIMENSIONS' and 'axis' order on the other? If not should this by stated somewhere? The current wording seems to have C order in mind.

latest/index.bs
joshmoore: The URL is in the README. Basically just replace the checksum e.g. http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/ome/ngff/e27bdecc06dedc71c6ec7e20a2bd7dd0755d43ea/latest/index.bs#trafo-md

I think I failed to get this to work which is by the version table at the bottom uses `<table>`.

latest/index.bs
```diff
@@ -205,25 +207,68 @@ Metadata {#metadata}
...
 
 "multiscales" contains a list of dictionaries where each entry describes a multiscale image.
 
-Each dictionary contained in the list MUST contain the field "datasets", which is a list of dictionaries describing
-the arrays storing the individual resolution levels.
+Each dictionary MUST contain the field "axes", see [[#axes-md]].
+The length of "axes" must be between 2 and 5 and MUST be equal to the dimensionality of the zarr arrays storing the image data (see "datasets:path").
+The "axes" MUST contain 2 or 3 entries of "type:space" and MAY contain one additional entry of "type:time" and MAY contain one additional entry of "type:channel" or a null / custom type.
+The order of the entries MUST correspond to the order of dimensions of the zarr arrays. In addition, the entries MUST be ordered by "type" where the "time" axis must come first (if present), followed by the  "channel" or custom axis (if present) and the axes of type "space".
+If there are three spatial axes where two correspond to the image plane ("yx") and images are stacked along the other (anisotropic) axis ("z"), the spatial axes SHOULD be ordered as "zyx".
```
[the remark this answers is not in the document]
will-moore: That sounds great, thanks

constantinpape:
I had two very busy weeks and didn't have much time to follow up here, but it looks like the discussion has more or less settled (addressing a few points below).
I think we can go ahead with the next steps: implementing 0.4 in some references (FIJI/MoBIE, ome-zarr-py, vizarr).
I can organize the FIJI/MoBIE part and I think @will-moore and @sbesson have already started working on ome-zarr-py.

@bjoernthiel 
> Is there any relation between array 'shape' and 'order' (C or F) on one hand side and '_ARRAY_DIMENSIONS' and 'axis' order on the other? If not should this by stated somewhere? The current wording seems to have C order in mind.

We recommend C order, this is addressed here: https://github.com/constantinpape/ngff/blob/axes-spec/latest/index.bs#L252.
Feel free to suggest some more clarification if you think this is not sufficient.

@will-moore I agree with the constraints you raised about udunits2 in https://github.com/ome/omero-cli-zarr/pull/93#issuecomment-961171948. Do you have an alternative suggestion for dealing with the unit strings?

jbms:
It is true that it is a bit unfortunate for interoperability that udunits2 allows quite a lot of aliases.

One possibility would be to say that that units must be in canonical form as produced by udunits2, e.g.:

`3 nanometer` would have to be specified as `3e-9 m`.

will-moore:
@constantinpape What I've done in OMERO.figure is get OMERO to generate a summary of all the supported units, which I can then use in the browser to generate the UI and allow converting between any units, using 'microns' as the reference (see below).
Maybe we can do the same with `cfunits`, as @jbms suggested, just choosing 1 enum for each unit, e.g. `nanometer`, `micrometer` etc
and just put that list in the spec (doesn't even need the symbol or conversion info since users can use the library to generate that themselves).
Using the OMERO units below as a starting point, the ones that are valid in cfunits, are (in lowercase):
```
['angstrom', 'attometer', 'centimeter', 'decimeter', 'exameter', 'femtometer', 'foot', 'gigameter', 'hectometer', 'inch', 'kilometer', 'megameter', 'meter', 'micrometer', 'mile', 'millimeter', 'nanometer', 'parsec', 'petameter', 'picometer', 'terameter', 'yard', 'yoctometer', 'yottameter', 'zeptometer', 'zettameter']
```
The spec could simply state that those are the supported units?
Then at least users only have to support 1 unique key for each unit, which makes it possible to support units with something generated like this (the keys here are uppercase because that's the enums we use in OMERO):
```
{"ANGSTROM":{"symbol":"Å","microns":0.0001},"ASTRONOMICALUNIT":{"symbol":"ua","microns":149597870700000000},"ATTOMETER":{"symbol":"am","microns":1e-12},"CENTIMETER":{"symbol":"cm","microns":10000},"DECAMETER":{"symbol":"dam","microns":10000000},"DECIMETER":{"symbol":"dm","microns":100000},"EXAMETER":{"symbol":"Em","microns":1e+24},"FEMTOMETER":{"symbol":"fm","microns":1e-9},"FOOT":{"symbol":"ft","microns":304799.83540808887},"GIGAMETER":{"symbol":"Gm","microns":1000000000000000},"HECTOMETER":{"symbol":"hm","microns":100000000},"INCH":{"symbol":"in","microns":25399.986284007406},"KILOMETER":{"symbol":"km","microns":1000000000},"LIGHTYEAR":{"symbol":"ly","microns":9.4607304725808e+21},"LINE":{"symbol":"li","microns":2116.665523667284},"MEGAMETER":{"symbol":"Mm","microns":1000000000000},"METER":{"symbol":"m","microns":1000000},"MICROMETER":{"symbol":"µm","microns":1},"MILE":{"symbol":"mi","microns":1609343130.9547093},"MILLIMETER":{"symbol":"mm","microns":1000},"NANOMETER":{"symbol":"nm","microns":0.001},"PARSEC":{"symbol":"pc","microns":3.0856776e+22},"PETAMETER":{"symbol":"Pm","microns":1e+21},"PICOMETER":{"symbol":"pm","microns":0.000001},"POINT":{"symbol":"pt","microns":352.7775872778806},"TERAMETER":{"symbol":"Tm","microns":1000000000000000000},"THOU":{"symbol":"thou","microns":25399986.284007408},"YARD":{"symbol":"yd","microns":914399.5062242666},"YOCTOMETER":{"symbol":"ym","microns":1e-18},"YOTTAMETER":{"symbol":"Ym","microns":1e+30},"ZEPTOMETER":{"symbol":"zm","microns":1e-15},"ZETTAMETER":{"symbol":"Zm","microns":1e+27}}
```

constantinpape:
@will-moore ok, I like that approach. Since you have done this for OMERO already, do you maybe want to add it to the PR here as well?

latest/index.bs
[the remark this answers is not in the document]
will-moore:
```suggestion
- SHOULD contain the field "unit" to specify the physical unit of this dimension. The value SHOULD be one of the following strings, which are valid units according to UDUNITS-2.
- Units for "space" axes: 'angstrom', 'attometer', 'centimeter', 'decimeter', 'exameter', 'femtometer', 'foot', 'gigameter', 'hectometer', 'inch', 'kilometer', 'megameter', 'meter', 'micrometer', 'mile', 'millimeter', 'nanometer', 'parsec', 'petameter', 'picometer', 'terameter', 'yard', 'yoctometer', 'yottameter', 'zeptometer', 'zettameter'
- Units for "time" axes: 'attosecond', 'centisecond', 'day', 'decisecond', 'exasecond', 'femtosecond', 'gigasecond', 'hectosecond', 'hour', 'kilosecond', 'megasecond', 'microsecond', 'millisecond', 'minute', 'nanosecond', 'petasecond', 'picosecond', 'second', 'terasecond', 'yoctosecond', 'yottasecond', 'zeptosecond', 'zettasecond'
```

will-moore:
@constantinpape Added a suggested update to the spec above. Hope that helps and is not too restrictive.
We don't mention anywhere the idea that was discussed above of including a multiplier e.g. "4nm", but I guess the restriction to "micrometer" etc is only SHOULD, so users could still do that if they wish.
Also I didn't mention dimension types other than "space" and "time", since the axes type SHOULD be one of those, so we might need a different approach when we allow more axes types.

constantinpape:
Thanks @will-moore. (Somehow I couldn't find the suggestion in this PR but I got a notification via mail so I just copied the changes from there).

> We don't mention anywhere the idea that was discussed above of including a multiplier e.g. "4nm",

I think it's good to leave that out for now, as we have the `scale` to specify the size and then having a size in the units as well will be confusing.

latest/index.bs
```diff
@@ -205,25 +207,69 @@ Metadata {#metadata}
...
+| `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
+| `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
+
+In addition, the field "axisIndices" MAY be given to specify the subset of axes that the transformation is applied to, leaving other axes unchanged. If not given, the transformation is applied to all axes. The length of "axisIndices" MUST be equal to the dimensionality of the transformation. If "axisIndices" are not given, the dimensionality of the transformation MUST be equal to the number of dimensions of the space that the transformation is applied to.
+If given, "axisIndices" MUST be given in increasing order. It uses zero-based indexing.
+
+The transformations in the list are applied sequentally and in order.
```
aeisenbarth, line 239: sequentially

tischi:
I see the discussion moved on already, but I just wanted to mention that I think the suggestion from @jbms would have some merit: `3 nanometer would have to be specified as 3e-9 m`. I think this would be very clean and also reduce the burden on the reader libraries (as they would not need to understand all those aliases). I guess it could put a bit of burden on the writer libraries because they may need to convert `nm` to `1e-9`, but personally, from a scientific point of view, I like to try to stick to  SI units on the level of the file format.

latest/index.bs
```diff
@@ -205,25 +207,69 @@ Metadata {#metadata}
 The various `.zattrs` files throughout the above array hierarchy may contain metadata
 keys as specified below for discovering certain types of data, especially images.
 
+"axes" metadata {#axes-md}
+--------------------------
+
+"axes" describes the dimensions of a physical coordinate space. It is a list of dictionaries, where each dictionary describes an dimension (axis) and:
+- MUST contain the field "name" that gives the name for this dimension. The values MUST be unique across all "name" fields.
```
will-moore, line 214: Although the `type` of axes is discussed below, it would probably be useful to include here, e.g:
 - MUST contain the field "type" with a value of "space", "time" or "channel". See [[#multiscale-md]] for more details.

latest/index.bs
```diff
@@ -205,25 +207,69 @@ Metadata {#metadata}
...
+| `identity`    |        | identity transformation, is the default transformation and is typically not explicitly defined |
+| `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
+| `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
+
+In addition, the field "axisIndices" MAY be given to specify the subset of axes that the transformation is applied to, leaving other axes unchanged. If not given, the transformation is applied to all axes. The length of "axisIndices" MUST be equal to the dimensionality of the transformation. If "axisIndices" are not given, the dimensionality of the transformation MUST be equal to the number of dimensions of the space that the transformation is applied to.
+If given, "axisIndices" MUST be given in increasing order. It uses zero-based indexing.
```
will-moore, line 237: This seems like a lot of rules to handle the case where some axes don't need a scale.
I wonder if instead of
`"scale": [0.5, 0.5, 0.5], "axisIndices": [2, 3, 4]` we could do `"scale": [null, null, 0.5, 0.5, 0.5]`. Then we can simply say that the length of `scale` list must be equal to the number of dimensions. And we don't need to have all the `axisIndices` rules (and validation logic). It's also less code to find the scale for e.g. dimension 2.
```
scale_dim2 = transform["scale"][2]
```
compared with:
```
index = transform["axisIndices"].indexOf(2)
if index == -1:
  scale_dim2 = None
else:
  scale_dim2 = transform["scale"][index]
```

satra: @will-moore - how would you know that those are the spatial axes? while a certain ordering is recommended it is not mandatory under the spec? the spec currently decouples dimension from type of axes.

bogovicj: Agree that the above case is shorter, but the code to generate a 3d -> 3d transform would be longer with the 
`[null, null, 0.5, 0.5, 0.5]` case.  I.e. the code would have to check which dimensions are spatial, then slice the scale array appropriately.  

I don't have much of a preference between the two, honestly.  But some things are easier if the transforms always consider dimensions of the same type.

latest/index.bs
```diff
@@ -205,25 +207,69 @@ Metadata {#metadata}
...
+Each dictionary MAY contain the field "transformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
+The transformations are defined according to [[#trafo-md]]. In addition, the transformation types MUST only be `identity`, `translation` or `scale`.
+They MUST contain at most one `scale` transformation per axis that specifies the pixel size in physical units.
+It also MUST contain at most one `translation` per axis that specifies the offset from the origin in physical units.
+If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates. If "transformations" is not given, the identity transformation is assumed.
+The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while
+being compatible with the general transformation spec.
+
+It MAY contain the field "transformations", describing transformations that are applied to each resolution level.
```
will-moore, line 270: This is repeating the line above "Each dictionary MAY contain the field "transformations", which...

latest/index.bs
```diff
@@ -205,25 +207,69 @@ Metadata {#metadata}
...
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
-
-It MUST contain the field "axes", which is a list of dimension names of the axes.
-The values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
-The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
-In addition, the "axes" values MUST be repeated in the field "_ARRAY_DIMENSIONS" of all scale groups
-(i.e. groups containing arrays with the multiscale data).
-This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
+All arrays MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
+Each dictionary MAY contain the field "transformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
```
will-moore, line 262: Nice to start a new paragraph here, since it's quite a big block of text (see http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/ome/ngff/23be956/latest/index.bs)

latest/index.bs
will-moore: Ah, no - I see that this is now talking about the "multiscales" dictionary. 
Maybe start with `Each "multiscales" dictionary MAY contain the field "transformations"...`

latest/index.bs
will-moore: As below, using `Each "datasets" dictionary MAY contain the field "transformations"...` may help avoid confusion here.

latest/index.bs
jbms: You don't need `null` for scale --- you can just use 1.  And I agree that `axisIndices` just adds unnecessary complexity.

latest/index.bs
jbms: It would be good to specify precisely the mathematical formula in terms of transformed space coordinates and original coordinates.  Otherwise there is a possibility for confusion as to the precise interpretation of the scale and translation values.

latest/index.bs
bogovicj: @will-moore @jbms 

Please just keep in mind that the `axisIndices` idea was motivated by the `affine` case. i.e. to avoid having to write `[1,0,0,0,0   0,1,0,0,0   0,0,2,0,-1,  0,0,0,3,-2]` for `ctyx` for example. 

but preferring
`[2,-1,  3,-2]` with `axisIndices=[2,3]`

I hope you agree in this example, the latter is nicer.  Of course, it only matters if we need the "full affine" (i.e. if dimensions are mixed).  

I'm sympathetic to wanting to be able to get the "native resolution." of some data, in which case its good to limit the transforms to `scale` or `scale_offset`.  In which case I'm in favor of this idea.  Please consider, other more complicated use cases.  In my view, the complexity is only unnecessary when the use case is simple.

jbms: @bogovicj It is true that it is more concise to specify an affine transform that only affects a subset of the dimensions by using `axisIndices`, but I'm not sure it is actually easier for humans to understand at a glance since then you also have to mentally perform the permutation.  But I would also expect affine transforms would be primarily read and written by computers, not by humans directly, and then it is simpler to not have to support axisIndices.

bogovicj: @jbms :+1:

Mostly I'm advocating for including *some way* to describe applying transforms over subsets of data (subsets of axes).
And if it's not in this version of the spec, that's okay with me.  But if it never makes it, then we (Saalfeld lab) and others won't be able to use this spec for some of our applications, which would be a shame.

specifically - please don't make me create a 5D non-linear transform to move only the x-y axes around.

> I'm not sure it is actually easier for humans to understand at a glance then you also have to mentally perform the permutation

It helps to know that those axes are of the same type (usually spatial).  So I agree it would be awful to have to do a mental permutation, usually my "at a glance" understanding is "this affine applies to spatial dimensions only"  That more nicely matches my (John's) mental model than the 5d affine above, for example.

Actually imo, better even than `axisIndices` would be to name the axes - something you mentioned in this thread - and have these transforms apply to spaces or sub-spaces (sets of named axes).  But that's a conversation for another time and place.

I appreciate your talking through this with me.

jbms: @bogovicj Can you explain a bit more why it is so important to be able to specify `axisIndices`?

Do you imagine frequently writing out this metadata fully manually?

With the current state of this proposal, it appears that affine transforms aren't even supported --- only translation and scale.  So there it is just a matter of specifying `0` or `1` (for translation or scale, respectively) for any axes that are to be left alone --- it seems like `axisIndices` would really only provide a benefit when manually writing out the metadata  for a full affine transform (not yet part of the proposal, but presumably planned to be added).

You mentioned the specific use case of translating x and y only.  For the use case you have in mind, is the translation by an integer amount?  If so, would that use case perhaps be better addressed by support for a non-zero origin in zarr?
https://github.com/zarr-developers/zarr-specs/issues/122

For programmatic reading and writing of this metadata, a single affine transform would seem to be the most straightforward way to deal with these transformations.  I suppose implementations that don't support anything but translation and scaling might instead normalize to a single translation vector and single scale vector.  But they could just as well use an affine transform but return an error if the rotation matrix is not diagonal.   It seems to me that composing of multiple transforms, and support for things like `axisIndices`, as specified in the current proposal, could instead by handled by library functions in the program used to generate the OME-zarr metadata.

Likewise, in a graphical user interface a single affine transform would also seem to be the most straightforward way to present the transformation, as otherwise there is a lot of added complexity in displaying possibly multiple different transforms of different types, and making it clear in what order they are applied.  In Neuroglancer affine transforms are shown in their normal matrix form but with the rows and columns labeled to indicate to which source/target dimension they correspond and which column corresponds to translation.  That makes it fairly easy to understand even if you have more than 3 dimensions.

will-moore: @bogovicj Sorry, I didn't appreciate how this would work with affine transforms. I agree that your example *with* `axisIndices` looks simpler, although I don't actually understand either!

latest/index.bs
constantinpape: Thanks for the comment @jbms. I think this is a good point.

latest/index.bs
constantinpape: To quickly revive this conversation: as @bogovicj explains, we have introduced `axisIndices` to enable specifying transformations for a subset of the axes, e.g. for the use-case of affine transformations.
I think having the option of specifying subsets for transformations is crucial to have in the transformations, but I agree that `axisIndices` may not be the best solution. I see a few potential options:
1. change to using named axis for all the transformation parameters, e.g. `scale={"x": 2, "y": 2}` (as discussed above). Names must be the same as given in the axes definition. In this case we don't need `axisIndices`
2. just remove `axisIndices` from the current proposal and deal with how to address subset of axes later.
3. keep everything as is and improve how `axisIndices` are documented

My personal preference would be option 1, because I think that this would be the most elegant solution.

constantinpape:
Sorry, I missed the on-going conversations over christmas and was pretty busy in January; so only had time to catch up on this now. Thanks @will-moore for your review; I think the comments all make sense and I will address them.
But I think there is still some changes or clarification needed for the `axisIndices`, see https://github.com/ome/ngff/pull/57#discussion_r791969401.  My personal preference would be changing to named axes, which seems to me the most elegant solution. I will double check this with the OME team, given that the implementations for 0.4 are already fairly advanced.

Finishing up 0.4 will also be a point on the agenda for Thursday:  https://forum.image.sc/t/next-call-on-next-gen-bioimaging-data-tools-2022-01-27/60885 @jbms @bogovicj @satra are you able to make it to one of the calls to discuss this?

satra:
unfortunately won't be able to make either time this thu. 

1 is fine with me as well.  however, i would suggest at least to keep the thought process that affines may return and we may still need to know what the axes of the affine correspond to. thus 3 should still be in consideration. 

tl;dr: every 3d format in neuroimaging uses affines instead of individual operators, but at least pixel/voxel dimensions are stored separately as well. for example, the nifti format has support for both quaternions and affines and different target spaces (both standard and custom). more details at: https://brainder.org/2015/04/03/the-nifti-2-file-format/ (perhaps some of the fields will give inspiration as to what information should exist). given some of the discussions here, it may be useful to consider a few operations (visualization, registration, resampling) to see how the choice of the spec impacts downstream applications. one advantage of an affine is that the user doesn't have to remember the order constraint of applying transformations.

jbms:
Yes I can make the calls on Thursday --- 18-20 CET would be much preferred.

To echo the comment by @satra, one thing that I think would be very helpful to aid understanding the overall context of this proposal is one or more "user stories" that describe how this transformation metadata is initially generated/stored and how it is consumed.

In particular the question in my mind is in which cases it makes sense for the transformation to be a property of the data as opposed to a property of a particular view of the data.

constantinpape:
@satra 

> 1 is fine with me as well. however, i would suggest at least to keep the thought process that affines may return and we may still need to know what the axes of the affine correspond to. thus 3 should still be in consideration.

Thanks for the feedback. I agree that we need a way to specify subsets of axes; if we remove the `axesIndices` here we need to tackle it again for affines. But I think right now that it may be better to discuss this again when affines come up, rather than fixing a solution here that we may regret when we actually implement the affine transformation (i.e. the `axisIndices`).
Anyway, I am still thinking out loud here a bit, we will have a short preparation meeting tomorrow to see how exactly we put this on the agenda and getting feedback now is very valuable.

Thanks for also providing the link to the definition in nifti; I will try to have a closer look tomorrow.

@jbms 

> Yes I can make the calls on Thursday --- 18-20 CET would be much preferred.

Great! @joshmoore and me will be in both calls to discuss this.

> To echo the comment by @satra, one thing that I think would be very helpful to aid understanding the overall context of this proposal is one or more "user stories" that describe how this transformation metadata is initially generated/stored and how it is consumed.
> 
> In particular the question in my mind is in which cases it makes sense for the transformation to be a property of the data as opposed to a property of a particular view of the data.

I agree that this would be very helpful. For now, I have started collecting example volumes [here](https://github.com/ome/ome-ngff-prototypes). It's lacking real descriptions of what these transformations do and there's maybe a better way of compiling examples; very interested in suggestions, we could also discuss on Thursday.

jbms:
@constantinpape As far as "user stories" I had in mind textual descriptions, like:

Story 1:
1. Microscope interfacing software reads data from microscope and writes in ome-zarr format.  Axis names and units are hard coded while the scales are set based on the imaging parameters.
2. User loads data in ImageJ/Neuroglancer and sees correct axis names and scale bars.

Story 2:
1. User runs CNN on cutout from existing volume in ome-zarr format and writes result as a new OME-Zarr volume, using a `translation` transformation to indicate the cutout location.
2. User loads CNN output and original volume in ImageJ/Neuroglancer and sees output correctly aligned to original full-size volume.

Story 3:
1. User runs affine registration between existing volumes A and B in ome-zarr format.  After computing affine transform, software modifies metadata of volume B in place to indicate the affine transform.
2. User loads data volumes A and B in Neuroglancer and sees them correctly aligned.

These could be fleshed out further to help evaluate certain design choices, like `axisIndices`.

axtimwalde:
Story 4:

1. User performs non-rigid registration between existing volumes A (a canonical atlas) and B in ome-zarr format. BigWarp updates transformation meta-data in volume B, the non-rigid transformation is stored as a displacement field in ome-zarr format.
2. User loads data volumes A and B in BigDataViewer and sees them correctly aligned.
3. User overlays segmentations performed on original volume B (Story 2) and sees them correctly aligned.

jbms:
@axtimwalde Under the current proposal, I guess it would also be necessary to copy the transformation metadata that was written to volume B to the segmentation volume as well.

To me a key question is how often we want to say "when using volume B, always use this transformation" as opposed to having some way of storing "here is a transformation from volume A coordinate space to volume B coordinate space" and then creating some sort of "view" of volume B with the transformation applied.

axtimwalde:
Great points that also apply to Story 1, 2, and 3, I do not see a clear boundary between these scenarios when it comes to the transformation spec and would therefore keep them as compatible as possible.  In my understanding all transformed volumes are views.  The storage location for that displacement field can be an absolute URL and does not have to exist in the same container as volume A and B, so it can be re-used.  It is also likely that volume A is in a different container than volume B.  We may want to add a more every-day story than the BigWarp scenario for non-rigid transformations like ultra-sound scanners or radars?

jbms:
Yeah, to be clear I wasn't really drawing a distinction between affine vs non-affine transformations here, but rather just getting at the idea that the transformation could be decoupled from the volume itself.

In my view Story 2 is a little different because there the cutout location is known upon creation and the "transformation" isn't going to change.

For an affine or non-rigid transform you may still wish to work with the untransformed volume for various reasons, and there may be more than one possible transform.

Additionally, the original volume already be referenced elsewhere before you have computed the affine or non-rigid transformation, and in that case you don't necessarily want to change the meaning of those existing references by modifying its metadata.

axtimwalde:
Yes, transformations can be decoupled from the volume. But I also think that there is not clear difference between a cutout offset or pixel scaling factor and a registration transformation.  E.g. imagine an SEM volume stored as a 3D dataset in ome-zarr format.  The microscope will most likely export data such as the nominal pixel spacing in x and y as well as a nominal spacing in z. These values can be used to display a view of the array data, but we also know that this information is not physically correct and therefore have to perform alignment and z-thickness compensation, the combination of the nominal values and the alignment transformations can now be used to display another view of the array data.  Mapping the result into a canonical atlas space is yet another view of the data.  Where do those transformations go?  Depends on the application.  Since we can store arbitrary meta-data in zarr/N5/HDF5 and friends, it is perfectly fine to for that information to be associated with a volume, or with a multi-scale group, or somewhere else.  Were you suggesting to wrap all transformation meta-data into named view-tags?  I find that appealing.  It brings us back to groups (sequence) of transformations (@bogovicj) had this in our original proposal.  Such groups could be used by named references to create compositions and subsets of transformations.  In the above FIB-SEM scenario, we would most likely have the groups "microscope", "alignment", "atlas" that can be composed at will.  I find this very appealing.

satra:
Story 5. slightly different registration problem.

1. User collects multiple overlapping/non-overlapping slices and stores each in ome-zarr format with appropriate transformations
2. User collates all slices into a single ome-zarr volume (registering neighboring slices, joining the data, or introducing zeros/nans as necessary).

jbms:
I agree that the pixel size from the microscope may be unreliable and in that sense is similar to a registration transformation.  But the cutout offsets would most naturally be expressed in units of voxels (https://github.com/zarr-developers/zarr-specs/issues/122) and then would be exact and immutable and independent of any other transformations.

As far as other transformations, though, I don't have a concrete proposal for exactly how it might work, but what I'm suggesting is to be able to specify named transformation objects independent of any particular volume, and optionally identify the source and target coordinate spaces with some string identifier.  And then similarly define named "view" objects that reference a transformation, or sequence of transformations, and an array or multiscale array.  Array or multiscale arrays could also optionally specify a named coordinate space, and then viewers could potentially automatically suggest compatible transformations that could be applied.

I think it could be useful to distinguish between "arrays" that are indexed by discrete coordinates and "views" that are indexed by continuous coordinates.  Otherwise if we reference a given "multiscale" or "array" that has transformations applied, we would have to additionally specify if we are referring to the original discrete coordinate space or the transformed continuous coordinate space.

tischi:
> As far as other transformations, though, I don't have a concrete proposal for exactly how it might work, but what I'm suggesting is to be able to specify named transformation objects independent of any particular volume, and optionally identify the source and target coordinate spaces with some string identifier. And then similarly define named "view" objects that reference a transformation, or sequence of transformations, and an array or multiscale array. 

This sounds to me very similar to what we currently do in MoBIE. Here is [an example](https://github.com/mobie/clem-example-project/blob/445f3c24788f891d260ffa902c5e1ac4526e1ac9/data/hela/dataset.json#L270) of a `view` of many images where in addition to display settings we also specify the [transformations](https://github.com/mobie/clem-example-project/blob/445f3c24788f891d260ffa902c5e1ac4526e1ac9/data/hela/dataset.json#L649) for that particular view. Here are the [current specs](https://mobie.github.io/specs/mobie_spec.html).

So, in that sense the transformations are decoupled from the volume itself and one can have different views on the same volume with different transformations. However, we currently require that the transformations in the `views` are from physical space to physical space. That means that currently the "volume" must contain one immutable information of how to get from data space (voxels) to physical space (`x`,`y`,`z`). We thought that is convenient because (i) one could just look at a volume without having to specify a particular view and (ii) it is easier to just work in physical space in the views without having to worry about the more complex issue of how to get from data space to physical space. But I guess a concept of a "default view"  and "views of views" could achieve something similar.

One other question: I am confused about the concept of a voxel size. Already when one rotates the data a little bit the voxel grid does not align anymore with the physical coordinate system and I am confused how to define voxel size in such cases?!

aeisenbarth:
The concept of views could also help for what we are doing:

We have a workflow doing segmentations and registrations, and during this process we incrementally update the Zarr metadata to store the transformations. We also may rerun the workflow with different sets of parameters, leading to versioned results that we want to be able to view and compare.

If we can only have a single transformation strongly coupled with an image, we need to create a copy of the Zarr for not overwriting the original. With views we would be able define a view for each set of parameters and tell which transformations belong to that view and which segmentations.

latest/index.bs
bogovicj: Thanks for the nice summary @constantinpape .

I like option (1) as well.

> please don't make me create a 5D non-linear transform to move only the x-y axes around.

I need to clarify this, so let me describe a simpler example as a case study. There is a fairly common use case in which we need to apply a nonlinear 2D transform to every plane of a 3D volume (e.g., lens correction).

This is not covered an offset or an affine. We like the `axisIndices` (or similar) idea here because it lets us store the transform in a straightforward way: as a 2D displacement field (a 3D volume: `[2,X,Y]`, where the 2 dimension is a vector describing the displacements at `[:,x,y]`).

Without `axisIndices` (or some way to communicate "this transform is applied over a subset of the coordinates), then what is the right way to store the transform / communicate this idea? I guess by making it a 3D displacement field instead? So a `[3,X,Y,Z]` array where the `[2,:,:,:]` is full of zeros, and `[0:1,:,:,z]` all contain identical data. That's wasteful.

I'm not sure how to deal with this use case without being explicit about this "subset idea".

bogovicj:
Happy to see this conversation and the interest in views.

@jbms 
> I'm suggesting is to be able to specify named transformation objects independent of any particular volume, and optionally identify the source and target coordinate spaces with some string identifier.

Yes, I want this too. and so does @axtimwalde wro write:

>  In the above FIB-SEM scenario, we would most likely have the groups "microscope", "alignment", "atlas" that can be composed at will. I find this very appealing.

 I'll write something up for tomorrow and would love to discuss.

sbesson:
My vote also goes towards a simplified version of the `scale` transforms as expressed in https://github.com/ome/ngff/pull/57#discussion_r791969401 (option 1). As this seems to be the remaining point of contention in this PR, this would help moving towards a releasable state of the specification. The discussion around affine transformations can continue separately and lead to the reintroduction of the `axisIndices` concept as needed.

latest/index.bs
jbms: I would absolutely agree that having to store extra zeros would be unacceptable, and agree that something like axisIndices is very useful in the flow field case.

You could accomplish the same thing, in a somewhat more general way, by allowing to specify an affine transform that is applied to every flow vector.  However, I'm not sure if that extra generality would actually be helpful.

constantinpape:
A quick wrap up before the meeting tomorrow (summary from a quick pre-meeting with @sbesson, @will-moore, @joshmoore):
- strategy for 0.4: merge this PR as is (just with some changes to address detail review comments); propose changes to simplify `scale` and `offset` to `dicts` in a separate PR; once that one is merged we can release 0.4. We will discuss this tomorrow and go ahead with this plan to make the release in the next few days unless there is any major veto.
- it's great to see the discussions and examples for more complex transforms and their application. We decided a while back not to include this in 0.4 to reduce the complexity for the 0.4 release. But this will be very helpful for extending the transform spec and I will make sure that we transfer the "user stories" to a separate issue when merging this PR so that we don't loose sight of this. A few comments w.r.t. extending the transform spec:
   - `transforms` is introduced as a separate object in this PR so that it's possible to refer to it not only in the `multiscales` spec but also later when `collections` or `views` are introduced, along the lines of many ideas brought up in the user stories.
   - once we introduce affine or more complex transforms we will need a mechanism to label the subset of dimensions that the transforms are applied to. `axisIndices` were the initial approach for this; but for the current `scale` and `offset` transforms this approach is an over-engieneered and using a dict is much simpler. So we will instead need to solve this problem once affine transforms are introduced and can either resort to `axisIndices` then or find  a more elegant solution. (RIght now I think that using the axis names instead of numerical indices would be better; but we should not open this discussion here).

latest/index.bs
constantinpape: Thanks for catching this @will-moore. I think `type` got lost here in some edits. I have added it again.

latest/index.bs
constantinpape: I have adapted the multiscales definition a bit, so that each paragraph clearly identifies which object is talked about in the beginniing.

latest/index.bs
will-moore: Just trying to implement the handling of scale and translation in `napari-ome-zarr`. Napari supports `scale` and `translate` metadata (see [add_image](https://napari.org/docs/0.3.8/api/napari.components.html#napari.components.add_layers_mixin.AddLayersMixin.add_image)). Napari doesn't specify an order, but it seems by testing that the translate values are in real-world values (applied *after* scaling).
Do we want to specify the same for translation in this NGFF spec? Or does this depend on the ordering of `translation` and `scale`? If the `translation` comes before `scale` then it's in pixels. If after then it's in real space?
Also, just a thought with naming. In napari it's `scale` and `translate` (both verbs) but we have `scale` and `translation`. Do we want to use `translate` instead?

bogovicj:
@constantinpape 

A quick point to think about too (we'll discuss in the call today, but wanted to write it down here next to everything else).

The spec allows for binary storage the transform data at some path in the container.  Using the dict to specify axes doesn't work in that case. So we need either to
* specify the axes those scale / translation refer to in a different way 
* require that the length of scale / translation match the dimensionality of the array

constantinpape:
Good point @bogovicj. I forgot that we allow binary data here already. Let's discuss during the meeting.

constantinpape:
I will merge this PR now and finalize the axes and initial transformations in a separate PR:
- changes for new PR
  - remove `axisIndices`, require that `scale` and `translation` have the same length as the axes (we discussed this in the meeting; the transformations on a subset of axes will be introduced in the next proposal on advanced transformations by @bogovicj)
   - rename `transformations` -> `coordinateTransformations` to clarify the scope
- other points:
  - move user stories to separate issue (I am on it)
  - address review comment by @will-moore about order of scale and translation (@bogovicj volunteered to write something up on this; let's do this in the new PR)
  - address comment by @jbms to give mathematical definition for the transformation

constantinpape:
Final house-keeping:
- the user stories have been moved to #84 
- the follow up PR is in #85 
- the plan for finalizing v0.4 in #86 

A big thank you to everyone who commented here and helped to bring this into shape!

PR #63, opened 2021-10-01
[Draft] Add transformation spec

constantinpape:
**UPDATE**

I have ported most of the content from here into #57, except for the more "complex" transformations (affines, defeormation fields, position fields).  The next version (0.4) will only support the simpler transformations and we will add the more complex transformations in a later version.

For discussions about v0.4, please refer to #57; feel free to keep discussing affines or deformation fields in this draft.

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -219,11 +219,14 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 
 It MUST contain the field "axes", which is a list of dimension names of the axes.
-The values MUST be unique and one of `{"t", "c", "z", "y", "x"}`.
-The number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
-In addition, the "axes" values MUST be repeated in the field "_ARRAY_DIMENSIONS" of all scale groups
-(i.e. groups containing arrays with the multiscale data).
+The values provide axes labels and MUST be unique, the number of values MUST be the same as the number of dimensions of the arrays corresponding to this image.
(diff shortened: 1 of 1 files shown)


constantinpape:
# First draft for separate transformation metadata

Adapted from https://github.com/saalfeldlab/n5-ij/wiki/Transformation-spec-proposal-for-NGFF, mixing parts of Proposal A and Proposal C.
Design choice and my reasoning behind it are below; comments / critique are welcome.

## Design Choice

Transformations are specified for each scale array in `multiscales:datasets`. They are not repeated in the "<SCALE_ARAAY>/.zmetadata" (same is true for axes in #57). Note that there have been discussions about duplicating or moving the transformation and axes metadata there. I would like to postpone this discussion and use #57 and this PR to define the metadata format first.

## Differences to the proposal by John and Stephan

I have deviated from the proposal in several places, either for simplicity or because I did not understand some part of the proposal. Of course this reflects my subjective preferences and is very much up to discussion:
- I have left out `sequence`, instead the transformations are specified in a list in `multiscales:dataset`.
- For `translation`, `scale` and `affine` I interpret `oneOf(path, paramField)` as exact (i.e. exactly one of those must be specified).
- For `d_field` and `p_field` I only allow internal paths for storing the fields to keep the spec simpler and make the container stand-alone.
- For `d_field` and `p_field` I leave out the parameter `interleaved`, as I don't its purpose (also not mentioned in the description in the proposal).
- I have left out `inverse_of`, because I am not sure about the exact implementation. Would this only be valid for invertible transformations? How do we know if they are invertible? E.g. `inverse_of({"type": "d_field", "path": "/my_deformations"})`.
- I have left out `outputIndexes` from `Proposal C` ("inputIndexes" correspond to "axisIndices") for the sake of simplicity. I.e. axes reordering is currently not possible. (I am not sure how we would do this in a software-interoperable way.)

PR #67, opened 2021-10-18
collections spec

will-moore:
This follows discussion of a Collections spec at #31

This contains the core of what was agreed in the issue above and omits `groups`, `layouts`, `transforms` etc.

This PR also includes the validation logic for json_schema within a new `0.4` directory (although the version for collections may be later). The test allows for different schemas to be used to validating samples under `0.4/examples/image/` and `0.4/examples/collection/`. The samples have been copied over from `0.3` with the expectation that these can be updated as the 0.4 spec PRs are merged.


joshmoore:
#69 is now merged with various investigations into validation & subclassing options for collections.

PR #70, opened 2021-10-29
Add camelCase naming style rule

### LINKED ISSUE #54: Spec naming style
When looking to add new keys to the spec I wonder how to name e.g. [position_x](https://github.com/ome/ngff/issues/31#issuecomment-910260493) I wonder what the naming style should be e.g.

- position_x
- positionX
- position-x
- positionx

Looking at the current spec, we have all of these styles there already, reflecting the origin of various terms (mea culpa):
 - `image-label`
 - `defaultT` - Rendering definition (omero). to be replaced by rendering spec?
 - `maximumfieldcount` - HCS, to be replace by Collections spec
 - `field_count` - Also HCS.

So, which of these do we prefer? Python style, Java style or something else?

will-moore:
Fixes #54 

Adding a note on naming convention, as agreed at #54.

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -543,6 +543,13 @@ the last two fields of view were part of the second acquisition.
   }
 ```
 
+Specification naming style {#naming-style}
+==========================================
+
+Multi-word keys in this specification should use the `camelCase` style.
+NB: some parts of the specification don't obey this convention as they
+were added before this was adopted, but they should be updated in due course.
+
 Implementations {#implementations}
 ==================================
 


sbesson:
Looking around the currently open PRs, the primary one that would be affected by this style is probably https://github.com/ome/ngff/pull/63https://github.com/ome/ngff/pull/63 e.g. `d_field`  /cc @constantinpape 

No objection to defaulting to camelCase from my side either especially if we can enforce long-term consistency.

constantinpape (approved):

constantinpape:
I am fully in favor of this!

> Looking around the currently open PRs, the primary one that would be affected by this style is probably #63 e.g. `d_field` /cc @constantinpape

At this point, #63 is only a draft to discuss ideas; when we move forward we will probably rename this to `displacementField`.

joshmoore:
Ok. Merging and assuming this will go live in 0.4.

Issue #65, opened 2021-10-05
Storing tables


tischi:
An outcome of [this hackathon](https://forum.image.sc/t/ome-ngff-spatial-omics-hackathon/57337/26) was that we would like to store tabular data in ome-zarr.

I wanted to ask whether we should consider that whatever we store can be easily mapped onto a csv file. Meaning that fromCSV and toCSV should work smoothly such that other software that can work with tables can be interoperable with the tabular content of the ome-zarr.

What do you think?

joshmoore:
toCSV will be easy enough. fromCSV will for many (if not most?) cases require extra metadata. There are a number of attempts to do so, e.g., https://specs.frictionlessdata.io/data-package/

unidesigner:
Just out of curiosity, what is the reason of wanting to store tabular data in zarr, v.s. using some existing, optimized data formats, like Avro, Parquet, Sqlite etc. ?

tischi:
I think the point was to be able to efficiently load chunks of the table from an object store. I don't have enough technological knowledge to judge if any of the existing solutions would do the trick. Do you know?

constantinpape:
> I wanted to ask whether we should consider that whatever we store can be easily mapped onto a csv file. Meaning that fromCSV and toCSV should work smoothly such that other software that can work with tables can be interoperable with the tabular content of the ome-zarr.

I think that compatibility with csv is desirable, but I am not sure how much can be done about this on the spec level.
I see this as more of a software than data standard question.

> Just out of curiosity, what is the reason of wanting to store tabular data in zarr,

I would say the main reason is to provide all relevant data in the same data format and container.
Also note that [AnnData](https://anndata.readthedocs.io/en/latest/anndata.AnnData.html), which the proposal is based on, is using zarr as storage already.

tischi:
> Also note that AnnData, which the proposal is based on, is using zarr as storage already.

My worry was that AnnData is much richer than a simple table and thus it may be difficult to map it onto a "simple table"? For example, both in Fiji and Napari there are ways to display a table. Do you think that one could also display AnnData in a "simple table viewer"?

constantinpape:
> My worry was that AnnData is much richer than a simple table

I don't think that AnnData is much richer than a simple table; at least not the subset that we are discussing here.

> Do you think that one could also display AnnData in a "simple table viewer"?

[...]

PR #85, opened 2022-01-27
Finalize axes & initial transformation

constantinpape:
Follow up on #57:
- rename `transformations` to `coordinateTransformations`, after discussion in todays meeting with @bogovicj @jbms @lassoan and others
- remove `axisIndices` for simplification. Transformations for a subset of axes will be tackled in future work.

Other points from #57 that should still be addressed:
- order of `scale` and `translation`, raised by @will-moore. We actually restrict this to first `scale`, then `translation`; I will leave a code comment to point this out.
- Give mathematical definition of the current transformation, raised by @jbms

Finally, we may want to remove the transformations applied to all transformations levels; I think that this introduces unnecessary complexity without offering much functionality. (Edit: on second thought we may want to keep it because this will be very useful once we have coordinateTransformations for subsets of axes)

@joshmoore @sbesson @will-moore @bogovicj please review with current implementation in ome-zarr-py / ome-napari, vizarr and the upcoming advanced transformation proposal in mind. Review from everyone else is of course also welcome, but please keep in mind that we don't want to introduce fundamental changes at this point for v0.4; these discussions can happen for the upcoming advanced transformation proposal that will be spearheaded by @bogovicj.

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -263,8 +263,9 @@ to the current zarr group. The "path"s MUST be ordered from largest (i.e. highes
 Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
 Each dictionary MAY contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
 The transformations are defined according to [[#trafo-md]]. The transformation types MUST only be `identity`, `translation` or `scale`.
-They MUST contain at most one `scale` transformation per axis that specifies the pixel size in physical units.
-It also MUST contain at most one `translation` per axis that specifies the offset from the origin in physical units.
+They MUST contain at most one `scale` transformation that specifies the pixel size in physical units or time duration.
+It also MUST contain at most one `translation` that specifies the offset from the origin in physical units.
+The length of the `scale` and/or `translation` array MUST be the same as the length of "axes".
 If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates. If "coordinateTransformations" is not given, the identity transformation is assumed.
 The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while
 being compatible with the general transformation spec.
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -237,7 +237,7 @@ Additional fields for the entry depend on "type" and are defined by the column `
 | `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 | `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 
-The transformations in the list are applied sequentally and in order.
+The transformations in the list are applied sequentially and in order.
 
 
 "multiscales" metadata {#multiscale-md}
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -261,14 +261,12 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 
 Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
-Each dictionary MAY contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
+Each dictionary MUST contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
 The transformations are defined according to [[#trafo-md]]. The transformation types MUST only be `identity`, `translation` or `scale`.
-They MUST contain at most one `scale` transformation that specifies the pixel size in physical units or time duration.
-It also MUST contain at most one `translation` that specifies the offset from the origin in physical units.
-The length of the `scale` and/or `translation` array MUST be the same as the length of "axes".
-If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates. If "coordinateTransformations" is not given, the identity transformation is assumed.
-The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while
-being compatible with the general transformation spec.
+They MUST contain exactly one `scale` transformation that specifies the pixel size in physical units or time duration. If scaling information is not available or applicable for one of the axes set it to 1.
+It MAY contain exactly one `translation` that specifies the offset from the origin in physical units. If `translation` is given it must be listed after `scale` to ensure that it is given in physical coordinates.
+The length of the `scale` and `translation` array MUST be the same as the length of "axes".
+The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while being compatible with the general transformation spec.
 
 Each "multiscales" dictionary MAY contain the field "coordinateTransformations", describing transformations that are applied to each resolution level.
 The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:coordinateTransformations".
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -264,7 +264,7 @@ Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT 
 Each dictionary MUST contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
 The transformations are defined according to [[#trafo-md]]. The transformation MUST only be of type `translation` or `scale`.
 They MUST contain exactly one `scale` transformation that specifies the pixel size in physical units or time duration. If scaling information is not available or applicable for one of the axes set it to 1.
-It MAY contain exactly one `translation` that specifies the offset from the origin in physical units. If `translation` is given it must be listed after `scale` to ensure that it is given in physical coordinates.
+It MAY contain exactly one `translation` that specifies the offset from the origin in physical units. If `translation` is given it MUST be listed after `scale` to ensure that it is given in physical coordinates.
 The length of the `scale` and `translation` array MUST be the same as the length of "axes".
 The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while being compatible with the general transformation spec.
 
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -262,7 +262,7 @@ to the current zarr group. The "path"s MUST be ordered from largest (i.e. highes
 
 Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
 Each dictionary MUST contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
-The transformations are defined according to [[#trafo-md]]. The transformation types MUST only be `translation` or `scale`.
+The transformations are defined according to [[#trafo-md]]. The transformation MUST only be of type `translation` or `scale`.
 They MUST contain exactly one `scale` transformation that specifies the pixel size in physical units or time duration. If scaling information is not available or applicable for one of the axes set it to 1.
 It MAY contain exactly one `translation` that specifies the offset from the origin in physical units. If `translation` is given it must be listed after `scale` to ensure that it is given in physical coordinates.
 The length of the `scale` and `translation` array MUST be the same as the length of "axes".
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -222,10 +222,11 @@ keys as specified below for discovering certain types of data, especially images
 If part of [[#multiscale-md]], the length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
 
 
-"transformations" metadata {#trafo-md}
+"coordinateTransformations" metadata {#trafo-md}
 -------------------------------------
 
-"transformations" describes a series of transformations, e.g. to map discrete data space of an array to the corresponding physical space.
+"coordinateTransformations" describe a series of transformations that map between two coordinate spaces (defined by "axes").
+For example, to map a discrete data space of an array to the corresponding physical space.
 It is a list of dictionaries. Each entry describes a single transformation and MUST contain the field "type".
 The value of "type" MUST be one of the elements of the `type` column in the table below.
 Additional fields for the entry depend on "type" and are defined by the column `fields`.
@@ -263,16 +264,16 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 
 Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
-Each dictionary MAY contain the field "transformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
+Each dictionary MAY contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
 The transformations are defined according to [[#trafo-md]]. In addition, the transformation types MUST only be `identity`, `translation` or `scale`.
 They MUST contain at most one `scale` transformation per axis that specifies the pixel size in physical units.
 It also MUST contain at most one `translation` per axis that specifies the offset from the origin in physical units.
-If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates. If "transformations" is not given, the identity transformation is assumed.
+If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates. If "coordinateTransformations" is not given, the identity transformation is assumed.
 The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while
 being compatible with the general transformation spec.
 
-Each "multiscales" dictionary MAY contain the field "transformations", describing transformations that are applied to each resolution level.
-The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:transformations".
+Each "multiscales" dictionary MAY contain the field "coordinateTransformations", describing transformations that are applied to each resolution level.
+The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:coordinateTransformations".
 These transformations are applied after the per resolution level transformations specified in "datasets". They can for example be used to specify the `scale` for a dimension that is the same for all resolutions.
 
 Each "multiscales" dictionary SHOULD contain the field "name". It SHOULD contain the field "version", which indicates the version of the multiscale metadata of this image (current version is 0.4).
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -237,9 +237,6 @@ Additional fields for the entry depend on "type" and are defined by the column `
 | `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 | `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 
-In addition, the field "axisIndices" MAY be given to specify the subset of axes that the transformation is applied to, leaving other axes unchanged. If not given, the transformation is applied to all axes. The length of "axisIndices" MUST be equal to the dimensionality of the transformation. If "axisIndices" are not given, the dimensionality of the transformation MUST be equal to the number of dimensions of the space that the transformation is applied to.
-If given, "axisIndices" MUST be given in increasing order. It uses zero-based indexing.
-
 The transformations in the list are applied sequentally and in order.
 
 
@@ -265,7 +262,7 @@ to the current zarr group. The "path"s MUST be ordered from largest (i.e. highes
 
 Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
 Each dictionary MAY contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
-The transformations are defined according to [[#trafo-md]]. In addition, the transformation types MUST only be `identity`, `translation` or `scale`.
+The transformations are defined according to [[#trafo-md]]. The transformation types MUST only be `identity`, `translation` or `scale`.
 They MUST contain at most one `scale` transformation per axis that specifies the pixel size in physical units.
 It also MUST contain at most one `translation` per axis that specifies the offset from the origin in physical units.
 If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates. If "coordinateTransformations" is not given, the identity transformation is assumed.
@@ -297,18 +294,18 @@ It SHOULD contain the field "metadata", which contains a dictionary with additio
             "datasets": [
                 {
                     "path": "0",
-                    "transformations": [{"type": "scale", "scale": [0.5, 0.5, 0.5], "axisIndices": [2, 3, 4]}]  # the voxel size for the first scale level (0.5 micrometer)
+                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 0.5, 0.5, 0.5]}]  # the voxel size for the first scale level (0.5 micrometer)
                 }
                 {
                     "path": "1",
-                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 1.0], "axisIndices": [2, 3, 4]}]  # the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
+                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 1.0, 1.0, 1.0]}]  # the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
                 },
                 {
                     "path": "2",
-                    "transformations": [{"type": "scale", "scale": [2.0, 2.0, 2.0], "axisIndices": [2, 3, 4]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
+                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 2.0, 2.0, 2.0]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
                 }
             ],
-            "transformations": [{"type": "scale", "scale": [0.1], "axisIndices": [0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
+            "transformations": [{"type": "scale", "scale": [0.1, 1.0, 1.0, 1.0, 1.0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
             "type": "gaussian",
             "metadata": {                                       # the fields in metadata depend on the downscaling implementation
                 "method": "skimage.transform.pyramid_gaussian", # here, the paramters passed to the skimage function are given
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -295,18 +295,18 @@ It SHOULD contain the field "metadata", which contains a dictionary with additio
             "datasets": [
                 {
                     "path": "0",
-                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 0.5, 0.5, 0.5]}]  # the voxel size for the first scale level (0.5 micrometer)
+                    "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 0.5, 0.5, 0.5]}]  # the voxel size for the first scale level (0.5 micrometer)
                 }
                 {
                     "path": "1",
-                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 1.0, 1.0, 1.0]}]  # the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
+                    "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 1.0, 1.0, 1.0]}]  # the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
                 },
                 {
                     "path": "2",
-                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 2.0, 2.0, 2.0]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
+                    "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 2.0, 2.0, 2.0]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
                 }
             ],
-            "transformations": [{"type": "scale", "scale": [0.1, 1.0, 1.0, 1.0, 1.0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
+            "coordinateTransformations": [{"type": "scale", "scale": [0.1, 1.0, 1.0, 1.0, 1.0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
             "type": "gaussian",
             "metadata": {                                       # the fields in metadata depend on the downscaling implementation
                 "method": "skimage.transform.pyramid_gaussian", # here, the paramters passed to the skimage function are given
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -262,15 +262,15 @@ to the current zarr group. The "path"s MUST be ordered from largest (i.e. highes
 
 Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
 Each dictionary MUST contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
-The transformations are defined according to [[#trafo-md]]. The transformation types MUST only be `identity`, `translation` or `scale`.
+The transformations are defined according to [[#trafo-md]]. The transformation types MUST only be `translation` or `scale`.
 They MUST contain exactly one `scale` transformation that specifies the pixel size in physical units or time duration. If scaling information is not available or applicable for one of the axes set it to 1.
 It MAY contain exactly one `translation` that specifies the offset from the origin in physical units. If `translation` is given it must be listed after `scale` to ensure that it is given in physical coordinates.
 The length of the `scale` and `translation` array MUST be the same as the length of "axes".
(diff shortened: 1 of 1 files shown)


latest/index.bs
```diff
@@ -263,16 +261,17 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
...
-If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates. If "transformations" is not given, the identity transformation is assumed.
+Each dictionary MAY contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
+The transformations are defined according to [[#trafo-md]]. The transformation types MUST only be `identity`, `translation` or `scale`.
+They MUST contain at most one `scale` transformation that specifies the pixel size in physical units or time duration.
+It also MUST contain at most one `translation` that specifies the offset from the origin in physical units.
+The length of the `scale` and/or `translation` array MUST be the same as the length of "axes".
+If both `scale` and `translation` are given `translation` must be listed after `scale` to ensure that it is given in physical coordinates. If "coordinateTransformations" is not given, the identity transformation is assumed.
```
constantinpape, line 269: @will-moore we actually restrict the order of scale and translation here to avoid the ambiguity discussed earlier. I suggest we leave this as is for v0.4. (cc @bogovicj, for now we probably don't need a longer explanation on transformation order; but I am sure this will become relevant for the advanced transformation proposal ;))

latest/index.bs
```diff
@@ -263,16 +261,17 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
...
 The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while
 being compatible with the general transformation spec.
 
-Each "multiscales" dictionary MAY contain the field "transformations", describing transformations that are applied to each resolution level.
-The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:transformations".
+Each "multiscales" dictionary MAY contain the field "coordinateTransformations", describing transformations that are applied to each resolution level.
```
constantinpape, line 273: I suggest we remove this 'global' transformation in order to keep the proposal simple. (and this additional transformation would not make the spec more expressive)

latest/index.bs
```diff
@@ -296,18 +295,18 @@ It SHOULD contain the field "metadata", which contains a dictionary with additio
...
+                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 1.0, 1.0, 1.0]}]  # the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
                 },
                 {
                     "path": "2",
-                    "transformations": [{"type": "scale", "scale": [2.0, 2.0, 2.0], "axisIndices": [2, 3, 4]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
+                    "transformations": [{"type": "scale", "scale": [1.0, 1.0, 2.0, 2.0, 2.0]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
                 }
             ],
-            "transformations": [{"type": "scale", "scale": [0.1], "axisIndices": [0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
+            "transformations": [{"type": "scale", "scale": [0.1, 1.0, 1.0, 1.0, 1.0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
```
constantinpape, line 309: If we decide to drop the 'global' transformation we need to remove it here as well and move the 0.1 into the per scale transforms.

latest/index.bs
```diff
@@ -263,16 +261,17 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 
 Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
-Each dictionary MAY contain the field "transformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
-The transformations are defined according to [[#trafo-md]]. In addition, the transformation types MUST only be `identity`, `translation` or `scale`.
-They MUST contain at most one `scale` transformation per axis that specifies the pixel size in physical units.
```
constantinpape, line 268: The "per axis" should be kept if we keep the 'global' transformation.

latest/index.bs
bogovicj: Sounds good, 
As you say I'll still write the explanation on order that I was planning on, but indeed - now you don't need it.
Thanks @constantinpape !

latest/index.bs
bogovicj: What are the right assumptions to make if the "multiscales" does not contain a `coordinateTransformations`?
Assume downsampled by 2 in all dimensions always? infer from dataset sizes?

jbms: I would really hope the answer is "fail".

constantinpape: > What are the right assumptions to make if the "multiscales" does not contain a coordinateTransformations?
Assume downsampled by 2 in all dimensions always? infer from dataset sizes?

Good point. In the previous versions we have infered from the dataset size because this was the only viable option.
But that was a very hacky solution, so I agree with @jbms that we can also say "Each multiscales dictionary MUST contain ..." instead of "... MAY contain ...". Any thoughts @will-moore @joshmoore @sbesson?

will-moore: In `napari` and `vizarr` I don't know a way to use transformations on each datasets level - I just pass in the pyramid and they make the assumption that each level covers the same area. In fact, `vizarr` even assumes the scaling factor is 2 between each level. 
I can only use a single `scale` or `translate` transform - I don't know if it makes a difference whether that is on the first `datasets` dict or on the `multiscales` dict.
So I think that it is a valid default behaviour to infer from the size of each level, and if you want the default behaviour then I don't think we MUST need `coordinateTransformations` on every datasets level or even on the `multiscales` dict.

will-moore:
While we can still rename things, can we consider renaming `translation` to `translate` (as I suggested on #57) since it is more consistent with `scale` (both verbs) and it matches what `napari` uses.
If `translation` is a convention elsewhere then that's OK, but it's always easier to avoid divergence if possible.

latest/index.bs
jbms: It seems to me that whatever default is chosen, if it is allowed that the downsampling factors and offsets be left unspecified, it will be non-obvious what the default is, at least to a significant fraction of users, which will lead to users either guessing wrong, or having to look up the spec every time they encounter an unspecified trnasformation.  I would suppose that in the case of an xy dataset where the dimensions are powers of 2, you will probably guess correctly, but if the sizes are not exactly divisible by 2, or there are more than two dimensions, there is a lot of room for guessing wrong.  Even worse than users guessing wrong, there is also a risk of implementation divergence.  If the downsampling factors and offsets are specified explicitly, that problem is avoided, and it is not very difficult to specify the factors explicitly when writing.

Regarding the issue with napari and vizarr, if the software that is reading the format only supports certain combinations of downsampling factors and offsets, it can just return an error if it receives an unsupported transformation.

d-v-b: +1 for requiring a `coordinateTransformation` for ever downsampled image. _Downsampled images are just images,_ so they warrant the same precise metadata as regular images. 

Inferring the coordinate grid of a downsampled image simply from its shape (or the ratio of its dimensions relative to some other image) is very brittle and should be avoided, given that we can just explicitly define the coordinate grid for each downsampled image.

will-moore: OK, so just to clarify. The `coordinateTransformation` `MUST` be on every `datasets` dict but `MAY` also be on the `multiscales` dict?
And if I don't have any pixel-size info or translation, then the first `datasets` will have a no-op transformations list. (e.g. `"coordinateTransformations": []`)? Or can we omit the `coordinateTransformations` for the first `dataset`?

constantinpape: Fully agree with @jbms and @d-v-b here; for `napari` / `vizarr` we can for now throw an error if an unavailable transformation is passed. For clarification @will-moore:

> The coordinateTransformation MUST be on every datasets

Yes, I will introduce this constraint later and additionally require that each of these `coordinateTransformations` MUST contain a `scale` transform (and optionally a `translation`). This is the only way to have an unambiguous way to specify the downscaling factors. And for the simplest use case of a single image without resolution information, one can use `scale=[1.0, 1.0 , 1.0]` in the first dataset.

> `MAY` also be on the `multiscales` dict?

This is still up for discussion, and was actually the point I raised initially; I will first change the dataset `coordinateTransform` spec, where we have strong agreement and then open the discussion about this again and also give the pros/cons from my point of view.

> And if I don't have any pixel-size info or translation, then the first `datasets` will have a no-op transformations list. (e.g. `"coordinateTransformations": []`)? Or can we omit the `coordinateTransformations` for the first `dataset`?

In this case, use `scale=[1.0, 1.0, 1.0]`. (More precisely `scale=[1.0] * number_of_axes` (python notation)).

constantinpape: Dataset transformations are mandatory now and need to contain scale 0241254. (Please let me know if you have very big concerns about this @joshmoore @sbesson or @will-moore, but from a developer perspective this is a much better choice than leaving this implicit; and of course comment if this still needs to be clarified further in the spec.)

latest/index.bs
```diff
@@ -263,16 +261,15 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
...
+The transformations are defined according to [[#trafo-md]]. The transformation types MUST only be `identity`, `translation` or `scale`.
+They MUST contain exactly one `scale` transformation that specifies the pixel size in physical units or time duration. If scaling information is not available or applicable for one of the axes set it to 1.
+It MAY contain exactly one `translation` that specifies the offset from the origin in physical units. If `translation` is given it must be listed after `scale` to ensure that it is given in physical coordinates.
+The length of the `scale` and `translation` array MUST be the same as the length of "axes".
+The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while being compatible with the general transformation spec.
+
+Each "multiscales" dictionary MAY contain the field "coordinateTransformations", describing transformations that are applied to each resolution level.
+The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:coordinateTransformations".
```
constantinpape, line 272: Now that we have resolved the "datasets:coordinateTransformations", let's return to whether we want to keep the "multiscales:coordinateTransformations", i.e. a transformation that is applied the same way to all scale levels after the individual transforms per scale level. For more context: the motivation for this feature is to refactor a transformation that is applied to all scale levels, e.g. scaling the time interval to be 0.1 seconds:
```yaml
"axes": [{"name": "t", "type": "time", "unit": "seconds"}, {"name": "y", "type": "space", "unit": "meter"}, {"name": "x", "type": "time", "unit": "meter"}],
# version with transformation only in datasets:
"datasets": [
  {"coordinateTransformations": [{"type": "scale", "scale": [0.1, 0.2, 0.2]}]},  # scale-level 0, phyiscal size is 20 cm
  {"coordinateTransformations": [{"type": "scale", "scale": [0.1, 0.4, 0.4]}]}   # scale-level 1, physical size is 40 cm, time scale is the same
]
# version with additional transformation in multiscales 
"datasets": [
  {"coordinateTransformations": [{"type": "scale", "scale": [1.0, 0.2, 0.2]}]},  # scale-level 0, phyiscal size is 20 cm
  {"coordinateTransformations": [{"type": "scale", "scale": [1.0, 0.4, 0.4]}]}   # scale-level 1, physical size is 40 cm
]
"coordinateTransformations": [{"type": "scale", "scale": [0.1, 1.0, 1.0]}]  # apply the timescale for both resolutions.
```

For our current transformations it is trivial to express the scale (or translation) in "multiscales:coordinateTransformations" without using "datasets:coordinateTransformations". However, for advanced transformations this is different: take the example of a non-uniform time axis. We could express this with a transformation that has a value for each discrete point along one axis, e.g. 100 values if we have 100 time points. In this case it would be much better to specify this once in "multiscales" and not several times in "datasets".

Given that we don't have this use-cases yet, I would vote to remove "multiscales:coordinateTransformations" from the current version to keep it simple. We can then introduce it (or a similar approach) once it becomes necessary in the next version. But I don't have a very strong opinion on this and would be ok with keeping it if a majority finds it useful already.

cc @bogovicj @will-moore (and comments are welcome from everyone of course)

constantinpape:
> While we can still rename things, can we consider renaming translation to translate (as I suggested on #57) since it is more consistent with scale (both verbs) and it matches what napari uses.
If translation is a convention elsewhere then that's OK, but it's always easier to avoid divergence if possible.

Sorry for missing this in #57 @will-moore. But I would be in favor of using nouns for all the transformations to be compatible with future versions (what's the verb for "affine"?). Note that `scale` is also a noun (our use here roughly matches 1. or 3. in https://www.collinsdictionary.com/dictionary/english/scale)

latest/index.bs
sbesson: > In this case, use scale=[1.0, 1.0, 1.0]. (More precisely scale=[1.0] * number_of_axes (python notation)).

This convention works for me. I expect in that case, the corresponding spatial axes should have no `unit` element, otherwise this should be interpretated as a pixel/voxel size of 1 micron/1 mm/...

latest/index.bs
sbesson: My understanding is that so far any transformations (including advanced transformations) can always be expressed using only `dataset:coordinateTransformations` and defining a transformation property at a higher level is primarily for optimization/efficiency purposes.

Within the scope of 0.4, my personal feeling is that the second form proposed above does not offer significant advantages while increasing the complexity of the metadata. For me, that's an incentive to leave this it out of the specification for now and introduce it as part of the next body of work when there are clear use cases.

Also, I assume you are talking about introducing the `coordinateTransfomations` level at the individual `multiscale`  level i.e. within an element of the `multiscales` array. One could also imagine defining it at the `multiscales` level i.e. a transformation that applies to all the multiscales lists (also true for `axes`).

will-moore:
Hi all, I've added a schema for v0.4 and valid/invalid examples at https://github.com/ome/ngff/pull/87.
If anyone wants to check that these match the current expectations of this PR, and suggest any other invalid examples, that would be great.
 Thanks!

latest/index.bs
thewtex: Either in 0.4 or at a later time we will want the global coordinate transformation for the use case of registration. With coordinate transformations that are the output of registration methods, they typically apply to all scales in the same way. While technically possible to duplicate these over the scales, when we support displacement field transformations this will result in unnecessarily large duplicated storage unless the `path` feature is used.

thewtex:
@constantinpape @will-moore in the documentation and examples, can we add the `translation` so the resulting image domains at multiple scales overlap? When downsampling, the center of the pixel shifts.

latest/index.bs
sbesson: :+1: thanks for the input @thewtex. Re-reading myself, I think my comment was hyperfocusing on the use case described in https://github.com/ome/ngff/pull/85#discussion_r794486756 and I still think a top-level `scale` element brings little value, primarily because `scale` is proposed to be mandatory at every `dataset` level.

The registration use case you are bringing up is a good one. Would it be useful to construct such an example where each dataset contains a `scale` and the multiscale object contains a `translation`? /cc @constantinpape

d-v-b:
>>> When downsampling, the center of the pixel shifts.

To elaborate on this, the effect of downsampling on the location of pixel centers is a degree of freedom in the downsampling routine. The multiscale metadata should not contain assumptions about the details of a downsampling routine, which mandates both scale and translation transforms for each scale level.

jbms:
@d-v-b @thewtex Regarding pixel centers, the "shift" also depends on whether integer pixel coordinates are on the boundary between two pixels or in the center of a pixel.  That should also be defined by this specification in order to specify the meaning of the transformations.  Note: Neuroglancer defines the coordinate space such that integer pixel coordinates are on the boundary between two pixels.

latest/index.bs
jbms: Neuroglancer internally uses a very similar multiscale volume representation to what is proposed here.  It specifies the following:

- For the top-level multiscale volume, there is a coordinate space that indicates:
  - Name of each dimension
  - Unit and scale for each dimension, e.g. "4 nm" or "0.5 s" or "" (for unitless dimensions like channel dimensions), that indicates the "native resolution" of the dataset.  For example if the unit is "4 nm", then changing the coordinate by 1 corresponds to a shift of 4 nm in physical space.
  - Optional lower and upper bounds for each dimension.
- Then for each scale there is an affine transform to transform from the voxel space of the scale to the native resolution indicated at the top level.  Typically the affine transform for the first scale is translation only, and the affine transform for subsequent levels are scale-and-translation-only, where the scale factors are the downsampling factors relative to the first scale.

While it breaks down once you introduce rotation or non-linear warping, the concept of a "native resolution" is very useful in practice when dealing with only translation and scale, so it may be valuable to preserve that in the ome-zarr spec.

While just supporting both per-multiscale and per-scale transformation may be somewhat useful to allow a more concise representation, I think it could be even more useful to specify the purpose of the separate transformations.  For example, if we wanted to follow what Neuroglancer does internally, we could say that the per-scale transformations should transform to the "native" resolution of the dataset, while the top-level transformation should indicate the native resolution of the dataset.  (If the top-level transform is not scale-and-translation-only, then we would not be able to infer a native resolution.)

thewtex:
> To elaborate on this, the effect of downsampling on the location of pixel centers is a degree of freedom in the downsampling routine. The multiscale metadata should not contain assumptions about the details of a downsampling routine, which mandates both scale and translation transforms for each scale level.

Yes, we do not have to make mandates on how downsampling is done. In the canonical examples for the most common use case of generating a multi-scale representation, it would be good to have the appropriate translations to help the reader's understanding.

> whether integer pixel coordinates are on the boundary between two pixels or in the center of a pixel. That should also be defined by this specification in order to specify the meaning of the transformations. 

Yes, we should clarify how a transformation applies to pixels. The center of a pixel is very strongly preferred because it is invariant to the spacing between pixels. It also means, for example, that pixels can be treated like points, https://github.com/ome/ngff/issues/65 , etc. and the transformation is applied in the same way.

lassoan:
> Note: Neuroglancer defines the coordinate space such that integer pixel coordinates are on the boundary between two pixels.

In 2D applications using one of the pixel corners as origin is probably more common, but in 3D imaging voxel coordinate system origin is typically the center. In VTK, ITK, and all applications based on these toolkits origin is in the pixel center. In all 3D image file formats that I know of (nrrd, nifti, metaio, dicom) origin is in the pixel center, too.

NGFF standard must specify this (if it hasn't been specified it already).

If any software that uses different coordinate system convention internally then it can convert to/from the standard coordinate system when reading/writing images.

latest/index.bs
constantinpape: Thanks for the feedback @thewtex and @jbms. I had similar use-cases as the registration one in mind when proposing the current structure. (But then I considered not introducing it in 0.4 and leaving it for the follow-up proposals to keep things simple, because with the current spec we can only express fairly simple transformations that won't get us all the way for many use-cases.) Anyway, I think given your comments there seems to be some consensus that having the `multiscales:coordinateTransformations` will be useful, so we might introduce it now already. I will try to get some feedback from @bogovicj on this as well, how this factors in with his initial approach to the extended transformation spec.

@sbesson: yes, we could introduce transformations even a level higher (that apply to each multiscales image), but I would not go there for now and would hope that such use-cases would instead be covered by an upcoming collections spec.

constantinpape:
@jbms @d-v-b @lassoan @thewtex 
Good that you bring up the issue of pixel centers. Indeed this needs to be defined clearly in the spec and I think this is not the case yet. I will read the relevant parts carefully again later to see if this is actually true and will open a separate issue about this, as I think that it can be separately discussed from the changes that are proposed here. (And mixing up too many discussions in a single PR makes it very hard to stay focused.)

Regarding examples: I agree that a case with `translation` should be in the examples. But I think we all agree that it should not be mandatory.

latest/index.bs
```diff
@@ -236,9 +237,6 @@ Additional fields for the entry depend on "type" and are defined by the column `
 | `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 | `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
```
constantinpape, line 238: It looks like this table still isn't rendered correctly:
table-rendering
 I don't understand why. Could someone with more bikeshed experience give this a shot? cc @joshmoore @will-moore @sbesson

bogovicj:
I am in favor of keeping the "multiscales:coordinateTransformations" for the reasons that @thewtex described.  

I agree with @thewtex and @lassoan re pixel centers, and so would be happy to say that explicitly without additional comment for `v0.4`. Having more details / examples will be important, but let's not delay this version for that.  Rather, I vote to move that discussion to another issue / new PR that could become `v0.4.1`. 

Thanks @constantinpape !

constantinpape:
Thanks for the comment @bogovicj. Given this feedback, let's keep the "multiscales:coordinateTransformations" in. 
Now we only need to decide about additional restrictions about scale cc @sbesson @will-moore:
Do we want to restrict the "multiscales:coordinateTransformations" to not contain scale to avoid redundancy with the "datasets:coordinateTransformations"?
My vote would be to not add this restriction now, as it will complicate the spec description and disallow the one actual use case we have for this in 0.4 (specifying the time axes for all res levels. Let me know what you think.

latest/index.bs
will-moore: I think I tried this way back on the previous PR, but didn't get much joy. I think @joshmoore suggested using an HTML table?

latest/index.bs
```diff
@@ -296,18 +293,18 @@ It SHOULD contain the field "metadata", which contains a dictionary with additio
...
+                    "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 1.0, 1.0, 1.0]}]  # the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
                 },
                 {
                     "path": "2",
-                    "transformations": [{"type": "scale", "scale": [2.0, 2.0, 2.0], "axisIndices": [2, 3, 4]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
+                    "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 2.0, 2.0, 2.0]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
                 }
             ],
-            "transformations": [{"type": "scale", "scale": [0.1], "axisIndices": [0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
+            "coordinateTransformations": [{"type": "scale", "scale": [0.1, 1.0, 1.0, 1.0, 1.0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
```
will-moore, line 307: I'm OK with keeping `multiscales:coordinateTransformations` and to not add restrictions to it.
But I wonder how well we want to support it before 0.4 is "done", since we've not done any work with it yet?
E.g. writing and reading it in `ome-zarr-py`, validating it in NGFF `image.schema`, reading it in `napari-ome-zarr` (combining it with the `datasets:coordinateTransformations`). This might come in follow-up PRs after the ones currently open. @sbesson?

will-moore (approved):

latest/index.bs
sbesson: I will give the current PR a final read tonight but overall I also agree with keeping the `multiscales:coordinateTransformations` in the spec as per the discussion above.

Let's review what our target should be in terms of implementations. At minimum, I think JSON schemas should validate `coordinateTransformations` at both levels - see https://github.com/ome/ngff/pull/87#pullrequestreview-869054310, our writers should be able to set `datasets:coordinateTransformations` and both `coordinateTransformations` should be available in the readers.

constantinpape:
This PR is good to go from my side now. @sbesson feel free to merge once you think it's good to go or ping me if you discover any more issues.

latest/index.bs
```diff
@@ -236,9 +237,6 @@ Additional fields for the entry depend on "type" and are defined by the column `
 | `translation` | one of: `"translation":List[float]`, `"path":str` | translation vector, stored either as a list of floats (`"translation"`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 | `scale`       | one of: `"scale":List[float]`, `"path":str` | scale vector, stored either as a list of floats (`scale`) or as binary data at a location in this container (`path`). The length of vector defines number of dimensions. |
 
-In addition, the field "axisIndices" MAY be given to specify the subset of axes that the transformation is applied to, leaving other axes unchanged. If not given, the transformation is applied to all axes. The length of "axisIndices" MUST be equal to the dimensionality of the transformation. If "axisIndices" are not given, the dimensionality of the transformation MUST be equal to the number of dimensions of the space that the transformation is applied to.
-If given, "axisIndices" MUST be given in increasing order. It uses zero-based indexing.
-
 The transformations in the list are applied sequentally and in order.
```
sbesson, line 240:
```suggestion
The transformations in the list are applied sequentially and in order.
```

latest/index.bs
```diff
@@ -263,17 +261,16 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
...
-The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while
-being compatible with the general transformation spec.
-
-Each "multiscales" dictionary MAY contain the field "transformations", describing transformations that are applied to each resolution level.
-The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:transformations".
-These transformations are applied after the per resolution level transformations specified in "datasets". They can for example be used to specify the `scale` for a dimension that is the same for all resolutions.
+Each dictionary MUST contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
+The transformations are defined according to [[#trafo-md]]. The transformation types MUST only be `translation` or `scale`.
```
sbesson, line 265:
```suggestion
The transformations are defined according to [[#trafo-md]]. The transformation MUST only be of type `translation` or `scale`.
```

latest/index.bs
```diff
@@ -263,17 +261,16 @@ Each dictionary in "datasets" MUST contain the field "path", whose value contain
...
+They MUST contain exactly one `scale` transformation that specifies the pixel size in physical units or time duration. If scaling information is not available or applicable for one of the axes set it to 1.
+It MAY contain exactly one `translation` that specifies the offset from the origin in physical units. If `translation` is given it must be listed after `scale` to ensure that it is given in physical coordinates.
```
sbesson, line 267:
```suggestion
It MAY contain exactly one `translation` that specifies the offset from the origin in physical units. If `translation` is given it MUST be listed after `scale` to ensure that it is given in physical coordinates.
```

sbesson (approved):
Thanks @constantinpape, the changes proposed match the discussions that happened as part of the last OME-NGFF meeting. Having re-read it carefully, I do not see any outstanding issue so I think it is a great time to freeze the proposed changes and move towards the 0.4 release.

Implementations and real-world examples can help exercising the new axes and transformations concepts introduced as part of this body of work. All feedback can be captured and reviewed together with the next body of work on transformations that will be led by @bogovicj.

d-v-b:
Following up on this suggestion from @thewtex: 
> Either in 0.4 or at a later time we will want the global coordinate transformation for the use case of registration. With coordinate transformations that are the output of registration methods, they typically apply to all scales in the same way. While technically possible to duplicate these over the scales, when we support displacement field transformations this will result in unnecessarily large duplicated storage unless the path feature is used.

I would like to discuss this global coordinate transformation, because the [example](https://github.com/ome/ngff/blob/main/0.4/index.bs#L315) given in the 0.4 spec confused me:

```json
"multiscales" : {
...
"datasets": [
    {
        "path": "0",
        "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 0.5, 0.5, 0.5]}]  # the voxel size for the first scale level (0.5 micrometer)
    }
    {
        "path": "1",
        "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 1.0, 1.0, 1.0]}]  # the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
    },
    {
        "path": "2",
        "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 2.0, 2.0, 2.0]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
    }
],
"coordinateTransformations": [{"type": "scale", "scale": [0.1, 1.0, 1.0, 1.0, 1.0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
...
}
```

In this example, we have two _different_ instances of the same transform applied to each scale level -- the global transform, and the scale-level-specific transform. Which one is correct? Should they be composed? In what order (for two scaling transforms the order doesn't matter...in which case why do we see two scaling transforms in this example?)

I suggest requiring that, for each `array` in `multiscales["datasets"]`, `array["coordinateTransformation"]` is _exactly the same_ as the `coordinateTransformation` attribute found in `array.path/.zarray`. I.e., `multiscales["datasets"][0]["coordinateTransformation"]` is complete on its own, without any extra information provided by other metadata fields in `multiscales`. This ensures that the semantic scope of `multiscales["datasets"]` is constrained to "transparently refer to a dataset + its metadata", and it also minimizes the number of different and possibly confusing `coordinateTransformation` objects lying around. Semantically, we can consider this requirement derived from the notion that `coordinateTransformations` are only properties of arrays, not groups of arrays.

@thewtex raised the concern that all scale levels might share some large `coordinateTransformation` (e.g.,  a displacement field), and we could use `multiscales["coordinateTransformation"]` to avoid redundant copies of the same large piece of metadata. I wonder how valid this concern is -- if a `coordinateTransformation` is so large that it's prohibitive to have 5 copies in metadata, then it's probably too large to have even once, and the correct approach is to store a reference to this transformation in all cases.

Curious to hear people's thoughts here. Personally I find this [example](https://github.com/ome/ngff/blob/main/0.4/index.bs#L315) so confusing that I think it needs to be clarified / amended.

cc @bogovicj

jbms:
> In this example, we have two different instances of the same transform applied to each scale level -- the global transform, and the scale-level-specific transform. Which one is correct? Should they be composed? In what order (for two scaling transforms the order doesn't matter...in which case why do we see two scaling transforms in this example?)

To me it seems natural that we have the following coordinate spaces:
- world (e.g. what is displayed in a viewer)
- multiscale
- dataset

The `coordinateTransformations` for each dataset indicate a transformation from the dataset coordinate space to the multiscale coordinate space.  The `coordinateTransformations` for the multiscale indicate a transformation from the multiscale coordinate space to the world coordinate space.

In cases where all scales other than the first (base level) are just computed from the base level via downsampling, it may be desirable to use the `dataset` -> `multiscale` transformations solely to indicate the downsampling factors and any offsets, and then rely on the `multiscale` -> `world` transformation to specify the actual voxel size.  That also means that if you later adjust the voxel size (e.g. because you previously estimated it incorrectly) there is just one place to change it, rather than having to change it for every scale as well.

You do raise another issue, though.  If we allow `coordinateTransformations` to be specified as a property of a single array : when referencing such an array as a dataset in the multiscale metadata, are you referring to the base (untransformed) array, or to the transformed array.  It seems to me that it would be desirable to be able to specify either.  Alternatively (my preference) we could avoid that problem by disallowing coordinate transformations on a plain array, and instead require them to be specified as a separate "view".  I would say that if a dataset refers to something that already has a coordinate transformation applied, the coordinate transformation specified for the dataset in the multiscale metadata should compose on top of any coordinate transformations specified by the referenced "dataset".

Note: it is not clear to me if the current spec allows coordinate transformations to be specified for a plain array --- I was under the impression that coordinate transformations could only be specified as part of the multiscale metadata.

There is also the question of whether a dataset in one multiscale group could actually itself be a `multiscale` group (e.g. to "import" all of the scales from another multiscale).

d-v-b:
> Note: it is not clear to me if the current spec allows coordinate transformations to be specified for a plain array --- I was under the impression that coordinate transformations could only be specified as part of the multiscale metadata.

Coordinate transformations _are_ defined for a single array (and I think this is their intended use) -- I believe this is covered [here](https://github.com/ome/ngff/blob/main/0.4/index.bs#L237) in the spec, although there is no example, and the introductory language [here](https://github.com/ome/ngff/blob/main/0.4/index.bs#L218) is a little confusing, i think instead of `.zattrs` it should refer to `.zarray` to make it clear that the subsequent fields are found in array metadata.

> In cases where all scales other than the first (base level) are just computed from the base level via downsampling, it may be desirable to use the dataset -> multiscale transformations solely to indicate the downsampling factors and any offsets, and then rely on the multiscale -> world transformation to specify the actual voxel size.

I think if each dataset in `multiscales` expresses its dataset -> world transformation, then there is no reason for consuming applications to ever formally represent the downsampling factors. The whole point of representing downsampling factors is to describe the scale and offset of the Nth scale level. If the Nth scale level directly communicates scale and offset, downsampling factors are not needed. 

> That also means that if you later adjust the voxel size (e.g. because you previously estimated it incorrectly) there is just one place to change it, rather than having to change it for every scale as well.

I agree that this is somewhat convenient, but I think the value of being explicit outweighs the burden in this case.

d-v-b:
(and no discussion of multiscale metadata is complete without a reference to the classic issue: https://github.com/zarr-developers/zarr-specs/issues/50)

constantinpape:
Thanks for raising this concern @d-v-b, a couple of quick comments from my side. 

> In this example, we have two _different_ instances of the same transform applied to each scale level -- the global transform, and the scale-level-specific transform. Which one is correct? Should they be composed

They should be composed, the transformation in `datasets` is applied first, then the "global" one in `multiscales` is applied.

The comment by @jbms summarizes the motivation for having this pretty well, with the caveat that we don't have `world` in the spec yet and that there might be multiple `world` transformations e.g. for different registrations. (Related motivation from @thewtex to don't have redundant transformations.)

> To me it seems natural that we have the following coordinate spaces:
> 
>     * world (e.g. what is displayed in a viewer)
> 
>     * multiscale
> 
>     * dataset

> Personally I find this [example](https://github.com/ome/ngff/blob/main/0.4/index.bs#L315) confusing that 

The idea to give the common timepoint scaing in `multiscale`. I agree that this example is artificial, it will make more sense once we have more advanced transformations, e.g. for specifiying non-uniform timepoints.

> Coordinate transformations _are_ defined for a single array (and I think this is their intended use) -- I believe this is covered [here](https://github.com/ome/ngff/blob/main/0.4/index.bs#L237) in the spec, although there is no example, and the introductory language [here](https://github.com/ome/ngff/blob/main/0.4/index.bs#L218) is a little confusing, i think instead of `.zattrs` it should refer to `.zarray` to make it clear that the subsequent fields are found in array metadata.

I think you are mixing up a couple of things here. First of all in the zarr spec `.zarray` only ever contains the metadata relevant for zarr, e.g. the chunk shape and compression. Other metadata is **always** stored in `.zattrs` (there is one exception from this done by netcdf, but I think that this is very unfortunate.) I think what you refer to is whether metadata is stored on the group level vs. on the array level:
```
image-group/
  .zgroup  <- this contains the zarr group metadata (just version atm)
  .zattrs <- this contains the group level metadata, which currently contains the ngff metadata
  scale-level0/  <- the array data for scale 0
    .zarray <- contains the zarr array metadata
    .zattrs <- contains additional metadata for the array, currently not used for ngff metadata
...
```
Ngff is currently not using `scale-level0/.zattrs` to store any metadata; this means that no transformations are currently not stored at the per array level.

Furthemore, `coordinateTransformation` are not defined specifically for a single array, they can describe the transformation of the array data space into a physical space (and that is there current use-case) but will also be useful for describing other transformations, e.g. registering an intermediate physical space to a world coordinate system. This is currently being fleshed out in #94.

>  I think it needs to be clarified / amended

There are multiple options to go forward:
- clarification: this is always welcome: please make a specific issue and point out briefly what exactly should be clarified, or make a PR to propose clarifications in https://github.com/ome/ngff/blob/main/latest/index.bs
- amendment: 0.4 is published and unless there is an internal inconsistency we will not change it more than semantically (e.g. for clarifications). It has been complicated enough to get many viewpoints together to get this release out, so we will not go back on it. So for technical changes there are two options:
  - propose changes for an intermediate release (0.4.1)
  - engage in the discussion around the extended transformation functionality in #94 and #101 for 0.5.0 (this is the option I would strongly encourage since I am doubtful we would want to change the reference implementations for a 0.4.1 if significant transformation changes are upcoming in 0.5.0)
  
Finally, I would suggest to move the discussion from this closed PR to either separate issues / PRs for specific problems or the future transformation discussions for general comments on transformations.

PR #87, opened 2022-01-28
Add 0.4 schemas and examples

will-moore:
In progress...

### DIFF
--- a/0.4/examples/invalid/invalid_axis_type.json
+++ b/0.4/examples/invalid/invalid_axis_type.json
@@ -1,0 +1,36 @@
+{
+    "multiscales": [
+        {
+            "axes": [
+                {
+                    "name": "y",
+                    "type": "invalid",
+                    "units": "micrometer"
+                },
+                {
+                    "name": "x",
+                    "type": "space",
+                    "units": "micrometer"
+                }
+            ],
+            "datasets": [
+                {
+                    "path": "0",
+                    "transformations": [
+                        {
+                            "scale": [
+                                1,
+                                1,
+                                0.5,
+                                0.13,
+                                0.13
+                            ],
+                            "type": "scale"
+                        }
+                    ]
+                }
+            ],
+            "version": "0.4"
+        }
+    ]
+}
--- a/0.4/examples/invalid/invalid_axis_units.json
+++ b/0.4/examples/invalid/invalid_axis_units.json
@@ -1,0 +1,33 @@
+{
+    "multiscales": [
+        {
+            "axes": [
+                {
+                    "name": "y",
+                    "type": "space",
+                    "units": "micron"
+                },
+                {
+                    "name": "x",
+                    "type": "space",
+                    "units": "micrometer"
+                }
+            ],
+            "datasets": [
+                {
+                    "path": "0",
+                    "transformations": [
+                        {
+                            "scale": [
+                                0.13,
+                                0.13
+                            ],
+                            "type": "scale"
+                        }
+                    ]
+                }
+            ],
+            "version": "0.4"
+        }
+    ]
+}
--- a/0.4/schemas/image.schema
+++ b/0.4/schemas/image.schema
@@ -1,0 +1,124 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ngff.openmicroscopy.org/0.3/schemas/image.schema",
+  "title": "NGFF Image",
+  "description": "JSON from OME-NGFF .zattrs",
+  "type": "object",
+  "properties": {
+    "multiscales": {
+      "description": "The multiscale datasets for this image",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "datasets": {
+            "type": "array",
+            "minItems": 1,
+            "items": {
+              "type": "object",
+              "properties": {
+                "path": {
+                  "type": "string"
+                }
+              },
+              "required": ["path"]
+            }
+          },
+          "version": {
+            "type": "string",
+            "enum": [
+              "0.4"
+            ]
+          },
+          "axes": {
+            "type": "array",
+            "minItems": 2,
+            "items": {
+              "type": "object",
+              "properties": {
+                "name": {
+                  "type": "string"
+                },
+                "type": {
+                  "type": "string",
+                  "enum": ["space", "time", "channel"]
+                },
+                "units": {
+                  "type": "string",
+                  "enum": [
+                    "angstrom", "attometer", "centimeter", "decimeter", "exameter", "femtometer", "foot", "gigameter", "hectometer", "inch", "kilometer", "megameter", "meter", "micrometer", "mile", "millimeter", "nanometer", "parsec", "petameter", "picometer", "terameter", "yard", "yoctometer", "yottameter", "zeptometer", "zettameter"
+                  ]
+                }
+              },
+              "required": ["name", "type"]
+            }
+          }
+        },
+        "required": [
+          "datasets", "axes"
+        ]
+      },
+      "minItems": 1,
+      "uniqueItems": true
+    },
+    "omero": {
+      "type": "object",
+      "properties": {
+        "channels": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "window": {
+                "type": "object",
+                "properties": {
+                  "end": {
+                    "type": "number"
+                  },
+                  "max": {
+                    "type": "number"
+                  },
+                  "min": {
+                    "type": "number"
+                  },
+                  "start": {
+                    "type": "number"
+                  }
+                },
+                "required": [
+                  "start",
+                  "min",
+                  "end",
+                  "max"
+                ]
+              },
+              "label": {
+                "type": "string"
+              },
+              "family": {
+                "type": "string"
+              },
+              "color": {
+                "type": "string"
+              },
+              "active": {
+                "type": "boolean"
+              }
+            },
+            "required": [
+              "window",
+              "color"
+            ]
+          }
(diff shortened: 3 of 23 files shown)


constantinpape:
Thanks for working on this @will-moore. I will give this a closer look once we have merged #85.

sbesson (approved):
Pushed a few extra commits to this PR adding more invalid examples and their associated constraints expressed as JSON schemas. From my side, happy to get this merged in its current state but I'll leave @constantinpape to take a look.

Probably the next todo will be for the schema to handle the definition `coordinateTransformations` at different levels (`dataset`, `multiscale` object) as https://github.com/ome/ngff/pull/85. As this conversation still needs to get finalized and to simplify the review, I'd propose to get this in a separate PR.

Also to avoid unnecessary copy-n-paste, I think it should be possible to use `$ref` and `$defs` to make some of these definitions re-usable at different levels - see https://json-schema.org/understanding-json-schema/structuring.html

0.4/examples/invalid/invalid_axis_type.json
```diff
@@ -0,0 +1,33 @@
+{
+    "multiscales": [
+        {
+            "axes": [
+                {
+                    "name": "y",
+                    "type": "invalid",
```
constantinpape, line 7: This is currently a SHOULD, see https://github.com/ome/ngff/blob/main/latest/index.bs#L217. Not sure how exactly we want to tackle this. I will leave a more general comment once I went through everything.

0.4/examples/invalid/invalid_axis_units.json
```diff
@@ -0,0 +1,33 @@
+{
+    "multiscales": [
+        {
+            "axes": [
+                {
+                    "name": "y",
+                    "type": "space",
+                    "units": "micron"
```
constantinpape, line 8: Same as for invalid type: this is currenrtly a SHOULD: https://github.com/ome/ngff/blob/main/latest/index.bs#L218

0.4/examples/invalid/mismatch_axes_units.json
```diff
@@ -0,0 +1,38 @@
+{
+    "multiscales": [
+        {
+            "axes": [
+                {
+                    "name": "t",
+                    "type": "time",
+                    "units": "micrometer"
```
constantinpape, line 8: Similar to the comment before this is technically valid if we have a SHOULD in the units.

0.4/schemas/image.schema
```diff
@@ -0,0 +1,258 @@
...
+          "datasets": {
+            "type": "array",
+            "minItems": 1,
+            "items": {
+              "type": "object",
+              "properties": {
+                "path": {
+                  "type": "string"
+                },
+                "coordinateTransformations": {
```
constantinpape, line 26: I think that coordinate transformations should be moved to a separate schema.

0.4/schemas/image.schema
```diff
@@ -0,0 +1,258 @@
...
+              "required": ["path", "coordinateTransformations"]
+            }
+          },
+          "version": {
+            "type": "string",
+            "enum": [
+              "0.4"
+            ]
+          },
+          "axes": {
```
constantinpape, line 101: Axes should also be moved to a separate schema.

constantinpape (commented):
A few comments from my side:
- In the current version of v0.4 units and type only have a SHOULD for the expected values.
- This brings up the general question: how do we deal with the SHOULD requirements? Maybe we need two schemas, one that is strict and fails for SHOULD and one that is lenient and allows it?
- I think that `axes` and `coordnateTransforms` should get their own schema, as they are meant to be referenced in other places in the spec. (But it's not strictly necessary to do this now.)

0.4/examples/invalid/invalid_axis_type.json
sbesson: Agreed,  this should not fail validation but probably be flagged by the stricter schema. I propose to open a follow-up PR with the infrastructure to differentiate the two of them.

0.4/examples/invalid/mismatch_axes_units.json
sbesson: Agreed, although as discussed yesterday, this is typically a real-world example where I find the consequences of the `SHOULD` requirement nonsensical. I'll open a corresponding issue

constantinpape: > Agreed, although as discussed yesterday, this is typically a real-world example where I find the consequences of the SHOULD requirement nonsensical.

Yeah, I agree that the consequences here are weird, but probably better to open an issue and not block the current release.

constantinpape: (and at least this would fail if we had a strict schema)

0.4/schemas/image.schema
sbesson: Agreed, separating `axes` and `coordinateTransformations`  matches the way the specification is written and should increase the readability and the management of these schemas.

A technical side, as these schemas are not yet published (and we need to have some deeper discussions around URL as well as packaging - see https://github.com/ome/ngff/pull/77), I suspect handling separate JSON schema files will require to make some decisions as relative references are not resolvable. Since these concepts are only used within the context of `multiscales` in v0.4, my inclination would be to start by defining internal sub-schemas using `#defs` - see https://github.com/ome/ngff/pull/87#pullrequestreview-869054310. As the intent is to use these concepts in other portions in the specification, I propose we review this assumption and define `axes.schema` and `coordinateTransformations.schema` when this becomes a requirement.

constantinpape: I agree, for now using them to `#defs` in the same schema is probably the best solution.

constantinpape:
Ok, after the feedback from @sbesson I suggest we proceed as follows:
- In this PR:
  - split `examples/invalid` into `examples/invalid_strict` and `examples/invalid_loose` (or so; I don't mind changing the names, but there should be two folders ;)). And move the examples in the corresponding folder.
  - make sure that all `valid` examples and `invalid_loose` pass the schema; `invalid_strict` fail the schema
  - move "axes" and "coordinateTransformations" to `#defs` to improve readability
- follow up work:
  - introduce separate schemas for MUST and SHOULD
  - move "coordinateTransformations" and "axes" to their own schemas

0.4/examples/invalid/invalid_axis_type.json
sbesson: Actually, re-reading this example, this should remain invalid.

As you rightfully mentioned `type` is a SHOULD and that is not sufficient ground to fail the strict schema. However there is only one `space` axe within the `multiscales` dictionary

constantinpape: yes, you're right. But the name of this example is a bit misleading then.

will-moore: And in fact it is failing because `type` is not `space, channel or time` rather than only-1-space-axis.
So the schema should allow 1 'invalid' axis.

constantinpape: > And in fact it is failing because type is not space, channel or time rather than only-1-space-axis.
So the schema should allow 1 'invalid' axis.

I am not sure what you mean by this @will-moore. But I agree that this one should fail the strict spec because it has only 1 spatial axis, as @sbesson points out. But I would maybe rename the example to make that clear.

sbesson: Nods, that's probably more tests worth injecting but my understanding of the `type` restrictions is the following for `multiscales.axes`:

- 1 space axes + anything else: invalid
- 2 (or 3) space axes, 1 channel and/or 1 time: valid
- 2 (or 3) space axes, 1 channel or 1 time), 1 custom type: valid but failing the strict schema
- 2 (or 3) space axes, 1 channel or 1 time, 1 unspecified: valid but failing the strict schema
- 2 (or 3) space axes, 2 customs or 2 unspecified or (1 custom and 1 unspecified): valid but failing the strict schema
- 4+ space axes + anything else: invalid

constantinpape: > 2 (or 3) space axes, 1 channel or 1 time, 1 unspecified: valid but failing the strict schema

with 3 spatial axes this would fail because we would have > 5 dim. Same for some of the other examples you bring up.
But otherwise I agree with your list.

sbesson: Heh the `or` in between 1 channel and 1 time was meant to capture that but I certainly failed to write the different combinations in a concise way. Expanding a bit, all the following scenarios are valid according to the minimal schema but should fail the strict schema:

- 2 space, 1 channel, 1 unspecified
- 2 space, 1 time, 1 unspecified
- 3 space, 1 channel, 1 unspecified
- 3 space, 1 time, 1 unspecified

constantinpape (approved):
I guess the separation of examples and following SHOULD / MUST schemas is a bit more tricky. So this is good to be merged from my side (and continued work could be done after the v0.4 release.)

will-moore:
@constantinpape @sbesson I think that's the essential things done for this PR.

Question: when `coordinateTransformation` is used at the `multiscales` level, `MUST` it include a `scale` object (as when used at the `dataset` level)? Currently the schema enforces that, since the same `coordinateTransformations` ref is used in both places.

constantinpape:
> Question: when `coordinateTransformation` is used at the `multiscales` level, `MUST` it include a `scale` object (as when used at the `dataset` level)? Currently the schema enforces that, since the same `coordinateTransformations` ref is used in both places.

@will-moore Yes, this is consistent with what we write in the spec right now, so I would go ahead and merge this.
Longer term: as far as I recall it's possible to extend definitions in jsonschema with a bit of magic, so that we don't need to require this in the coordinateTransformation schema directly.

PR #88, opened 2022-01-28
Define OME-Zarr and add Zarr v2 requirement

### LINKED ISSUE #80: Specify which zarr version ome.zarr is using
Currently it's zarr v2. And at some point we want/will switch to v3 (which is still under development).
cc @sbesson

sbesson:
Fixes #80.

Also attempts to define OME-Zarr i.e. a hierarchical organization of arrays as defined by the Zarr format with metadata defined by the OME-NGFF specification.

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -102,6 +102,8 @@ is represented here as it would appear locally but could equally
 be stored on a web server to be accessed via HTTP or in object storage
 like S3 or GCS.
 
+OME-Zarr is an implementation of the OME-NGFF specification using the Zarr format. Arrays MUST be defined and stored in a hierarchical organization defined by to the [version 2 of the Zarr specification ](https://zarr.readthedocs.io/en/stable/spec/v2.html). OME-NGFF metadata MUST be stored as attributes in the corresponding Zarr groups.
+
 Images {#image-layout}
 ----------------------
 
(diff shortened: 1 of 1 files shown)


will-moore:
`by to the` -> `by the`

constantinpape (approved):

will-moore:
Great 👍

latest/index.bs
```diff
@@ -102,6 +102,13 @@ is represented here as it would appear locally but could equally
 be stored on a web server to be accessed via HTTP or in object storage
 like S3 or GCS.
 
+OME-Zarr is an implementation of the OME-NGFF specification using the Zarr
+format. Arrays MUST be defined and stored in a hierarchical organization as
```
joshmoore, line 106: All for defining this, but it _might_ be that it would already "just work" with some implementations (e.g. a pre-release of zarr-python). Very similar to the `dimension_separator`. I imagine to drop the restriction will be a matter of testing the implementations that we consider MUST HAVEs, and if they pass, setting their minimal versions and dropping this clause.

sbesson: All agreed, I confess I have little knowledge of the compatibility with the upcoming spec. I went for the conservative approach of starting with a restrictive definition of the supported Zarr layout. This requirement can certainly be reviewed and relaxed. Maybe worth capturing as an issue?

joshmoore: :+1: for restrictive-first.

constantinpape: I would also vote for restrictive first here.

Issue #48, opened 2021-05-05
Compatibility with xarray


thewtex:
With the aspiration for OME-Zarr to be *The One Imaging Format to Rule them All* 💍  , I would like to propose compatibility with [xarray](https://xarray.pydata.org/en/stable/). For the most part, the needs of the:

- bioimaging
- geospatial imaging
- medical imaging
- many other scientific imaging domains

overlap. A common, well-supported standard will facilitate integration and cross-pollination across communities, and avoid those I/O headaches 🤯 . 

In summary, we could extend the current OME-Zarr spec to be compatible with the result of [xarray.Dataset.to_zarr](https://xarray.pydata.org/en/stable/generated/xarray.Dataset.to_zarr.html?highlight=to_zarr#xarray.Dataset.to_zarr), in a way that adds spatial metadata, addressing #28 #12, through the xarray encoded [`coords`](https://xarray.pydata.org/en/stable/terminology.html#term-Coordinate) using scientific imaging dimensions, `x`, `y`, `z`, `c`, `t`,  standard in OME-Zarr, for the [xarray array dimensions](https://xarray.pydata.org/en/stable/internals.html#zarr-encoding), making their name and order explicit #35.

<details>
<summary>Resulting consolidated metadata from <a href="https://idr.openmicroscopy.org/webclient/?show=screen-2603">idr0094</a></summary>

```
{
    "metadata": {
        ".zattrs": {
            "multiscales": [
                {
                    "datasets": [
                        {
                            "path": "0/idr0094"
                        },
                        {
                            "path": "1/idr0094"
                        },
                        {
                            "path": "2/idr0094"
                        },
                        {
                            "path": "3/idr0094"
                        },
                        {
                            "path": "4/idr0094"
                        },
                        {
                            "path": "5/idr0094"
                        }
                    ],
                    "name": "idr0094",
                    "version": "0.1"
                }
            ]
        },
        ".zgroup": {
            "zarr_format": 2
        },
        "0/.zattrs": {},
        "0/.zgroup": {
            "zarr_format": 2
        },
        "0/c/.zarray": {
            "chunks": [
                3
            ],
            "compressor": {
                "blocksize": 0,
                "clevel": 5,
                "cname": "lz4",
                "id": "blosc",
                "shuffle": 1
            },
            "dtype": "<u4",
            "fill_value": null,
            "filters": null,
            "order": "C",
            "shape": [
                3
            ],
            "zarr_format": 2
        },
        "0/c/.zattrs": {
[...]

Issue #86, opened 2022-01-27
Finalizing v0.4


constantinpape:
- [ ] MUST DO
  - [x] Finalize axes & transformations #85 
  - [x] Update the example data (e.g. in https://github.com/ome/ome-ngff-prototypes) for changes in #85 
  - [ ] Finalize support in ome-zarr-py/ome-zarr-napari, vizarr and BDV/MoBIE
  - [x] Updates to the HCS spec (cc @sbesson; I am not sure what the current status or related PR is)
  - [x] jsonschema for v0.4 #82
- [ ] CAN DO
  - [ ] top level version information #83 
  - [x] specify zarr version #80 PR: #88
  - [ ] specify encoding in json #81
- [ ] POST-RELEASE: write up a quick summary about how to contribute a spec extension (PR & release process)

Decided to push to next release:
  - Origin definition #89, see https://github.com/ome/ngff/issues/89#issuecomment-1027692519
  
Timeline: I will work on this next week during the hackathon and hope that we can release in the next 2 weeks or so.

sbesson:
@constantinpape for the HCS point, I'll review the recording of the evening session for the 6th NGFFcommunity call as I missed it. In the absence of vetoes, the next step will be to merge https://github.com/ome/ngff/pull/24

constantinpape:
See also https://github.com/ome/ome-ngff-prototypes/pull/2 for updating the example data

constantinpape:
I have updated the example data now, see https://github.com/ome/ome-ngff-prototypes#data-availability.
So now the only important thing left to do is to get the viewers up to speed ... I will ping the MoBIE side about this.

How's the progress on this in python/napari and vizarr, @sbesson @will-moore?

And do you still want to merge #92 before we release @sbesson? I would suggest to leave the other open points for the next minor release, or an intermediate patch release.

constantinpape:
I just found this old issue, which is still relevant: https://github.com/ome/ngff/issues/2
A good candidate for a patch release. (Just linking it here for visibility)

sbesson:
> And do you still want to merge https://github.com/ome/ngff/pull/92 before we release @sbesson?

Although I suspect this particular PR might be merged soon, I would not consider this as a blocker for  the specification release and rather mark it as a CAN DO item. Overall, there is still some work to perform on these JSON schemas including improving the checks in the strict image schemas (units), adding schemas for other specifications (plate, well). I suspect this will be broken out into PRs/issues and continue after the 0.4.0 release.

[...]

PR #96, opened 2022-02-07
Remove requirement for xarray-compatible metadata

sbesson:
See https://github.com/ome/ngff/issues/48 and https://github.com/ome/ome-zarr-py/issues/166

With the introduction of axes, metadata at the multiscales dataset level allowing compatibility with `xarray` was added in the 0.3 version of the OME-NGFF specification. Some feedback and testing revealed the metadata currently required by the specification generates Zarr group which fail to open in `xarray` due to the different shapes in the resolution `datasets`.

This PR proposes a last-minute amendment to the 0.4 specification which is due for publication imminently to remove the `xarray` metadata. This should prevent libraries from writing datasets known to be not fit-for-purpose at the moment. In the mean time, the discussion can keep going on in https://github.com/ome/ngff/issues/48 about the best way to ensure xarray-compatibility and lead to a formal proposal that can be released in an upcoming versions of the specification.

cc @thewtex

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -264,9 +264,6 @@ The length of "axes" must be between 2 and 5 and MUST be equal to the dimensiona
 The "axes" MUST contain 2 or 3 entries of "type:space" and MAY contain one additional entry of "type:time" and MAY contain one additional entry of "type:channel" or a null / custom type.
 The order of the entries MUST correspond to the order of dimensions of the zarr arrays. In addition, the entries MUST be ordered by "type" where the "time" axis must come first (if present), followed by the  "channel" or custom axis (if present) and the axes of type "space".
 If there are three spatial axes where two correspond to the image plane ("yx") and images are stacked along the other (anisotropic) axis ("z"), the spatial axes SHOULD be ordered as "zyx".
-The values of the "name" fields must be given as a list in the field "_ARRAY_DIMENSIONS" in the attributes (.zattr) of the zarr arrays.
-This ensures compatibility with the [xarray zarr encoding](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html#zarr-encoding).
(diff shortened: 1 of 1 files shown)


constantinpape (approved):
Good to go from my side.

joshmoore (approved):
:+1: with the caveat that there's a desire to find a way forward with this:
 - https://github.com/ome/ome-zarr-py/issues/166#issuecomment-1031415433
 - https://github.com/ome/ngff/issues/48

sbesson:
Summarizing the current state of the conversation happening in https://github.com/ome/ngff/issues/48, it was confirmed the current specification is not expected to work with `xarray` without additional constraints.

For anyone wanting to use `xarray` with OME-Zarr, there are two working layouts:

- either use distinct  axes names for the `_ARRAY_DIMENSIONS` of each multiscales datasets e.g. suffix it with the resolution index `["x0", "y0", "z0"]` under `0/.zattrs`, `["x1", "y1", "z1"]` under `1/.zattrs` ...
- or store the multiscale arrays under a nested layout - see https://github.com/ome/ngff/issues/48#issuecomment-1031668203`. In this case, the same axe names can be stored in the attributes of the dataset e.g. `["x", "y", "z"]` under `0/<name>/.zattrs`, `["x", "y", "z"]` under `1/<name>/.zattrs` ..

I'll leave this PR open until tomorrow morning for final vetoes/objections or suggestions and I am planning to get this in tomorrow morning to move forward with #86

sbesson:
Merging as per https://github.com/ome/ngff/pull/96#issuecomment-1031712772. 

@constantinpape how much more of https://github.com/ome/ngff/issues/86, would you consider as a must have? Otherwise, I would be inclined to start going through the steps of https://github.com/ome/ngff#release-process and get the 0.4 version of the specification published.

constantinpape:
@sbesson I think we can start the release process. The only "musts" left are making sure that it's possible to open the new spec in ome-zarr-py / vizarr / MoBIE. Implementation is pretty advanced for all three I think, so we don't need to block the release due to it. (But should maybe wait with advertising until it's possible to read it in the 3 tools.)

PR #97, opened 2022-02-08
Publish 0.4 specification

sbesson:
See #86.

Follows the release steps mentioned in https://github.com/ome/ngff#release-process. 

Happy to take suggestions on the wording and the rendering of the version history block. Previously we had one item per version but there were more changes in 0.4.0

See http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/sbesson/ngff/0.4_publication/0.4/index.bs for a rendered version of the page

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -20,7 +20,7 @@ Abstract: This document contains next-generation file format (NGFF)
 Abstract: specifications for storing bioimaging data in the cloud.
 Abstract: All specifications are submitted to the https://image.sc community for review.
 Status Text: The current released version of this specification is
-Status Text: <a href="../0.3/index.html">0.3</a>. Migration scripts
+Status Text: <a href="../0.4/index.html">0.4</a>. Migration scripts
 Status Text: will be provided between numbered versions. Data written with these latest changes
 Status Text: (an "editor's draft") will not necessarily be supported.
 </pre>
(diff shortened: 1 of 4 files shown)


constantinpape (approved):

constantinpape:
> Previously we had one item per version but there were more changes in 0.4.0

I think two items are good here, because the changes to the multiscales and the hcs spec are unrelated.

latest/index.bs
```diff
@@ -820,6 +820,16 @@ Version History {#history}
       <td>Description</td>
     </tr>
   </thead>
+  <tr>
+    <td>0.4.0</td>
+    <td>2022-02-08</td>
+    <td>multiscales: add axes type, units and coordinateTransformations</td>
+  </tr>
+  <tr>
+    <td>0.4.0</td>
```
joshmoore, line 829: I'd be for combining the two rows.

joshmoore: Ah, I see @constantinpape was for the opposite. Hmm.... happy to look into it later. (Also need to look into removing the "W3C" in the versioned pages)

constantinpape: I don't have a strong opinion here. Maybe we could start to introduce sub-points? like
```
- v0.4
  - multiscales: ...
  - high-content-screen: ...
```

sbesson: FWIW, I am not the biggest fan of the double row either but I was hoping someone would come with an authoritative suggestion 😄 

In the absence of a strong consensus, I'd be inclined to not delay the publication of the spec and capture this as an issue to investigate. We will certainly face similar cases in the future and it would be good not to be constrained in terms of what we want to document.

Issue #8, opened 2020-11-25
Extract metadata blocks for testing


joshmoore:
All the metadata JSON could be stored separately and included in order to be tested by one or more implementations.

sbesson:
With the ongoing work on validation and JSON schemas, we could look into extract the metadata snippets as a separate JSON files, have them tested by the CI build and include them as code files - https://tabatkins.github.io/bikeshed/#including-code

lubianat:
Hi,

Thank you all for the discussion here. As part of an effort to keep the NGFF issue tracker manageable, we're closing tickets that haven't seen activity in several years and whose discussion may be out of date.

If this issue is still relevant, feel free to reopen it with details in the current context of NGFF.

Best,
Tiago

PR #98, opened 2022-02-08
Extract specification snippets as standalone JSON files

sbesson:
See https://github.com/ome/ngff/issues/8, this is a proof-of-concept extracting the `multiscales` example snippet and using the JSON schema testing infrastructure introduced in https://github.com/ome/ngff/pull/92.

See http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/sbesson/ngff/json_snippets/0.4/index.bs#multiscale-md for the rendered version of this PR.

Incidentally, this allowed to flag several issues with the snippet which were fixed in https://github.com/ome/ngff/commit/7a62d84dd5709c4c75158b9506620ca00eae578c.

This work raised a few issues for managing these snippets moving forward:

- https://github.com/ome/ngff/commit/7a62d84dd5709c4c75158b9506620ca00eae578c removes the inline comments as there is no such markup in JSON. Should we keep these as an explanatory sentence before the example?
- I placed the JSON example under the `valid_strict` category introduced in [#92](https://github.com/ome/ngff/pull/92). Alternate suggestions for category or naming welcome.
- the long lines in the `coordinateTransformations` someway raise the question of the formatting for these snippets. Unless someone has a strong feeling re indentation and style, I assume we can be kept ad-hoc for now with the goal to improve readability

### DIFF
--- a/0.4/index.bs
+++ b/0.4/index.bs
@@ -285,45 +285,11 @@ Each "multiscales" dictionary SHOULD contain the field "name". It SHOULD contain
 Each "multiscales" dictionary SHOULD contain the field "type", which gives the type of downscaling method used to generate the multiscale image pyramid.
 It SHOULD contain the field "metadata", which contains a dictionary with additional information about the downscaling method.
 
-```
-{
-    "multiscales": [
-        {
-            "version": "0.4",
-            "name": "example",
-            "axes": [
-                {"name": "t", "type": "time", "unit": "millisecond"},
-                {"name": "c", "type": "channel"},
-                {"name": "z", "type": "space", "unit": "micrometer"},
-                {"name": "y", "type": "space", "unit": "micrometer"},
-                {"name": "x", "type": "space", "unit": "micrometer"}
-            ],
-            "datasets": [
-                {
-                    "path": "0",
-                    "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 0.5, 0.5, 0.5]}]  # the voxel size for the first scale level (0.5 micrometer)
-                }
-                {
-                    "path": "1",
-                    "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 1.0, 1.0, 1.0]}]  # the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
-                },
-                {
-                    "path": "2",
-                    "coordinateTransformations": [{"type": "scale", "scale": [1.0, 1.0, 2.0, 2.0, 2.0]}]  # the voxel size for the second scale level (downscaled by a factor of 4 -> 2 micrometer)
-                }
-            ],
-            "coordinateTransformations": [{"type": "scale", "scale": [0.1, 1.0, 1.0, 1.0, 1.0]],  # the time unit (0.1 milliseconds), which is the same for each scale level
-            "type": "gaussian",
-            "metadata": {                                       # the fields in metadata depend on the downscaling implementation
-                "method": "skimage.transform.pyramid_gaussian", # here, the paramters passed to the skimage function are given
-                "version": "0.16.1",
-                "args": "[true]",
-                "kwargs": {"multichannel": true}
-            }
-        }
-    ]
-}
-```
+<pre class=include-code>
+path: examples/valid_strict/multiscales_example.json
+highlight: json
+</pre>
+
 
 If only one multiscale is provided, use it. Otherwise, the user can choose by
 name, using the first multiscale as a fallback:
--- a/0.4/index.bs
+++ b/0.4/index.bs
@@ -99,7 +99,7 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL 
 “RECOMMENDED”, “MAY”, and “OPTIONAL” are to be interpreted as described in
 [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
-Some of the JSON examples in this document include commments. However, this only for
+Some of the JSON examples in this document include commments. However, these are only for
 clarity purposes and comments MUST NOT be included in JSON objects.
 
 On-disk (or in-cloud) layout {#on-disk}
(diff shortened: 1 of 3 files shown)


constantinpape:
> * [7a62d84](https://github.com/ome/ngff/commit/7a62d84dd5709c4c75158b9506620ca00eae578c) removes the inline comments as there is no such markup in JSON. Should we keep these as an explanatory sentence before the example?

Could we store them as yaml and make use of the fact that it supports comments somehow? (just a thought, not sure how this plays with the infrastructure here)

> * the long lines in the `coordinateTransformations` someway raise the question of the formatting for these snippets. Unless someone has a strong feeling re indentation and style, I assume we can be kept ad-hoc for now with the goal to improve readability

I think "ad-hoc with the goal to improve readability" sounds good

sbesson:
> Could we store them as yaml and make use of the fact that it supports comments somehow? (just a thought, not sure how this plays with the infrastructure here)

But even then, the YAML snippets would be reconverted as JSON to be rendered as `.zattrs` examples in the specification? Or would you envision showing the same examples using different formats?

constantinpape:
> But even then, the YAML snippets would be reconverted as JSON to be rendered as `.zattrs` examples in the specification?

Is there a way to load them as plaintext instead? Or convert them to yaml for rendering as well? Yaml could be rendered exactly the same as json + comments, since it's a superset. (Not sure if this is feasible with the bikeshed infrastructure of course)

joshmoore:
I imagine we could add a `build.sh` (or `Makefile`!) which has a pipeline of `yaml -> .json -> #include`.

sbesson:
While adding a YML-> JSON build step is certainly doable, my two main worries are 1- the added complexity both in terms of developing/building the specification and 2- the fact these comments will still be lost as part of the translation and will disappear from the specification.

Trying to come up with another alternative, we could take advantage of the fact that the specification is meant to be extensible and store these as custom explanatory JSON keys e.g.:

```json
                          {
                              "type": "scale",
                              "scale": [1.0, 1.0, 0.5, 0.5, 0.5],
                              "comment": "the voxel size for the first scale level (0.5 micrometer)"
                         }
```

Thoughts @constantinpape ?

joshmoore:
re `"comment":` see an extended discussion at https://github.com/w3c/json-ld-syntax/issues/32#issuecomment-409014209

sbesson:
Thanks @joshmoore, it's good to know that others have the same problem and that there are discussions to formalize this concept although it's clear from the thread and https://github.com/w3c/json-ld-syntax/issues/32#issuecomment-410464422 that there is no native JSON-LD solution at this point.

I pushed additional commits that implement a version of the proposal above, using in that case `description` as a "magic" key, in the sense that it allows to store additional information but is neither defined by the specification nor disallowed - see http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/sbesson/ngff/json_snippets/0.4/index.bs#multiscale-md for the output.

sbesson:
Additional searches did not allow me to identify a solution that would both produce valid JSON while respecting the concerns expressed in https://github.com/ome/ngff/pull/98#issuecomment-1064392610 regarding the usage of magic/private/reserved keys in the official specification.

Largely inspired by https://google.github.io/styleguide/jsoncstyleguide.xml, the last set of commits ([fbf4aee](https://github.com/ome/ngff/pull/98/commits/fbf4aee3b2ee6deb9ba411eacab7a2cc0bc28304), [ae8839c](https://github.com/ome/ngff/pull/98/commits/ae8839c35bd2fc4e35d6be4ea77b80b12ec8db78) and [a7038b6](https://github.com/ome/ngff/pull/98/commits/a7038b6fc518b2f23c6c1ad08c47a27268caa999)) makes the following proposal:

- comments are reinserted into the extracted JSON as previously (using `//` as the separator but we can keep `#` if that's preferred)
- the Python tests are updated to strip the comment lines before running the validation against the JSON schemas
- a new `Document conventions` section is added to the specification which uses the standard RFC 2119 blurb and also clarifies the explanation scope of the comments in the JSON snippets

With this change, there should be no loss of information compared to the current published specification. The outcome is rendered at:

http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/sbesson/ngff/json_snippets/0.4/index.bs#document-conventions
http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/sbesson/ngff/json_snippets/0.4/index.bs#multiscale-md

0.4/index.bs
```diff
@@ -92,6 +92,16 @@ of bioimaging data, whether during acquisition or sharing in the cloud.
...
 
+Document conventions
+--------------------
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”,
+“RECOMMENDED”, “MAY”, and “OPTIONAL” are to be interpreted as described in
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+Some of the JSON examples in this document include commments. However, this only for
+clarity purposes and comments SHOULD NOT be included in JSON objects.
```
constantinpape, line 103: I would change this to MUST NOT, since it will not be valid json.

constantinpape (approved):
This solution looks good to me @sbesson. (I would only change SHOULD NOT to MUST NOT for the remark on json comments since any comments would result in invalid json)

0.4/index.bs
```diff
@@ -92,6 +92,16 @@ of bioimaging data, whether during acquisition or sharing in the cloud.
...
 for all examples.
 
+Document conventions
+--------------------
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”,
+“RECOMMENDED”, “MAY”, and “OPTIONAL” are to be interpreted as described in
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+Some of the JSON examples in this document include commments. However, this only for
```
joshmoore, line 102:
```suggestion
Some of the JSON examples in this document include commments. However, these are only for
```

joshmoore:
> https://github.com/ome/ngff/pull/98#issuecomment-1070598704 comments are reinserted into the extracted JSON as previously (using // as the separator but we can keep # if that's preferred)

Much happier with this than with magic keys based on the stage that we are at. :+1:

sbesson:
Pushed minor changes. I'll leave until tomorrow for people to comment but otherwise I propose to get this merged to update the 0.4 specification.
I'll follow-up with a PR against `latest` with the similar changes but also introducing examples and JSON schemas.

PR #106, opened 2022-03-04
Clarifies that scale should always capture the downsampling factor

sbesson:
See discussion at https://github.com/glencoesoftware/bioformats2raw/pull/134#issuecomment-1059095720

For instance in the case of a synthetic 5D dataset downsampled by a factor of 2 alongside the last two dimensions, the scale vectors for each of the datasets should look like

```
[1.0, 1.0, 1.0, 1.0, 1.0]
[1.0, 1.0, 1.0, 2.0, 2.0]
[1.0, 1.0, 1.0, 4.0, 4.0]
```

I am not overly happy with the sentence but I thought it was preferable to make a first suggestion rather than opening an issue. Alternate suggestions welcome :)


0.4/index.bs
```diff
@@ -271,7 +271,7 @@ to the current zarr group. The "path"s MUST be ordered from largest (i.e. highes
 Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
 Each dictionary MUST contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
 The transformations are defined according to [[#trafo-md]]. The transformation MUST only be of type `translation` or `scale`.
-They MUST contain exactly one `scale` transformation that specifies the pixel size in physical units or time duration. If scaling information is not available or applicable for one of the axes set it to 1.
+They MUST contain exactly one `scale` transformation that specifies the pixel size in physical units or time duration. If scaling information is not available or applicable for one of the axes, the value MUST express the scaling factor between the current resolution and the first resolution for the given axis, defaulting to 1.0 if there is no downsampling alongside the axis.
```
will-moore, line 274: "alongside the axis" maybe should be "along the axis" or "for that axis"

will-moore:
Just coming back to the spec having not read it recently, I was initially confused by `Each dictionary MUST contain the field "coordinateTransformations"` thinking that `coordinateTransformations` was in the "multiscales" dictionary. I wonder if you could add `Each "datasets" dictionary MUST contain...` as we do elsewhere, just to improve clarity.

I also find it a bit confusing that we use 'dataset' and 'resolution' interchangably to refer to the same thing. E.g. you refer to `current resolution` which means "the dataset that contains this coordinateTransform". It's a shame we didn't use "resolutions" instead of "datasets" since "datasets" doesn't mean anything. Probably not much we can do about this now.

will-moore:
LGTM 👍

d-v-b:
While we are making changes, I would recommend not using "pixel size" (since array data is not made of pixels), and instead refer to "grid spacing". 

Likewise, "resolution" implies something about the content of the data, e.g. how well physical distances can be resolved. Probably best to avoid that kind of semantics, since for oversampled data you can have multiple levels of a multiresolution group that are all the same effective resolution. The term "dataset" is probably a vestige of hdf5 / n5 terminology to denote what in zarr is called an array.

joshmoore:
One other possible though: is the change sufficiently covered by examples in the text?

constantinpape (approved):

constantinpape:
The changes look good to me. 

> One other possible though: is the change sufficiently covered by examples in the text?

Sorry, I am not quite sure what exactly you mean by this, @joshmoore.

I also agree with the points in https://github.com/ome/ngff/pull/106#issuecomment-1059346740, but I think it would make sense to do this in a separate PR, from my experience the smaller the better. @d-v-b do you want to go ahead and do this?

joshmoore:
> Sorry, I am not quite sure what exactly you mean by this, @joshmoore.

Just whether or not we want an example on the webpage that shows this explicitly, e.g. under https://ngff.openmicroscopy.org/latest/#trafo-md

constantinpape:
> > Sorry, I am not quite sure what exactly you mean by this, @joshmoore.
> 
> Just whether or not we want an example on the webpage that shows this explicitly, e.g. under https://ngff.openmicroscopy.org/latest/#trafo-md

I guess in general wouldn't hurt to have more examples... But since I understood @sbesson this is blocking some work on bioformats2raw, so I would suggest to merge now?!

joshmoore:
> I guess in general wouldn't hurt to have more examples... But since I understood @sbesson this is blocking some work on bioformats2raw, so I would suggest to merge now?!

Yes, not a blocker :+1:

Issue #104, opened 2022-02-25
OME Metadata Support


joshmoore:
*This issue captures the requirements as well as the possible implementation choices for a first integration of OME metadata into the OME-NGFF container. The UoD team is targeting the specification as well as implementations, including within OMERO, by mid-2022.*

## Goal

NGFF specifications up to version 0.4 contain only minimal metadata fields that cover the existing OME model (e.g. physical pixel size). As a result, converting data into OME-NGFF, e.g. with `bioformats2raw`, loses more metadata than the equivalent conversion to OME-TIFF. **The goal of this issue is to achieve a parity between the two formats in terms of capturing metadata contained in the OME model (2016-06).**

Here we would like to discuss, plan, and specify an initial integration of the OME model into OME-NGFF. As with other specifications, this initial work will likely be followed by multiple, possibly breaking, changes to expand the scope. Where possible, we will also try to capture that roadmap here.

## In-scope requirements

-   The primary requirement is the ability to fully convert an OME-TIFF in its entirety into an OME-NGFF dataset.

-   As a corollary, it should be also possible to capture what Bio-Formats knows about proprietary files formats (PFFs). This should cover minimally the intermediate Zarr output of `bioformats2raw` and be readable by all readers (not just `raw2ometiff`).

-   The specification should make clear how the combination of the OME model metadata and NGFF metadata is to be interpreted and what readers should do in the case of a conflict. 

## Out-of-scope requirements

-   This need not be the final mechanism used for OME-NGFF to store metadata. It is more important to capture the metadata that exists today.

-   It is not necessary, at least initially, that an OME-NGFF be fully convertible back into an OME-TIFF since, e.g., there is no location for labels, transformations, or file annotations in OME-TIFF.

----

## Design decision \#1: Location of metadata

The current NGFF specification solely uses the formats custom-attributes (`.zattrs`) for storing metadata. Several other locations are conceivable, though some are more or less within the bounds of the Zarr specification (See <https://github.com/zarr-developers/zarr-specs/issues/112> for more discussion.)

### Option a) `.zattrs`

[...]

Issue #94, opened 2022-02-02
Proposing spaces and transforms


bogovicj:
This is a preliminary proposal and discussion for axes and coordinate transformations. For more details see:
* [this early version of the specification](https://github.com/bogovicj/ngff/blob/coord-transforms/latest/index.bs).
* [this page of examples](https://github.com/bogovicj/ngff/wiki/Transform-Examples) for a growing list of concrete descriptions of particular [use cases](https://github.com/ome/ngff/issues/84).
*  [this page of notes, rationale, and brainstorming](https://github.com/bogovicj/ngff/wiki/Transforms-notes,-examples,-proposals)

## Add discrete axes
Extend the [axes specification](https://ngff.openmicroscopy.org/latest/#axes-md), adding a new optional field `discrete` whose values are booleans.  Discrete axes should not be interpolated, whereas continuous axes may be interpolated. For example, the channels are usually discrete:

```json
{"name": "c", "type": "channel", "discrete": true },
{"name": "y", "type": "space", "unit": "micrometer"},
{"name": "x", "type": "space", "unit": "micrometer"},
```

## Add spaces

A space is a named list of axes and defines the coordinate system of the data. Two simple examples - a viewer may prefer the physical spatial coordinates of data:

```json
{ 
"name" : "physical-micrometers", 
"axes" : [ 
  {"name": "y", "type": "space", "unit": "micrometer"},
  {"name": "x", "type": "space", "unit": "micrometer"}
]
}
```

where an algorithm that processes pixels may prefer the discrete pixel grid:
```json
"name" : "pixel-space", 
"axes" : [ 
  {"name": "j", "type": "space", "discrete": true },
  {"name": "i", "type": "space", "discrete": true }
]
```

### array space

This is a nice default that does not hurt much that I can see, and is more concise than alternatives.  [See my brainstorming here](https://github.com/bogovicj/ngff/wiki/Transforms-notes,-examples,-proposals#convenience-and-shorthands-for-spaces).

Every array / dataset has a default space whose name is the empty string.  It's axes have default names `dim_i`, are discrete, and there exist as many axes as array dimensions.  For example, a 3D dataset's array space is 

```json
{
"name" : "", 
"axes" : [ 
  {"name": "dim_0", "discrete": true },
  {"name": "dim_1", "discrete": true }
]
}
```

Array space is shared across all datasets, so if any applications need to differentiate between them, a new discrete space should be explicitly defined with other, unique axis names. 

## coordinateTransformations

[...]

PR #110, opened 2022-03-18
 Extract specification snippets as standalone JSON files (latest)

sbesson:
Follow-up of #98, this ports the additions made to the 0.4 specification i.e. the extraction of the JSON snippets and new DOcuments conventions paragraph to the `latest` specification.

Additionally, 0765b4d4aa00509e0263887a425f38f7a930028e and  ff936b894f9047508849bee8072b6388d0ac935a proposes to add the rest of the validation infrastructure i.e. the JSON schemas and the validation tests run as part of the CI to the `latest` subfolder.
From my perspective, the secondary benefit of adding this infrastructure is to make the JSON examples and schemas truly first-class citizens of upcoming NGFF proposals made against `latest`. The primary downside is that it might introduce some friction for instance if multiple features are under development. Interesting to hear what people who proposed the next set of breaking changes of the specification feel about this approach especially @bogovicj (#94) and @joshmoore  (#104).


bogovicj:
Nice @sbesson !

>  make the JSON examples and schemas truly first-class citizens of upcoming NGFF proposals 

My first impression is that this is a really idea. Better to have to think about existing examples when making changes than to forget about them!  I'll look more carefully next week

latest/schemas/strict_image.schema
```diff
@@ -0,0 +1,19 @@
+{
+  "$id": "https://ngff.openmicroscopy.org/0.4/schemas/strict_image.schema",
+  "allOf": [
+    {
+      "$ref": "https://ngff.openmicroscopy.org/0.4/schemas/image.schema"
```
will-moore, line 5: Do these IDs get updated directly to `/0.5/` at this point? (don't need to use `/latest/`, right?)
NB: I notice that https://ngff.openmicroscopy.org/0.4/schemas/image.schema is 404, so this schemas won't work without the custom resolver used in the tests.

sbesson: No you are correct,  I didn't do the update work as I was waiting to get a consensus about the value of adding these scneas.

A few comments:
- definitely need to replace `0.4`. I would not use `0.5`  as that would communicate a released schema and that has the potential of creating confusion. `latest` would be an option or maybe something that explicitly says this is a development schema e.g. `0.5dev`?
- the individual `version` elements in the JSON snippets/examples will likely need to be updated to follow the same convention
- [the release process](https://github.com/ome/ngff/blob/main/README.md#release-process) will need to be amended to include these steps

joshmoore:
https://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/sbesson/ngff/json_snippets_latest/latest/index.bs continues to look good. There are still a few `0.4` matches in the document. For those in the index.bs itself, adding:

```
Text Macro: NGFFVERSION 0.5-dev
```

to the header will lead to replacement. This unfortunately does *not* work for the `<pre class=include-code>` block. (Only `class=include` which doesn't do any formatting.)

sbesson:
Created a text macro to manage the replacements (all the other usages of 0.4 are legit and refer either to the stable version and/or to the version history). Also replaced the versions in the JSON snippets manually. My intent once this is merged is to extract all included code blocks into standalone JSON files so that we can validate them via JSON schemas and include them in the spec.

latest/examples/valid_strict/multiscales_example.json
```diff
@@ -0,0 +1,54 @@
...
+                    "path": "2",
+                    "coordinateTransformations": [{
+                        // the voxel size for the third scale level (downscaled by a factor of 4 -> 2 micrometer)
+                        "type": "scale",
+                        "scale": [1.0, 1.0, 2.0, 2.0, 2.0]
+                    }]
+                }
+            ],
+            "coordinateTransformations": [{
+                // the time unit (0.1 milliseconds), which is the same for each scale level
```
will-moore, line 40: Are these comments valid JSON? Do they cause any issues? (I see they are the same as this under 0.4/)

joshmoore: from https://github.com/ome/ngff/pull/98#issuecomment-1070598704

> a new Document conventions section is added to the specification which ... clarifies the explanation scope of the comments in the JSON snippets

joshmoore:
Any objections to moving forward with this?

will-moore:
GFI 👍

bogovicj:
looks good to me

PR #112, opened 2022-04-07
bioformats2raw.layout

joshmoore:
First update for supporting bioformats2raw output as described in #104. Introduces the concept of "transitional" metadata to allow readers to start supporting extensions that have grown in the community before they are added here.

Draft previwable under: http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/joshmoore/ngff/bf2raw/latest/index.bs#bf2raw

Screen Shot 2022-04-07 at 16 04 04

Note: currently builds on top of #110 to prevent conflicts.


latest/index.bs
```diff
@@ -241,9 +249,28 @@ keys as specified below for discovering certain types of data, especially images
...
+------------------------------------------------
+
+[=Transitional=] "bioformats2raw.layout" metadata identifies a group which implicitly describes a series of images.
+The need for the collection stems from the common "multi-image file" scenario in microscopy. Parsers like Bio-Formats
+define a strict, stable ordering of the images in a single container that can be used to refer to them by other tools.
+In order to capture that information within an OME-NGFF dataset, `bioformats2raw` internally introduced a wrapping
+layer. An upcoming NGFF specification will replace this layout with explicit metadata.
+
+Conforming groups:
+- MUST have the value "3" for the "bioformats2raw.layout" key at the top of the hierarchy.
```
will-moore, line 262: Does this mean that the `.zattrs` has `"bioformats2raw.layout": 3`?
An example layout would help I think.

latest/index.bs
```diff
@@ -231,9 +249,34 @@ keys as specified below for discovering certain types of data, especially images
...
+Conforming groups MUST have the value "3" for the "bioformats2raw.layout" key in their `.zattrs` metadata at the top of the hierarchy:
+
+<pre class=include-code>
+path: examples/valid_strict/bioformats2raw_layout.json
+highlight: json
+</pre>
+
+Additionally:
+- MUST store separate "multiscales" images in consecutively numbered groups starting from 0 (i.e. "0/", "1/", "2/", "3/", ...).
+- MAY have OME metadata representing the entire collection of images in a file named "OME/METADATA.ome.xml".
```
melissalinkert, line 270: Do we want to specify what `OME/METADATA.ome.xml` must or should contain if it exists? E.g. something to the effect of `MUST represent every "multiscales" group as exactly one OME-XML "Image" in the same order as the group numbers`.

joshmoore: Good point. :+1:

joshmoore: Pushed.

latest/index.bs
```diff
@@ -93,6 +93,24 @@ of bioimaging data, whether during acquisition or sharing in the cloud.
...
 
+Document conventions
+--------------------
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”,
+“RECOMMENDED”, “MAY”, and “OPTIONAL” are to be interpreted as described in
+[RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+<p>
+<dfn>Transitional</dfn> metadata is added to the specification with the
```
will-moore, line 104: Does this need it's own `Transitional metadata` heading (instead of being under `Document conventions`?

will-moore: Ah - I see that's got it's own section below. ok

joshmoore:
Pushed some schema fun, @will-moore, if you want to take a look. I need to force push away the commits from #110 but I'll hold off on that a bit.

sbesson:
> Pushed some schema fun, @will-moore, if you want to take a look. I need to force push away the commits from https://github.com/ome/ngff/pull/110 but I'll hold off on that a bit.

I think the latest commit is missing `bf2raw.schema`. Unbearable suspense ! :)

joshmoore:
> I think the latest commit is missing `bf2raw.schema`. 

Thanks. Pushed.

> Unbearable suspense ! :)

Compared to the debate?

will-moore:
Now that #110 is merged, can you merge `main` into here so that the diffs are cleaner?

joshmoore:
> can you merge `main` into here so that the diffs are cleaner?

Done, @will-moore

joshmoore:
Pushed a statement about the recent realization that we want this layout for _previous_ versions as opposed to upcoming versions. There might be a function in bikeshed along the lines of "applicable-to-version" that we may want to make use of.

latest/index.bs
```diff
@@ -242,9 +250,36 @@ keys as specified below for discovering certain types of data, especially images
...
+this layout with explicit metadata.
+
+<pre class=include-code>
+path: examples/bf2raw/bioformats2raw_layout.json
+highlight: json
+</pre>
+
+Conforming groups:
+- MUST have the value "3" for the "bioformats2raw.layout" key in their `.zattrs` metadata at the top of the hierarchy:
+- MUST store separate "multiscales" images in consecutively numbered groups starting from 0 (i.e. "0/", "1/", "2/", "3/", ...);
```
sbesson, line 271: At least for backwards compatibility, the primary exception to this rule would be HCS OME-NGFF dataset with metadata stored as `OME/METADATA.ome.xml`. In that case the top-level group (`plate`) MUST include the `bioformats2raw.layout` key but the group hierarchy should remain compliant with the HCS specification described below (i.e. 2 levels of hierarchies for wells with alphanumeric characters)

latest/tests/test_validation.py
```diff
@@ -7,19 +7,25 @@
...
     image_schema['$id']: image_schema,
     strict_image_schema['$id']: strict_image_schema,
 }
 
 resolver = RefResolver.from_schema(image_schema, store=schema_store)
 validator = Draft202012Validator(image_schema, resolver=resolver)
 strict_validator = Draft202012Validator(strict_image_schema, resolver=resolver)
+bf2raw_validator = Draft202012Validator(bf2raw_schema, resolver=resolver)
 
+bf2raw_files = list(glob.glob("examples/bf2raw/*.json"))
```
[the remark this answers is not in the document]
sbesson: With #118 included, we should have a framework for extending the JSON schema tests to other specifications

latest/index.bs
```diff
@@ -242,9 +250,36 @@ keys as specified below for discovering certain types of data, especially images
...
+- MAY have OME metadata representing the entire collection of images in a file named "OME/METADATA.ome.xml";
```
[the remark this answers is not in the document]
will-moore: How much metadata `SHOULD` METADATA.ome.xml have?
What does the format of it look like? If readers of this spec find that they should be creating such a file, where can they find a minimal example of what they should write? It probably needs to be in this spec itself?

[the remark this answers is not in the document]
melissalinkert: Maybe referencing https://docs.openmicroscopy.org/ome-model/6.2.2/specifications/minimum.html (but with no `TiffData` or `BinData`) would be enough?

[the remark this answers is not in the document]
sbesson: Related to the discussion in https://github.com/glencoesoftware/bioformats2raw/issues/137, could we make the statement that the metadata MUST be compliant with the OME schema and that each `Pixels` element MUST use `MetadataOnly`?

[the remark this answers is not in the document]
melissalinkert: Adding `MUST be compliant with the OME schema` or similar makes sense to me (which implies that every `Image` has `Pixels` with minimal attributes populated). I think I'm OK with `MetadataOnly` being a `MUST`, I just need to address https://github.com/glencoesoftware/bioformats2raw/issues/137 and a corresponding fix in raw2ometiff in order to be compliant.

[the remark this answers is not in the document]
sbesson: Coming back to this, as per the work on https://github.com/glencoesoftware/bioformats2raw/pull/156, the upcoming `bioformats2raw 0.5.0` release should produce valid OME-XML with an OME-NGFF 0.4 specification.

As earlier versions would not strictly produce valid XML, I am struggling to find a way to express that the XML MUST be compliant with the OME schema from 0.4 and onwards? The alternative to https://github.com/ome/ngff/pull/112#issuecomment-1123560478 would be to update the specification every published version (possibly only limited ourselves to 0.2+) with a section about the transitional layout.

latest/index.bs
joshmoore: Good catch. I was a bit unsure how to work this in without having ever more complicated if/then logic. See what you think about the version I just pushed which basically says the "plate" takes precedence.

latest/index.bs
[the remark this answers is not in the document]
joshmoore: Added a link.

joshmoore:
* Thanks for the comments, all. Pushed a few commits.
* @sbesson: updated validation tests to match.

If we're starting to feel fairly confident of the change, I can start on the backport to 0.4 (and beyond?)

sbesson:
Now I almost want to have https://github.com/ome/ngff/pull/128 in first to have a review link handy :)

latest/index.bs
```diff
@@ -242,9 +250,40 @@ keys as specified below for discovering certain types of data, especially images
...
+<pre class=include-code>
+path: examples/bf2raw/bioformats2raw_layout.json
+highlight: json
+</pre>
+
+Conforming groups:
+- MUST have the value "3" for the "bioformats2raw.layout" key in their `.zattrs` metadata at the top of the hierarchy;
+- MUST store separate "multiscales" images in consecutively numbered groups starting from 0 (i.e. "0/", "1/", "2/", "3/", ...);
+- SHOULD have OME metadata representing the entire collection of images in a file named "OME/METADATA.ome.xml" which
+  MUST adhere to the specification available at https://docs.openmicroscopy.org/ome-model/6.2.2/specifications/minimum.html;
```
will-moore, line 273: And MUST use `<MetadataOnly/>` block instead of `<BinData/>`?

joshmoore:
Review URL: http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/ome/ngff/e2119e5ccd9c97da8717c53b889f377c39d121a9/latest/index.bs

joshmoore:
One more time (from 3f10889bb30511b70b973ca6d483368e411d863a)

joshmoore:
Should be ok to give it a try, @will-moore.

joshmoore:
No, the comments get updated in place. Moving forward, it should be the first comment on a PR directly under the description.

will-moore:
LGTM 👍

latest/index.bs
```diff
@@ -242,9 +250,52 @@ keys as specified below for discovering certain types of data, especially images
...
+path: examples/bf2raw/bioformats2raw_layout.json
+highlight: json
+</pre>
+
+Conforming groups:
+- MUST have the value "3" for the "bioformats2raw.layout" key in their `.zattrs` metadata at the top of the hierarchy;
+- MUST store separate "multiscales" images in consecutively numbered groups starting from 0 (i.e. "0/", "1/", "2/", "3/", ...);
+- SHOULD have OME metadata representing the entire collection of images in a file named "OME/METADATA.ome.xml" which:
+  - MUST adhere to the OME-XML specification but
+  - MAY make use of the minimum specification available at https://docs.openmicroscopy.org/ome-model/6.2.2/specifications/minimum.html;
```
will-moore, line 274: Might be nice to render this as a link?

will-moore (approved):
This looks good to merge?
Do we have any public sample data for bioformats.layout?

sbesson:
Some of the latest 0.4 plates published at https://idr.github.io/ome-ngff-samples/ also include the `bioformat2raw.layout` e.g. https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0056B/7361.zarr

We have not produced any public examples of non-HCS Zarr datasets with this layout. As elicited in https://github.com/ome/ome-zarr-py/issues/210#issuecomment-1173841118 we have limited tooling that can write this layout. From the IDR front, extending https://github.com/ome/omero-cli-zarr/pull/118/ to also support the export of `Fileset:<id>` is the most promising way forward in my opinion. Otherwise, @pwalczysko has produced some examples derived from the public OME-TIFF samples during the testing of OMEZarrReader and we could publish elsewhere

latest/index.bs
will-moore: Also, it could be helpful here to mention that the OME-XML should use `<MetadataOnly/>` instead of `<BinData/>` (if I'm reading that right)?

joshmoore:
As we move towards a merging of this work, it might be worth also considering @melissalinkert's https://github.com/melissalinkert/ngff/commit/056b4e905a0edfaa275fd352888f5b7edd410adb (on top of this PR). It proactively adds a feature that's currently needed by clients of the bioformats2raw output to deal with the loading of large filesets (e.g. HCS) over latent connections (e.g. S3):

```
# in OME/.zattrs
{
  "series" : [ "0", "1", "..." ]
}
```

## Options

### Hold off

* Pro: this PR captures the current state and therefore is useful.
* Con: as currently written, the bump would require going to `bioformats2raw.layout=4`.

### Cherry-pick here

* Pro: Less churn with the spec
* Pro: we could loosen the "MUST" on the integer-based naming of the series ("0", "1", ...)
* Con: needs unification with `bioformats2raw` (or we loosen the MUST requirement on the `series` field)

## Open questions
- Should an error or warning be raised if the various representations of the fileset don't match?
  - the Zarr groups
  - the OME-XML
  - the new `series`

joshmoore:
Followed the cherry-pick strategy from above but updated the semantics to be more permissive of missing "series" indices. cc: @melissalinkert

latest/index.bs
```diff
@@ -242,9 +250,81 @@ keys as specified below for discovering certain types of data, especially images
...
+    - MUST use `<MetadataOnly/>` elements as opposed to `<BinData/>`, `<BinaryOnly/>` or `<TiffData/>`;
+    - MAY make use of the [minimum specification](https://docs.openmicroscopy.org/ome-model/6.2.2/specifications/minimum.html).
+
+Additionally:
+
+- If "OME/METADATA.ome.xml" is present, "OME" MUST be a Zarr group which:
+    - MAY contain a "series" attribute.
+    - "series" MUST be a list of string objects, each of which is a path to an image group.
+    - The order of the paths MUST match the order of both the "multiscales" metadata and the "Image" elements in "OME/METADATA.ome.xml".
+- If if "OME/METADATA.ome.xml" or the "series" attribute do not exist:
```
will-moore, line 312: If if

latest/index.bs
```diff
@@ -242,9 +250,81 @@ keys as specified below for discovering certain types of data, especially images
...
+    - separate "multiscales" images MUST be stored in consecutively numbered groups starting from 0 (i.e. "0/", "1/", "2/", "3/", ...).
+- Every "multiscales" group MUST represent exactly one OME-XML "Image" in the same order as either the series index or the group numbers.
+
+Please note that the "plate" key takes precedence and parsing of such datasets should follow [[#plate-md]]. It is not
+possible to mix collections of images with plates at present.
+
+Conforming readers:
+- SHOULD detect the presence of more than one image and make users aware of the fact;
+- MAY use the "series" attribute in the "OME" group to determine a list of valid groups;
```
will-moore, line 321: These first 2 points seem to say the same thing?
Readers SHOULD detect the presence of more than one image... by using the "series" attribute or the "OME/METADATA.ome.xml ... and make users aware of the fact?

latest/index.bs
```diff
@@ -242,9 +250,81 @@ keys as specified below for discovering certain types of data, especially images
...
+    - MUST adhere to the OME-XML specification but
+    - MUST use `<MetadataOnly/>` elements as opposed to `<BinData/>`, `<BinaryOnly/>` or `<TiffData/>`;
+    - MAY make use of the [minimum specification](https://docs.openmicroscopy.org/ome-model/6.2.2/specifications/minimum.html).
+
+Additionally:
+
+- If "OME/METADATA.ome.xml" is present, "OME" MUST be a Zarr group which:
+    - MAY contain a "series" attribute.
+    - "series" MUST be a list of string objects, each of which is a path to an image group.
+    - The order of the paths MUST match the order of both the "multiscales" metadata and the "Image" elements in "OME/METADATA.ome.xml".
```
will-moore, line 311: I'm not sure I understand `MUST match the order of both the "multiscales" metadata`.
If the paths are "image1", "image2", "image3" and these are all groups that contain multiscales images, how can they be in the wrong order?

joshmoore: Hmmm.... agreed. Not sure what I was thinking. I'll limit this to just saying the series paths order must match the OME-XML series order.

latest/index.bs
joshmoore: I'll limit the first SHOULD to just the "making users aware" (i.e. don't silently hide an image)

joshmoore:
Thanks for the feedback, @will-moore! Changes pushed.

joshmoore:
Likely a last call for comments here. I'm going to move on to the various PRs:

 - [x] https://github.com/ome/ZarrReader/pull/31
 - [ ] https://github.com/glencoesoftware/bioformats2raw/pull/157
 - [ ] https://github.com/ome/ome-ngff-validator/pull/13
 - [ ] https://github.com/ome/ome-zarr-py/pull/174
 - [ ] https://github.com/hms-dbmi/vizarr/issues/149

Note: https://github.com/ome/ome-zarr-metadata will _not_ be used for this release.

latest/index.bs
```diff
@@ -101,6 +101,14 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL
 “RECOMMENDED”, “MAY”, and “OPTIONAL” are to be interpreted as described in
 [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
+<p>
+<dfn>Transitional</dfn> metadata is added to the specification with the
+intention of removing it in the future. Implementations may be expected (MUST) or
+encouraged (SHOULD) support the reading of the data, but writing will usually
```
melissalinkert, line 107:
```suggestion
encouraged (SHOULD) to support the reading of the data, but writing will usually
```

latest/index.bs
```diff
@@ -242,9 +250,87 @@ keys as specified below for discovering certain types of data, especially images
...
+    ├── OME                   # Special group for containing OME metadata
+    │   ├── .zgroup
+    │   ├── .zattrs           # Contains "series" metadata
+    │   └── METADATA.ome.xml  # OME-XML file stored within the Zarr fileset
+    ├── 0                     # First image in the collection
+    ├── 1                     # Second image in the collection
+    └── ...
+</pre>
+
+<h4 id="bf2raw-attributs" class="no-toc">Attributes</h4>
```
melissalinkert, line 281:
```suggestion
<h4 id="bf2raw-attributes" class="no-toc">Attributes</h4>
```

melissalinkert (commented):
Noticed a couple of small typos while reading latest comments from @sbesson.

latest/index.bs
```diff
@@ -242,9 +250,87 @@ keys as specified below for discovering certain types of data, especially images
...
+
+Conforming groups:
+
+- MUST have the value "3" for the "bioformats2raw.layout" key in their `.zattrs` metadata at the top of the hierarchy;
+- SHOULD have OME metadata representing the entire collection of images in a file named "OME/METADATA.ome.xml" which:
+    - MUST adhere to the OME-XML specification but
+    - MUST use `<MetadataOnly/>` elements as opposed to `<BinData/>`, `<BinaryOnly/>` or `<TiffData/>`;
+    - MAY make use of the [minimum specification](https://docs.openmicroscopy.org/ome-model/6.2.2/specifications/minimum.html).
+
+Additionally:
```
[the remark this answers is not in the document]
joshmoore: Agreed. If we can hold off and do a larger clean up that would be my preference. (Especially since we likely want multiple representations)

latest/index.bs
```diff
@@ -242,9 +250,87 @@ keys as specified below for discovering certain types of data, especially images
...
+
+- If "OME/METADATA.ome.xml" is present, "OME" MUST be a Zarr group which:
+    - MAY contain a "series" attribute.
+    - "series" MUST be a list of string objects, each of which is a path to an image group.
+    - The order of the paths MUST match the order of the "Image" elements in "OME/METADATA.ome.xml".
+- If "OME/METADATA.ome.xml" or the "series" attribute do not exist:
+    - separate "multiscales" images MUST be stored in consecutively numbered groups starting from 0 (i.e. "0/", "1/", "2/", "3/", ...).
```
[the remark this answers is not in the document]
joshmoore: Isn't that covered by "plate" taking priority? Does it suffice to re-state that?

joshmoore:
>  Is the proposal still to port this transitional metadata specification only to the current stable 0.4 specification?

Yes. I'll do once all edits are made (Hopefully this morning).

latest/index.bs
```diff
@@ -242,9 +250,87 @@ keys as specified below for discovering certain types of data, especially images
...
 
+"bioformats2raw.layout" (transitional) {#bf2raw}
+------------------------------------------------
+
+[=Transitional=] "bioformats2raw.layout" metadata identifies a group which implicitly describes a series of images.
+The need for the collection stems from the common "multi-image file" scenario in microscopy. Parsers like Bio-Formats
+define a strict, stable ordering of the images in a single container that can be used to refer to them by other tools.
+
+In order to capture that information within an OME-NGFF dataset, `bioformats2raw` internally introduced a wrapping layer.
+The bioformats2raw layout can be found in NGFF versions v0.2, v0.3, and v0.4. An upcoming NGFF specification will replace
```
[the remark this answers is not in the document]
joshmoore: Agreed. Pushed a commit limiting this to v0.4 (for now) and tried to clarify the motivation a bit.

latest/index.bs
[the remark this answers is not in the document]
joshmoore: Screen Shot 2022-09-22 at 08 28 19

joshmoore:
From my POV, I think we are ready to move forward with this PR as well as the bioformats2raw PR itself. Python I think will follow _relatively_ quickly and then we'll look into Javascript.

cc: @melissalinkert @sbesson

@tischi: do you think there is interest in trying to add this to MoBIE now? or wait for the non-transitional version?

tischi:
> do you think there is interest in trying to add this to MoBIE now? or wait for the non-transitional version?

I would be interested adding this to the `hackathon` package in `mobie-io`.

latest/index.bs
```diff
@@ -315,10 +315,11 @@ Conforming groups:
...
 
 - If "OME/METADATA.ome.xml" is present, "OME" MUST be a Zarr group which:
-    - MAY contain a "series" attribute.
-    - "series" MUST be a list of string objects, each of which is a path to an image group.
-    - The order of the paths MUST match the order of the "Image" elements in "OME/METADATA.ome.xml".
+    - MAY contain a "series" attribute. If so:
+        - "series" MUST be a list of string objects, each of which is a path to an image group.
+        - The order of the paths MUST match the order of the "Image" elements in "OME/METADATA.ome.xml".
 - If "OME/METADATA.ome.xml" or the "series" attribute do not exist:
+    - existing "plate" metadata will take precedence if it exists, or
```
melissalinkert, line 322: Does `will` here actually mean `MUST`?

melissalinkert (commented):
Apart from one small clarification, I think this is fine to merge.

latest/index.bs
joshmoore: Hmmm.... I _think_ that's right, but what I was trying to get across is that "the next statement only holds when `"plate"` isn't defined". Is that the same thing as the MUST?

joshmoore: I think what bothers me in the logic is that the "plate MUST take precedence" applies much higher, e.g. even if the XML exists. Let me try to break all the clauses apart.

joshmoore: Primary question from my side is if it's actually:

```
Matching "series" metadata (as described next) SHOULD be provided ...
```

melissalinkert: Latest version is definitely clearer, thank you. I think `Matching "series" metadata (as described next) SHOULD be provided` makes sense, but I don't have strong objections to `MAY`.

joshmoore: If we're unsure, I'll err on the side of the stricter `SHOULD` and then we can downgrade it later.

latest/index.bs
```diff
@@ -242,9 +250,89 @@ keys as specified below for discovering certain types of data, especially images
...
+- MUST have the value "3" for the "bioformats2raw.layout" key in their `.zattrs` metadata at the top of the hierarchy;
+- SHOULD have OME metadata representing the entire collection of images in a file named "OME/METADATA.ome.xml" which:
+    - MUST adhere to the OME-XML specification but
+    - MUST use `<MetadataOnly/>` elements as opposed to `<BinData/>`, `<BinaryOnly/>` or `<TiffData/>`;
+    - MAY make use of the [minimum specification](https://docs.openmicroscopy.org/ome-model/6.2.2/specifications/minimum.html).
+
+Additionally, the logic for finding the Zarr group for each image follows the following logic:
+
+- If "plate" metadata is present, images MUST be located at the defined location.
+    - Matching "series" metadata as described next MAY be provided for tools that are unaware of the "plate" specification.
```
joshmoore, line 318:
```suggestion
    - Matching "series" metadata as described next MAY be provided for tools that are unaware of the "plate" specification.
    - Matching "series" metadata (as described next) SHOULD be provided for tools that are unaware of the "plate" specification.
```

0.4/index.bs
```diff
@@ -240,6 +248,85 @@ keys as specified below for discovering certain types of data, especially images
...
+- MUST have the value "3" for the "bioformats2raw.layout" key in their `.zattrs` metadata at the top of the hierarchy;
+- SHOULD have OME metadata representing the entire collection of images in a file named "OME/METADATA.ome.xml" which:
+    - MUST adhere to the OME-XML specification but
+    - MUST use `<MetadataOnly/>` elements as opposed to `<BinData/>`, `<BinaryOnly/>` or `<TiffData/>`;
+    - MAY make use of the [minimum specification](https://docs.openmicroscopy.org/ome-model/6.2.2/specifications/minimum.html).
+
+Additionally, the logic for finding the Zarr group for each image follows the following logic:
+
+- If "plate" metadata is present, images MUST be located at the defined location.
+    - Matching "series" metadata as described next MAY be provided for tools that are unaware of the "plate" specification.
```
joshmoore, line 316:
```suggestion
    - Matching "series" metadata (as described next) SHOULD be provided for tools that are unaware of the "plate" specification.
```

sbesson (approved):

PR #113, opened 2022-04-19
Consume main branch of ome/spec-prod action

sbesson:
This should include the latest changes allowing to publish the JSON schemas under ngff.openmicroscopy.org

See https://github.com/ome/spec-prod/pull/2 for more details

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
-      - uses: ome/spec-prod@v4
+      - uses: ome/spec-prod@main
         with:
           GH_PAGES_BRANCH: gh-pages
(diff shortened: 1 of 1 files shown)


joshmoore:
MSTM, though I assume nothing is being generated for actual review (though that does make me wonder.... :smile:)

sbesson:
As it stands, I understand the generation will only happen on push to the `main` branch. For the sake of  review, I forced pushed a87990dbd9db7b017deaad0ed0237dac24a737ff to the `gh-pages` branch on my fork and f24828335845ab22791bdb68457ef574e4a0cc7a to the `main` branch on my fork. https://github.com/sbesson/ngff/commit/c4dab4c03c4ee1fcdb0986f08ce9bca8f6f289ce is the resulting commit generated from this change.

Beyond the scope of this PR and if additional work happens on the action, I imagine this diff could be recorded e.g. as a CI artifact to facilitate the review.

PR #115, opened 2022-04-21
Proposal: Add UUID field to multiscale objects

joshmoore:
In order to uniquely identify the images that are being generated,
this change proposes that a UUID should be included with each multiscale.
The UUID can either be generated at creation or perhaps copied from
an existing array if all meta(data) is identical (e.g. a rechunking).

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -295,6 +295,11 @@ Each "multiscales" dictionary:
 
  * SHOULD contain the field "name" which is informative to users but not sufficient for unique identifiation.
 
+ * SHOULD contain the field "uuid" which is a
+   [RFC 4122](https://www.ietf.org/rfc/rfc4122.txt) compliant
+   string representation of a universally unique identifier, e.g., "urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6" and SHOULD use
+   a compliant generator to minimize potential conflicts.
+
 * SHOULD contain the field "version", which indicates the version of the
   multiscale metadata of this image (current version is [NGFFVERSION]).
 
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -288,14 +288,21 @@ It MAY contain exactly one `translation` that specifies the offset from the orig
 The length of the `scale` and `translation` array MUST be the same as the length of "axes".
 The requirements (only `scale` and `translation`, restrictions on order) are in place to provide a simple mapping from data coordinates to physical coordinates while being compatible with the general transformation spec.
 
-Each "multiscales" dictionary MAY contain the field "coordinateTransformations", describing transformations that are applied to all resolution levels in the same manner.
-The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:coordinateTransformations" and are applied after them.
-They can for example be used to specify the `scale` for a dimension that is the same for all resolutions.
+Each "multiscales" dictionary:
+ * MAY contain the field "coordinateTransformations", describing transformations that are applied to all resolution levels in the same manner.
+   The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:coordinateTransformations" and are applied after them.
+   They can for example be used to specify the `scale` for a dimension that is the same for all resolutions.
 
-Each "multiscales" dictionary SHOULD contain the field "name". It SHOULD contain the field "version", which indicates the version of the multiscale metadata of this image (current version is [NGFFVERSION]).
+ * SHOULD contain the field "name" which is informative to users but not sufficient for unique identifiation.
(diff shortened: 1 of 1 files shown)


joshmoore:
To view: http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/joshmoore/ngff/uuid/latest/index.bs

latest/index.bs
```diff
@@ -290,26 +290,42 @@ The requirements (only `scale` and `translation`, restrictions on order) are in
...
+   They can for example be used to specify the `scale` for a dimension that is the same for all resolutions.
+</li>
+
+<li>
+   SHOULD contain the field "name" which is informative to users but not sufficient for unique identifiation.
+</li>
+
+<li>
+   SHOULD contain the field "uuid" which is a
+</li>
```
will-moore, line 306: This `<li>` shouldn't terminate here? (should include the next one too)

joshmoore:
Pushed.

> How is this likely to happen? Not via any JSON schema id or ref, but simply by some user's code or unstructured csv/file/DB/notes ?

How do you mean, @will-moore?

will-moore:
Where will I use the UUID?
E.g. In a description of another NGFF Image `This image was processed from Image "urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6"` or in a JSON `collection: [{"id": "urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6"}, ...]` or is it just not known just now, but it seems like it could be useful in the future?

sbesson:
On the feature itself, I am also in favor or introducing some semantics allowing to identify of these objects. I can imagine several use cases as these objects are made available on object stored and potentially replicated in many places to allow some form of integrity check.

A secondary proposal included this PR is the refactoring the definition of the `multiscales` dictionary to use a bullet-point list with clear MAY/SHOULD/MUST mapping - see http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/joshmoore/ngff/uuid/latest/index.bs#multiscale-md.
The current plate specification uses an vairant with a list of definitions e.g. as in https://ngff.openmicroscopy.org/latest/#plate-md. However, this currently lacks a clear expression requirement level and I have on my todo to review this paragraph.
I do not have a strong preference on one style vs the other (bullet point vs definition list) but I think it would make sense to unify. Adding a bit to the complexity, some of the keys include are themselves dictionaries and the specification needs to define sub-keys. I assume the advantage of the current proposal is that this would be simply represented as indented bullet points

latest/index.bs
```diff
@@ -288,14 +288,50 @@ It MAY contain exactly one `translation` that specifies the offset from the orig
...
+   They can for example be used to specify the `scale` for a dimension that is the same for all resolutions.
+</li>
+
+<li>
+   SHOULD contain the field "name" which is informative to users but not sufficient for unique identifiation.
+</li>
+
+<li>
+  <p>
+   SHOULD contain the field "uuid" which is a
```
thewtex, line 306: Could we also / alternatively use a [content identifier (CID)](https://py-cid.readthedocs.io/en/master/)? This can help prevent duplicates (it can be generated from the data).

thewtex: Nevermind, after further thought, modifying the content will change the CID.

But, it worth noting that with content-addressed storage, there is no need for a uuid.

joshmoore: Guess the question to you is how long until you think content-addressed storage is main stream :)

More seriously, do you think this should be made flexible/configurable so that a CID could be stored in place of a UUID?

melonora:
I would like to present a use case here with regards to view configurations:

I have some work on view configurations that allow reproducing the view across a visualization ecosystem. Below an example of defining in this case `SpatialData` in that config.

[image.png](https://imagesc.zulipchat.com/user_uploads/16804/C9BD2BkPfU-5JZGlXFNwoTyx/image.png)


Without having UUIDs I see some operations as impossible. Starting with the case that is possible without UUIDs

1. I locally have my viewconfigs at the root of the zarr store. Now I deposit the data in a database so the url has changed. In this case, you would not need a UUID as the config is contained in the zarr store and you can just update the path in the config.

Now 2 cases that I think are not possible without UUIDs:

1. I have viewconfigs outside the zarr store. I move the zarr store locally. The URL is not valid anymore. In case of a  UUID in the top-level zarr.json I could still just match the viewconfig to the zarr store by having an api that takes in a viewconfig and a list of zarr containes that have the UUID. After a match update the URL.

2. Same case as 1 but then instead of moving the zarr store locally, deposit it in a database.

This is my specific case at the moment, but I think there are many other cases where this could be useful.

PR #118, opened 2022-05-12
Proposal: use test suite structure

joshmoore:
Alternate proposal to #116 described under https://github.com/ome/ngff/pull/116#issuecomment-1124780956 to make use of the test suite structure from https://github.com/json-schema-org/JSON-Schema-Test-Suite#structure-of-a-test

<details>
<summary>
sciprt used to combine the JSON
</summary>

```
#!/usr/bin/env python

import glob
import json

template = """{{
    "description": "TBD",
    "schema": {{
        "id": "TBD"
    }},
    "tests": [
{tests}
    ]
}}
"""

test = """
  {{
    "formerly": "{filename}",
    "description": "TBD",
    "data": {data},
    "valid": "TBD"
  }}"""


tests = []
for prefix in ("valid*/**", "invalid*/**"):
    for filename in glob.glob(prefix):
        with open(filename) as o:
            try:
                data = json.load(o)
                data = json.dumps(data, indent=4)
            except:
                print(f"Error on {filename}")
                raise
        tests.append(test.format(filename=filename, data=data))

tests = ",\n".join(tests)
output = template.format(tests=tests)
print(json.dumps(
    json.loads(output),
    indent=2))
```

</details>

Here we are not really using the `schema` field correctly, and perhaps we agree to use:

```
"schemas": [ "schema/file_1", "schema/file_2"]
```


latest/tests/strict_image_suite.json
```diff
@@ -0,0 +1,388 @@
+{
+  "description": "TBD",
+  "schema": {
+    "id": "schemas/strict_image.schema"
```
sbesson, line 4: Potentially, we could use `$ref` here https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/bf383b4cf2a0b83957b232d3ba5a92222f632a08/tests/draft2019-09/id.json#L4 ?

latest/tests/test_validation.py
```diff
@@ -1,73 +1,85 @@
 import json
 import glob
 
+from dataclasses import dataclass
+from typing import List
+
 import pytest
 
-from jsonschema import RefResolver, Draft202012Validator
+from jsonschema import RefResolver, Draft7Validator
```
sbesson, line 9: I can't remember why we used the 202012 implementation. Is the goal here to have use the common-denominator? In that case would the version-agnostic `Validator` API be an option?

latest/tests/test_validation.py
```diff
@@ -1,73 +1,85 @@
...
+
+        # Validation
+        for filename in glob.glob("tests/*.json"):
+            with open(filename) as o:
+                suite = json.load(o)
+            schema = suite["schema"]
+            with open(schema["id"]) as f:
+                schema = json.load(f)
+            for test in suite["tests"]:
+                ids.append("validate_" + str(test["formerly"]).split("/")[-1][0:-5])
```
sbesson, line 56: Possibly points to using the `id` key - see https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/master/test-schema.json#L61 ?

latest/examples/multiscales_strict/.config.json
```diff
@@ -0,0 +1,3 @@
+{
+    "schema": "schemas/strict_image.schema"
```
sbesson, line 2: This almost makes me wonder whether this configuration file could be structured like the JSON schema test suite with some relative pointers to the real data e.g.

```json
{
    "description": "TBD",
    "schema": "schemas/strict_image.schema",
    "tests": [
        {
            "description": "TBD",
            "data": {
              "$ref": "multiscales_example.json"
            }
        },
                {
            "description": "TBD",
            "data": {
              "$ref": "multiscales_transformations.json"
            }
        }
    ]
}
```
 
 The primary downside is that we will have to remember to actively register new examples. The current implementation automatically detects them via `glob`

sbesson (changes requested):
In general, the proposal here closely matches what is described in https://github.com/ome/ngff/pull/116#issuecomment-1124780956.

Thanks for pushing a first draft of its implementation. Overall, I am more convinced that it is a valuable step forward as we try to increase the coverage of our entire specification with validation schemas.

A few inline questions which are more RFEs about using more of the upstream test-schema.json features.

Apart from fixing the CI, I assume the main remaining thing is to port the invalid tests and consuming the `valid` key in the test generation? And possibly apply the same changes to `0.4` in a separate PR?

latest/tests/strict_image_suite.json
joshmoore: We could use $ref, and then have a class or function elsewhere that provides the file to ref mapping, yes.

latest/tests/test_validation.py
joshmoore: My local version no longer had the dated Draft so I went with the latest. @will-moore may be able to say more.

latest/tests/test_validation.py
joshmoore: Yes, definitely happy to move away from these for the proper suite ones. For examples, we'll still need to generate.

latest/tests/strict_image_suite.json
joshmoore: Ah, I think I misunderstood. Not sure if adding `"$ref"` is within the intent of the suite writers, but we could certainly use it.

latest/tests/test_validation.py
will-moore: Sorry, don't remember

sbesson:
> Is that not what https://github.com/ome/ngff/pull/118/files#diff-55735767b6ee79dde0d4c0f1f3d88dced97cbeafb3bffc4ed05d514087a3e416R57 is doing?

Ah you're right that both valid and invalid `data` are tested in `image_suite.json`. I got slightly confused because `latest/tests/strict_image_suite.json` has `"valid": "TBD"`. I assume this gets intepreted as `true`?

latest/examples/multiscales_strict/.config.json
sbesson: Fair enough. :+1: for lowering the burden of extraction

latest/tests/strict_image_suite.json
sbesson: Mostly inspired by https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/060caae0dd58e34af0449baec1103606a0ef4977/test-schema.json#L23 which suggests that creating a ref to a meta-schema is not fully out of scope. But I do not advocate for making this PR too complex/hard to integrated

sbesson:
Great, this leave one remaining failed test  `duplicate_axes_name`. This is a tricky one as it the JSON is invalid as per the specification but this belongs to the class of data invalidities that the JSON schema is simply not suited to detect. Possible we just remove this test case to avoid the confusion?

joshmoore:
Looks like we're down to just the one "invalid_but_dont_fail" test:

```
self = Suite(schema={'$schema': 'https://json-schema.org/draft/2020-12/schema', '$id': 'https://ngff.openmicroscopy.org/lates...: '0', 'coordinateTransformations': [{'scale': [0.13, 0.13], 'type': 'scale'}]}], 'version': '0.5-dev'}]}, valid=False)
validator = Draft202012Validator(schema={'$defs': {'axes': {'contains': {'properties': {'name': {'type': 'string'}, 'type': {'enum...ma', '$schema': 'https://json...020-12/schema', 'description': 'JSON from OME-NGFF .zattrs', ...}, format_checker=None)

    def validate(self, validator) -> None:
        if not self.valid:
            with pytest.raises(ValidationError):
>               validator.validate(self.data)
E               Failed: DID NOT RAISE <class 'jsonschema.exceptions.ValidationError'>
```

@will-moore, what would you expect to do here? set `valid = true`?

will-moore:
Agree with @seb. Probably best to simply remove the `duplicate_axes_name` example. It's not an example of valid JSON, and it also can't be used to test the schema, so I think it's confusing to keep it anywhere.

joshmoore:
Thanks, both. That gets us back to green. Other then potentially refactoring some of the code to the top-level to be re-used between versions, I think I'm happy with this as a first round of refactoring.

sbesson (approved):
All good from my side. Happy to handle the conversion of the `0.4` tests as a follow-up

PR #119, opened 2022-05-12
Refactor 0.4 schema validation tests

sbesson:
Ports changes made against the latest version of the specification to consolidate the JSON test using the JSON schema test suite format

See https://github.com/ome/ngff/pull/118 for the corresponding PR against `latest`


joshmoore (approved):
Looks good and generally fine to get this in ASAP and iterate. As with #118 though, I still keep wondering if the test code shouldn't be in a usable library. Conceivably even the duplicated test suites could be shared at the top-level to prevent copy-n-paste but that perhaps is more confusing than it's worth.

PR #122, opened 2022-05-23
Plate/well JSON schemas (latest specification)

sbesson:
Ports all changes from #120 under `latest` including:

- the JSON schemas for the `plate` and `well` specification
- the plate/well JSON test suites
- the extraction of the specification snippets under `examples`

The changes should be identical to the original PR with the only exception being that the version is `0.5-dev` rather than `0.4`


joshmoore (approved):
Sorry! I thought this was actually in. (In fact, I just tried to show it to @will-moore in the mainline.)

PR #127, opened 2022-05-27
Add missing plate and well schemas for 0.1, 0.2, 0.3

will-moore:
As discussed in https://github.com/ome/ome-ngff-validator/pull/3, we can't validate HCS data for v 0.1, 0.2, 0.3 because of missing schemas.

This adds plate schema for 0.2 & 0.3 (copying over the 0.1 schema and updating version)
and adds well schema for 0.1, 0.2, 0.3 (copying the 0.4 schema). This only requires 'path' attribute which is correct for all versions.

cc @sbesson


0.2/schemas/well.schema
```diff
@@ -0,0 +1,47 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ngff.openmicroscopy.org/0.4/schemas/well.schema",
```
sbesson, line 3:
```suggestion
  "$id": "https://ngff.openmicroscopy.org/0.2/schemas/well.schema",
```

0.3/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "http://localhost:8000/plate.schema",
```
sbesson, line 3:
```suggestion
  "$id": "https://ngff.openmicroscopy.org/0.3/schemas/plate.schema"
```

0.3/schemas/well.schema
```diff
@@ -0,0 +1,47 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ngff.openmicroscopy.org/0.4/schemas/well.schema",
```
sbesson, line 3:
```suggestion
  "$id": "https://ngff.openmicroscopy.org/0.3/schemas/well.schema",
```

0.2/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "http://localhost:8000/plate.schema",
```
sbesson, line 3:
```suggestion
  "$id": "https://ngff.openmicroscopy.org/0.2/schemas/plate.schema",
```

0.2/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
...
+  "title": "NGFF Plate",
```
sbesson, line 4:
```suggestion
  "title": "OME-NGFF plate schema",
```

0.2/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
...
+          "description": "Maximum number of fields per view across all wells."
+        },
+        "acquisitions": {
+          "description": "Rows of the Plate grid",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "id": {
+                "type": "number"
```
sbesson, line 81: Specified as `integer` in 0.4/latest schemas

0.2/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
...
+              },
+              "maximumfieldcount": {
+                "type": "number"
```
sbesson, line 84: Specified as `integer` in 0.4/latest schemas

0.2/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
...
+              },
+              "name": {
+                "type": "string"
+              },
+              "description": {
+                "type": "string"
+              },
+              "starttime": {
+                "type": "number"
```
sbesson, line 93: Specified as `integer` in 0.4/latest schemas

0.2/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
...
+              "maximumfieldcount": {
+                "type": "number"
+              },
+              "name": {
+                "type": "string"
+              },
+              "description": {
+                "type": "string"
+              },
+              "starttime": {
```
sbesson, line 92: Should we also have `endttime`?

0.3/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
...
+          "description": "Maximum number of fields per view across all wells."
+        },
+        "acquisitions": {
+          "description": "Rows of the Plate grid",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "id": {
+                "type": "number"
```
sbesson, line 81: Same as above

0.3/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
...
+              },
+              "maximumfieldcount": {
+                "type": "number"
```
sbesson, line 84: Same as above

0.3/schemas/plate.schema
```diff
@@ -0,0 +1,112 @@
...
+              },
+              "name": {
+                "type": "string"
+              },
+              "description": {
+                "type": "string"
+              },
+              "starttime": {
+                "type": "number"
```
sbesson, line 93: Same as above

sbesson (changes requested):
A few comments based on manual comparison with  https://github.com/ome/ngff/blob/main/0.4/schemas/plate.schema and https://github.com/ome/ngff/blob/main/0.4/schemas/well.schema.

Ideally, it would be useful to have a JSON test suite for each version/specification with minimally a compliant example so that we can test the validation logic works.

will-moore:
@sbesson Thanks for the fixes etc. I think the other issues are fixed in [09ab485](https://github.com/ome/ngff/pull/127/commits/09ab485d2869fbae8e3346f0694b399613e79b9e)

will-moore:
I'll add tests...

will-moore:
"feels like an overhead for the scope of this work" -> you think it should be done in this PR, or it outside the scope of this PR?
Pushed fix to 0.1 plate.schema.

sbesson (approved):

sbesson:
> you think it should be done in this PR, or it outside the scope of this PR?

Not worth it imo unless we start adding more content to these old versions schemas.

will-moore:
I'm missing something. Having merged this, I'd expect https://github.com/ome/ngff/tree/main/0.3/schemas to show plate.schema but I don't see it there??

sbesson:
@joshmoore would this be related to the afternoon force push events?

PR #128, opened 2022-05-30
Add Review comment to all PRs.

joshmoore:
Replaces #117 since testing workflows from forks is painful.

On each push to a PR, this workflow should add if necessary or otherwise update a comment pointing to the URLs which can be used for reviewing the changes to latest.bs:

Screen Shot 2022-05-30 at 12 06 29

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -22,5 +22,5 @@ jobs:
           body: |
             #### Automated Review URLs
             * [render latest/index.bs](http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/ome/ngff/${{github.sha}}/latest/index.bs)
-            * [diff latest modified](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fngff.openmicroscopy.org%2Flatest%2F&doc2=http%3A%2F%2Fapi.csswg.org%2Fbikeshed%2F%3Furl%3Dhttps%3A%2F%2Fraw.githubusercontent.com%2Fome%2F%ngff%2F${{github.sha}}%2Flatest%2Findex.bs)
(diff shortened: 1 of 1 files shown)


joshmoore:
Doh. Even on a branch from `ome` this won't be testable. Manually adding the markup:

#### Automated Review URLs
 * [render latest/index.bs](http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/ome/ngff/23b71cb16321bca05d2e4dd8de20c7e8f4f29de7/latest/index.bs)
 * [diff latest modified](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fngff.openmicroscopy.org%2Flatest%2F&doc2=http%3A%2F%2Fapi.csswg.org%2Fbikeshed%2F%3Furl%3Dhttps%3A%2F%2Fraw.githubusercontent.com%2Fome%2F%ngff%2F 23b71cb16321bca05d2e4dd8de20c7e8f4f29de7%2Flatest%2Findex.bs)

sbesson (approved):
As mentioned separately, I almost want this PR with some real content change in order to assess and/or suggest change. Immediate limitation I can think of is that only `latest/index.bs` is considered even if changes are made to other specifications and/or different files are modified. 

No objections to getting this in and start to iterate as the publication of review links should reduce the maintenance burden.

.github/workflows/review.yml
```diff
@@ -0,0 +1,26 @@
+---
+name: Add review url
+
+on:
+  pull_request_target:
```
sbesson, line 5: Should this have been `pull_request` as https://github.com/ome/ngff/pull/112#issuecomment-1141013652 seems to link to the last commit on the base branch as per https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target?

joshmoore: Grrr. The actions can't write to the same PR if only `pull_request` is used. I'll have to look for a way to find the sha regardless. Let me know if we need to revert.

cc: @jburel in case you've run into these issues before.

PR #129, opened 2022-05-30
Review workflow: fix generated URL.

sbesson:
As per https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
github.sha will point to the base branch SHA


joshmoore:
Just doing this to but I don't think a pull_request_target PR will likely not let you appropriately test. That was my problem with the previous attempts.

sbesson:
> Just doing this to but I don't think a pull_request_target PR will likely not let you appropriately test. 

Ah now I get the "This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does.". Yes not sure how to test this now

joshmoore:
Shall we reopen my `origin/review` branch, push your commit there, and open PRs against it?

joshmoore:
I'll move this to `origin/main` and revert if necessary.

joshmoore:
After pushing head.ref this has been updated, but that doesn't resolve, since `ome/ngff/review_sha` doesn't exist. I'll revert to the sha.

PR #130, opened 2022-06-14
image-label: JSON schemas and clarifications

### LINKED ISSUE #125: image-label metadata: invalid JSON in example
The `image-label` [example in the spec](https://ngff.openmicroscopy.org/latest/#label-md) is not a valid JSON. It mixes object and array syntax (`"image-label"` & `"source"` keys, but closing `]` in the last line):
```
"image-label":
  {
    "version": "0.5-dev",
    "colors": [
      {
        "label-value": 1,
        "rgba": [255, 255, 255, 0]
      },
      {
        "label-value": 4,
        "rgba": [0, 255, 255, 128]
      },
      ...
      ],
    "properties": [
      {
        "label-value": 1,
        "area (pixels)": 1200,
        "class": "foo"

      },
      {
        "label-value": 4,
        "area (pixels)": 1650
      },
      ...
      ]
  },
  "source": {
    "image": "../../"
  }
]
```
Also: are the `"image-label"` and `"source"` keys on the same level, or should `"source"` be an optional key in `"image-label"`?
From the text I would have gathered that `"source"` is a child of `"image-label"` but in the example they are on the same level.

sbesson:
Fixes #125 

This PR reviews the `image-label` specification and brings it inline with recent improvements made to other specifications including:

- the addition of JSON schemas, a minimal schema enforcing all MUST requirements and a strict schema enforcing all SHOULD requirements, together with tests
- the extraction of the  embedded JSON snippets as standalone examples and their correction to be compliant with the schema
- an update to the specification paragraph to clarify the content and requirement levels of all keys defined by the specification

In addition to reviewing the validation tests keep passing and that the modified text is sensible, the two biggest changes that should be reviewed, notably by the original authors of the specification (all marked as reviewers of this PR) are: 

- the current specification  defines `colors` and `properties` keys with no requirement level and does not define the `version` key. In addiition to clarifying the expected content, this PR proposes a set of requirement levels using the keywords from RFC2119 and implemented through the JSON schemas
- the name of the JSON schema is initially set to `strict_image-label.schema` to follow the convention used for other JSON schemas while keeping the hyphen separator using in `image-label`. The mixture of underscore and hyphen is most likely confusing. Should we change to underscores? 
 Update (2022-07-05): updated the proposal to use `label.schema` and `strict_label.schema`


0.4/index.bs
```diff
@@ -375,63 +375,37 @@ in which each unique pixel value represents a separate segmented object.
...
+who choose to not throw an error SHOULD ignore all except the _last_ entry.
+
+Some implementations MAY represent overlapping labels by using a specially assigned
 value, for example the highest integer available in the pixel range.
 
-The `properties` key defines a list of JSON objects which also describes the unique
-label values. Each entry in the list MUST contain the key "label-value" with the
-pixel value for that label. Additionally, an arbitrary number of key-value pairs
+The `image-label` dictionary MAY contain a `properties` key whose value MUST be a
+ist of JSON objects which also describes the unique label values. Each property object
```
k-dominik, line 390:
```suggestion
list of JSON objects which also describes the unique label values. Each property object
```

joshmoore:
> Automated Review URLs

Argh. Again not working.

Otherwise, :+1: though I agree that the schema name is not particularly friendly.

sbesson:
> Otherwise, 👍 though I agree that the schema name is not particularly friendly.

Open to alternative suggestions but my concern is that this might force us to modify the name of existing published schemas - https://github.com/ome/ngff/tree/main/0.4/schemas.

0.4/examples/image-label_strict/colors_properties.json
```diff
@@ -0,0 +1,29 @@
+{
+  "image-label": {
+    "version": "0.4",
+    "colors": [
+      {
+        "label-value": 1,
+        "rgba": [255, 255, 255, 1]
```
will-moore, line 7: The `alpha: 1` here makes it look like the `alpha` is a fraction (as in css) which isn't what the spec defines. `255` would make a better (less confusing) example.

0.4/schemas/image-label.schema
```diff
@@ -0,0 +1,77 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://ngff.openmicroscopy.org/0.4/schemas/image-label.schema",
+  "title": "OME-NGFF well schema",
```
will-moore, line 4: well -> image-label

0.4/schemas/image-label.schema
```diff
@@ -0,0 +1,77 @@
...
+                  "type": "integer",
+                  "minimum": 0,
+                  "maximum": 255
+                },
+                "minItems": 4,
+                "maxItems": 4
+              }
+            },
+            "required": [
+              "label-value"
```
will-moore, line 34: Is it not required to have `rgba` in the colors object? If you don't have a color for a particular label, you probably shouldn't have a `color` object for that value?

sbesson: That was my initial thought but the existing specification says `Additionally, the "rgba" key MAY be present`

So making it required would be a breaking change

latest/index.bs
```diff
@@ -377,63 +377,37 @@ in which each unique pixel value represents a separate segmented object.
...
-pixel value for that label. Additionally, the "rgba" key MAY be present, the
-value for which is an RGBA unsigned-int 4-tuple: `[uint8, uint8, uint8, uint8]`
-All `label-value`s must be unique. Clients who choose to not throw an error
-should ignore all except the _last_ entry.
-
-Some implementations may represent overlapping labels by using a specially assigned
+The `image-label` dictionary SHOULD contain a `colors` key whose value MUST be a
+list of JSON objects describing the unique label values. Each color object MUST
+contain the `label-value` key whose value MUST be an integer specifying the
+pixel value for that label. It MAY contain an `rgba` key whose value MUST be an tuple
```
will-moore, line 383: `a tuple` (not 'an') - Same as for 0.4/index.bs. Actually is this really a 'tuple'? In JSON you can only have a list?
"tuple" is also used in the schemas rgba descriptions

0.4/examples/image-label_strict/colors_properties.json
sbesson: But the second item in the array with alpha value `128` should make this unambiguous no?
Easy enough to invert and/or change values

latest/index.bs
sbesson: Yes, that's definitely a Python-ism. Proposing to use `array` everywhere which is the  JSON type and also matches the underlying schema?

will-moore:
LGTM 👍

sbesson:
Re the remaining sticking point (the name), considering simply naming the schemas as `label.schema` and `strict_label.schema`.

will-moore:
Those names sound fine to me 👍

joshmoore (approved):
* Generally :+1:
 * I imagine there are some downstream uses that might need updates: e.g. https://github.com/ome/ome-ngff-validator/blob/0985fcc2d360c9cf82b51dbefba1320a885021a0/src/JsonKeyLink.svelte#L20
 * For colors/labels/etc., I assume we are increasingly approaching a reworking.

Issue #71, opened 2021-11-02
document client feature support


will-moore:
As the spec evolves, it's increasing likely that there will be parts of it that are not supported by all the clients, either because of technical limitations or the work just hasn't been done/released.

It would be really nice for new and existing users to be able to see what is supported by which clients, or what's NOT supported (without having to search the issues of each repo). This would also be a nice way to show what clients are available to view OME-NGFF data.

I'm thinking of something like a support table, similar to e.g. https://caniuse.com/flexbox (although in that case it's a table for a single feature, with different versions of each tool). I'd suggest we limit it to documenting only the latest release of each tool.

| viewer | Z-downsampling | omero info | multiscales factor not=2 | URL (not s3) | v0.3 axes | 3D view | labels | HCS plate |
|-------|------------------|------------|-------------------------|--------------|-----------|---------|-------|-----|
| napari | y | y | y | y | y | y (1) |  y | y |
| vizarr | [n](https://github.com/hms-dbmi/vizarr/pull/71) | y | [n](https://github.com/hms-dbmi/vizarr/issues/101) | y | y | n | n| y |
|MoBIE | y | n | y | [n](https://github.com/mobie/mobie-viewer-fiji/issues/351) | n | y | y | n|

1) napari 3D view only supported for lowest level of multiscales pyramid

If this looks like a popular idea, where would this doc/table live? Include it in the spec page, or a doc page on this repo?

joshmoore:
cc: @manzt @constantinpape

constantinpape:
This would be great to have!

manzt:
Agreed, great idea!

PR #138, opened 2022-09-15
Coordinate systems and new coordinate transformations proposal

bogovicj:
This PR has three main contributions:
* The [existing axis specification](https://ngff.openmicroscopy.org/latest/#axes-md) into `coordinateSystem`s - named collections of axes. 
* `coordinateTransformations` now have "input" and "output" coordinate systems.
* Adds many new useful type of coordinate transformations.
  *  informative [examples](https://github.com/bogovicj/ngff/tree/coord-transforms/latest/examples/transformations) of each type are now given in the spec.

See also:
* https://github.com/ome/ngff/issues/84 
* https://github.com/ome/ngff/issues/94 
* https://github.com/ome/ngff/issues/101

PR #139, opened 2022-09-29
Ngff tools table

will-moore:
Started to look again at #71

View page at https://github.com/will-moore/ngff/blob/ngff_tools_table/ngff-tools.md

Aims of this table/documentation:

- Allow data producers and consumers to know whether a particular feature is supported in their viewer of choice
- Document the various viewers available that support OME-NGFF 
- Link to sample data that can be used for testing a particular feature - NB: some overlap here with https://idr.github.io/ome-ngff-samples/

Created a top-level `ngff-tools.md` page.
Tried tables (duplicates) in markdown format or html.
The current table is for "viewers" but I think we could possibly list other tools for producing OME-NGFF too (bioformats2raw, ome-ngff, omero-cli-zarr etc).

First questions:
 - Is this a suitable location and page format?
 - Prefer tables in markdown, but with data in a different file (e.g. yaml or csv) - but how & where to build/publish etc?
 - Also I think that the last table with the axes switched works better since it allows lots of features (rows) and we don't need so many columns (tools)
 - FYI: compare with web browser support of various JS/HTML/CSS features, with 1 table per feature: e.g. https://caniuse.com/arrow-functions 
 - Other viewers to include?

cc @manzt @constantinpape @joshmoore @sbesson

### DIFF
--- a/ngff-tools.md
+++ b/ngff-tools.md
@@ -161,7 +161,8 @@ The following versions of each viewer were used in testing:
     </tr>
     <tr>
       <td>multiscales coordinateTransformations (v0.4)</td>
-      <td><a href="https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457539.zarr">13457539.zarr</a></td>
+      <td><a href="https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457539.zarr">13457539.zarr</a>
+       translate onto <a href="https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457227.zarr">13457227.zarr</a></td>
       <td>n</td>
       <td>n</td>
       <td>?</td>
--- a/ngff-tools.md
+++ b/ngff-tools.md
@@ -83,7 +83,12 @@ The following tables document which features of the OME-NGFF spec are supported 
     </tr>
   </thead>
   <tbody>
-  <th>omero info</th>
+    <tr>
+      <td>omero info</td>
+      <td>y</td>
+      <td>y</td>
+      <td>n</td>
+    </tr>
     <tr>
       <td>Z-downsampling</td>
       <td>y</td>
--- a/ngff-tools.md
+++ b/ngff-tools.md
@@ -134,7 +134,7 @@ The following tables document which features of the OME-NGFF spec are supported 
       <td>n</td>
     </tr>
     <tr>
-      <td>datasets.coordinateTransformations (v0.4)</td>
+      <td>datasets coordinateTransformations (v0.4)</td>
       <td><a href="https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr">13457537.zarr</a>
         translate onto <a href="https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457227.zarr">13457227.zarr</a></td>
       <td>y (scale, translate)</td>
@@ -145,7 +145,7 @@ The following tables document which features of the OME-NGFF spec are supported 
       <td>?</td>
     </tr>
     <tr>
-      <td>multiscales.coordinateTransformations (v0.4)</td>
+      <td>multiscales coordinateTransformations (v0.4)</td>
       <td><a href="https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457539.zarr">13457539.zarr</a></td>
       <td>n</td>
       <td>n</td>
--- a/ngff-tools.md
+++ b/ngff-tools.md
@@ -1,7 +1,9 @@
 # OME-NGFF tools
 
 Various tools are available for viewing and working with OME-NGFF data.
-The following table documents which features of the OME-NGFF spec are supported by which tools:
+The following tables document which features of the OME-NGFF spec are supported by which tools.
+
+## Viewers
 
 | viewer | Z-downsampling                                  | omero info | multiscales factor not=2                           | URL (not s3)                                               | v0.3 axes | 3D view | labels | HCS plate |
 | ------ | ----------------------------------------------- | ---------- | -------------------------------------------------- | ---------------------------------------------------------- | --------- | ------- | ------ | --------- |
@@ -10,7 +12,6 @@ The following table documents which features of the OME-NGFF spec are supported 
 | MoBIE  | y                                               | n          | y                                                  | [n](https://github.com/mobie/mobie-viewer-fiji/issues/351) | n         | y       | y      | n         |
 
 1. napari 3D view only supported for lowest level of multiscales pyramid
-
 
 <table>
   <thead>
@@ -24,17 +25,19 @@ The following table documents which features of the OME-NGFF spec are supported 
       <th>3D view</th>
       <th>labels</th>
       <th>HCS plate</th>
+      <th>bioformats2raw.layout</th>
     </tr>
   </thead>
   <tbody>
     <tr>
-      <td>napari</td>
+      <td><a href="https://github.com/ome/napari-ome-zarr/">napari-ome-zarr</a> (1)</td>
       <td>y</td>
       <td>y</td>
       <td>y</td>
       <td>y</td>
       <td>y</td>
-      <td>y (1)</td>
+      <td>y (2)</td>
+      <td>y</td>
       <td>y</td>
       <td>y</td>
     </tr>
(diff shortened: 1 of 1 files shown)


joshmoore:
> * Is this a suitable location and page format?

I think we are approaching the point (see also https://github.com/ome/ngff/issues/140) where we need to decide if _everything_ is going into the spec page or if we're going to host one or more additional sites. If so, then we should gather thoughts (like https://ome.github.io/ngff)

> * Prefer tables in markdown or html? I think I prefer html: easier to manage with more complex tables

If we end up using markdown, then I'd suggest a data file (like in yaml)

> * Other viewers to include?

- neuroglancer
- webknossos

sbesson:
> I think we are approaching the point (see also https://github.com/ome/ngff/issues/140) where we need to decide if everything is going into the spec page or if we're going to host one or more additional sites. If so, then we should gather thoughts (like https://ome.github.io/ngff)

This also echoes some of the questions I raised in https://github.com/ome/ngff/pull/138#pullrequestreview-1126499068 re hosting and maintaining different but complementary types of documentation.

thewtex:
Could we also add [itkwidgets](https://itkwidgets.readthedocs.io/en/latest/)? The current values are:

```
      <td><a href="(https://itkwidgets.readthedocs.io/en/latest">itkwidgets</a></td>
      <td>y</td>
      <td><a href="https://github.com/InsightSoftwareConsortium/itkwidgets/issues/546">n</a></td>
      <td>y</td>
      <td>y</td>
      <td>y</td>
      <td>y</td>
      <td<a href="https://github.com/InsightSoftwareConsortium/itkwidgets/issues/547">n</a>></td>
      <td>n</td>
      <td>n</td>
```

Also add a column for v0.4 metadata?

normanrz:
For webKnossos:
```
      <td><a href="https://webknossos.org">webKnossos</a></td>
      <td>y</td>
      <td>n</td>
      <td>y</td>
      <td>y</td>
      <td>y</td>
      <td>y</td>
      <td>y</td>
      <td>n</td>
      <td>n</td>
```

We also maintain the [`webknossos` Python package](https://docs.webknossos.org/webknossos-py/) that can be used to create, downsample, compress, rechunk OME-NGFF datasets.

will-moore:
Thanks @normanrz - I've been playing with webknosses for the first time, importing an OME-NGFF image from https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0062A/6001240.zarr, which is a 2-channel image with labels. But I can only see the first channel imported into webknosses. Is there a way to import multiple layers from the one source?

normanrz:
Hey @will-moore, so far we only support single-channel images, but support for multi-channel is coming next month. Labels should work. Perhaps you're running into a bug here. I'll investigate.

thewtex:
>I had an issue while testing the itkwidgets in a conda env on Mac M1 See https://github.com/InsightSoftwareConsortium/itkwidgets/issues/548

Thanks for the report. I followed up there. Here is what I get on my Mac M1:

[OME-NGFF-Bra… - JupyterLab.webm](https://user-images.githubusercontent.com/25432/194921486-b3aec198-076f-4095-8d10-06f4f62ee389.webm)

itkwidgets 1.X is built around OME-NGFF. But, maybe we should clarify:

``` 
<td><a href="(https://itkwidgets.readthedocs.io/en/latest">itkwidgets 1.X</a></td>
```

will-moore:
@normanrz Thanks. Here's another I tried - I wanted to see if the `coordinateTransformations translate` was read by webknosses. Here's the [image url](https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr) or you can see it [validates in ome-ngff-validator](https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr) and it is [viewable in vizarr](https://hms-dbmi.github.io/vizarr/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr).

The response in `exploreRemote` POST included:

```
Error when reading https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr as OME NGFF Zarr v0.4: Failed to read OME NGFF header at https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr/.zattrs <~ Failed to validate json against data schema <~ Validating Json Failed: {\"obj.multiscales[0].datasets[0].coordinateTransformations[1].scale\":[{\"msg\":[\"error.path.missing\"],\"args\":[]}]}
```
The `obj.multiscales[0].datasets[0].coordinateTransformations[1]` doesn't contain `scale` since it is 
```
"translation" : {"0": 0 , "1": 0 , "2": 1.2 , "3": 52.109135 , "4": 60.88427} ,
"type": "translation"
```

normanrz:
Thanks for testing these different samples!
It is true, that we don't support translation yet. It is not straightforward to implement in our current architecture, but we'll get back to it once we implement more advanced transformations in webKnossos next year. 
This particular dataset is interesting, because only one of the multiscales contains a translation. What would be the semantics of that?

will-moore:
@joshmoore @sbesson a few additions (to the last table) and updated the description.
Mostly need to decide how/where to deploy etc.
Feedback here, or do we need a chat?

will-moore:
@normanrz as I understand it, the webknosses viewer only reads WKW format and only talks to the webknosses server. So it's the webknosses server that can read OME-NGFF, rather than the viewer directly reading zarr? Did I get that right?

normanrz:
> @normanrz as I understand it, the webknosses viewer only reads WKW format and only talks to the webknosses server. So it's the webknosses server that can read OME-NGFF, rather than the viewer directly reading zarr? Did I get that right?

webKnossos is an integrated system, so I wouldn't make the distinction that one part supports OME-NGFF and the other doesn't. It is not really possible to use the frontend or backend independent of each other. The backend reads a bunch of different formats, including WKW, Zarr, and N5. The frontend requests chunks from the backend and doesn't care about what metadata governs the chunk storage. This is actually an internal API, which supports multiple chunks per request, and not WKW.

ngff-tools.md
```diff
@@ -0,0 +1,198 @@
...
+    </tr>
+    <tr>
+      <td>datasets coordinateTransformations (v0.4)</td>
+      <td><a href="https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457537.zarr">13457537.zarr</a>
+        translate onto <a href="https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457227.zarr">13457227.zarr</a></td>
+      <td>y (scale, translate)</td>
+      <td>n</td>
+      <td>?</td>
+      <td>y (scale)</td>
+      <td>y (scale)</td>
```
pwalczysko, line 149: This line, which says ``webknossos opens ?`` should say something like 

```
errors for the first image (with scale and transform)
opens for the second image (with only scale)
```
This invites a split of this line into the first image (scale & transform) and the second image (scale).

We also need a better expression for the ``n`` which at present means "does not support the feature which the particular image is trying to highlight, it opens the image but ignores the feature".

jburel:
We should also probably consider expanding https://idr.github.io/ome-ngff-samples/ and add the viewers icons to that table too

will-moore:
@jburel What do you mean "add the viewers icons to that table"? There are icons (links) for vizarr and itk-viewer where we have useful links. But we don't have links for the other viewers. So there's no meaning to icons for them either?

jburel:
I meant if for example it is supported in Mobie, we could add the mobie icon and link to either the guide of an example script.
It will not launch the viewer necessarily. As it stands it might be interpreted as the list of the viewers capable of opening such file which is obviously not the case.

will-moore:
We have done a bunch of testing and it's time to update the table but...

 - Editing the HTML is getting painful
 - The table in markdown doesn't allow any styles etc. https://github.com/will-moore/ngff/blob/ngff_tools_table/ngff-tools.md
 - We also want the ability to do pop-ups, footnotes etc similar to e.g. https://caniuse.com/dialog

So this *really* needs to be somewhere it can be deployed. 
Options:
 - This repo: I don't know how to deploy this with gh-pages, since that is what already handles the bikeshed build etc. 
 - https://github.com/IDR/ome-ngff-samples/ - This is also kinda painful to preview any builds with jekyll, and isn't really the right place
 - https://github.com/ome/ome-ngff-validator - This has great preview builds via netlify, but is also probably not the right place
 - A new repo! E.g. ome/ome-ngff-tools (since we might also want to list creation/validation tools as well as viewers)

This decision is really a blocker at the moment, so to make progress I might just create a new repo or pick one of those above in the absence of any other decision...

joshmoore:
Thanks to everyone for all the contributions here! It was exciting to catch up on all the back and forth. Just a heads up that we decided today to migrate this work to its own repo so that we have more control over the rendering of the various cells. Hopefully that will also attract others to participate as well!

🙌🏽

will-moore:
@joshmoore Moved to
 https://will-moore.github.io/ome-ngff-tools/ (work in progress - old table still there for now for comparison)
built from yaml at
 https://github.com/will-moore/ome-ngff-tools/blob/main/docs/_data/features.yml

joshmoore:
> built from yaml at
> https://github.com/will-moore/ome-ngff-tools/blob/main/docs/_data/features.yml

:heart: _Hugely_ readable.

will-moore:
So the table at https://will-moore.github.io/ome-ngff-tools/ is getting mostly complete now, just need to look at neuroglancer...

@jburel @joshmoore next steps? Does the table 

 - add neuroglancer
 - fix "units" -> "unit" and re-test
 - get feedback from tool authors to check accuracy / create issues etc.
 - move repo to ome?

joshmoore:
> move repo to ome?

:+1:

PR #148, opened 2022-10-09
[troubleshooting: ignore for now] Get CI rendering working 

ivirshup:
Getting CI rendering working. First commit is empty to do some troubleshooting.

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -21,6 +21,6 @@ jobs:
           issue-number: ${{ github.event.pull_request.number }}
           body: |
             #### Automated Review URLs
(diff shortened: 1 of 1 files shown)


ivirshup:
It looks like I can't actually modify the CI from here, so I'm going to close this.

joshmoore:
@ivirshup: when trying to create review.yml I definitely had to do all testing on my own branch due to the security constraints around `pull_request_target`.

That being said, I think your suggestion of RTD is likely the easiest to swing assuming we can adapt (or just rewrite) https://github.com/ome/ngff/blob/main/.github/workflows/publish.yml#L20 to that purpose.

PR #149, opened 2022-10-10
Coordinate transforms json-schema (request for feedback)

ivirshup:
`json-schema` for coordinate transforms proposal #138.

This branch is based off the spec PR. The spec was only modified to move inline examples to the `examples` directory. All other changes should be to files under:

```
latest/examples
latest/schemas
latest/tests
```

This isn't 100% done, but I would appreciate feedback on the implementation and structure of the schemas.

### The plan

@bogovicj and I would like to get feedback on the schema here. At a later point we would like to merge it with the coordinate transform spec.

What's left to do:

- [ ] coordinates and displacements (currently waiting on spec question at https://github.com/ome/ngff/pull/138#discussion_r990802343)
  - A few dependent test cases are currently from this
- [ ] bijections (test case uses coordinates and displacements)
- [ ] `"description"` fields. These should be added, but likely should just be copied from the spec.
- [ ] Multiscales schema should be updated

### Questions from me

* As structured, this is not extensible. Transformations are explicitly allowed. Suggestions for how to make this extensible, while not being too permissive are welcome.
* How could I test that arrays and multiscales must have coordinate transform metadata stored on them (See spec: 027b55bdeff20f0e3634c903923faf900a33cb29)? All current example just test on stand-alone metadata documents. It would be good to be able to validate an entire store. How could this be achieved? The consolidated metadata representation of a store?

### Current state of tests

Since the CI isn't currently running, I figured I'd show the state of the tests here:

<details>
<summary> pytest output </summary>

```bash
$ pytest -v --tb=no
===================================================== test session starts =====================================================
platform darwin -- Python 3.9.12, pytest-7.1.2, pluggy-1.0.0 -- /usr/local/opt/python@3.9/bin/python3.9
cachedir: .pytest_cache
rootdir: /Users/isaac/github/ngff/latest
plugins: xdist-2.5.0, assume-2.4.3, nunit-1.0.0, anyio-3.4.0, profiling-1.7.0, forked-1.3.0, cov-3.0.0
collected 59 items                                                                                                            

tests/test_validation.py::test_run[validate_multiscales_example] PASSED                                                 [  1%]
tests/test_validation.py::test_run[validate_multiscales_transformations] PASSED                                         [  3%]
tests/test_validation.py::test_run[validate_image_metadata] PASSED                                                      [  5%]
tests/test_validation.py::test_run[validate_image] PASSED                                                               [  6%]
tests/test_validation.py::test_run[validate_image_omero] PASSED                                                         [  8%]
[...]

PR #152, opened 2022-10-20
Add identity to image coordinateTransformations schema

thewtex:


constantinpape (commented):
No objections on a technical level. I think we initially decided that we don't need an explicit representation of the identity, are there use-cases where we need it now?
(sorry, I didn't have time to keep up with the discussions around the transformation spec.)

ivirshup:
@thewtex, what's the relationship here to #149?

joshmoore:
@ivirshup: my understanding is that this is solely focused on adding the schema for the previous version (0.4). I assume you might have the identical change for `/latest`?

ivirshup:
@joshmoore ah, hadn't noticed the difference in files changed. Thanks for the heads up.

thewtex:
> No objections on a technical level. I think we initially decided that we don't need an explicit representation of the identity, are there use-cases where we need it now? 

This came up when `coordinateTransformations: []` was present. The schema currently complains that there must be at least one transformation. Additionally, we could also update the schema so that the list can be empty. However, I think the approach to list an Identity transformation is a good explicit way to indicate that this is intentional.

> @thewtex, what's the relationship here to #149?
> @ivirshup: my understanding is that this is solely focused on adding the schema for the previous version (0.4).

Correct, the `identity` transformation already exists in the documentation for v0.4, but it is not supported by the schema.

clbarnes:
Could we instead remove identity from the schema and get rid of the requirement that there be at least one coordinate transformation? Requiring that there be an empty array (as opposed to not including the field) is enough to remind people that there is meant to be some mapping.

thewtex:
@clbarnes that would be a functional way to indicate that there is an identity coordinateTransformation, too. Removing identity would be a change to the already published spec, though.

I would like one of the proposed solutions, a) support for an explicit identity transformation, b) support for an empty `coordinateTransformation[]`, that address the situation of an identity transformation to be integrated so I can use the schema with v0.4 datasets. @constantinpape @joshmoore do you have any guidance on this?

A required, empty `coordinateTransformation[]` is a good idea, but, as an additional requirement, probably a candidate for v0.5.
@bogovicj what will work best and what do you recommend considering the current v0.5 draft?

joshmoore:
> do you have any guidance on this?

My tendency would be to do the thing _most like the 0.4 spec text_ for the moment.

thewtex:
> My tendency would be to do the thing most like the 0.4 spec text for the moment.

Since this is following the 0.4 spec, can we merge this? And make additional improvements as needed in 0.5?

clbarnes:
> Since this is following the 0.4 spec

By my reading, the identity transform is not allowed anywhere in v0.4, even though it is defined. From the 0.4 schema:

> Each "datasets" dictionary ... The transformations are defined according to [§ 3.3 "coordinateTransformations" metadata](https://ngff.openmicroscopy.org/0.4/#trafo-md). **The transformation MUST only be of type translation or scale.**

> Each "multiscales" dictionary ... . **The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:coordinateTransformations"** and are applied after them.

d-v-b:
> > Since this is following the 0.4 spec
> 
> By my reading, the identity transform is not allowed anywhere in v0.4, even though it is defined. From the 0.4 schema:
> 
> > Each "datasets" dictionary ... The transformations are defined according to [§ 3.3 "coordinateTransformations" metadata](https://ngff.openmicroscopy.org/0.4/#trafo-md). **The transformation MUST only be of type translation or scale.**
> 
> > Each "multiscales" dictionary ... . **The transformations MUST follow the same rules about allowed types, order, etc. as in "datasets:coordinateTransformations"** and are applied after them.
> 
> The same is true in `latest`. The thing most like the 0.4 text, then, is to just scrub the single mention of the identity transform which can't be included in the metadata anyway.
> 
> As written, the way to express an identity transform is a trivial scale (a single scale is required) so that's probably the right place here). If that wasn't explicit enough, I think support for a zero-length (but extant) `coordinateTransformations` array gets the point across, while also simplifying the JSON structure (fewer characters), the schema (no length constraints, no required scale), and the spec (fewer transforms, no required scale, and fewer optional keys).

+1 to this. I don't think an explicit identity transformation can make any sense in 0.4 without making a lot of changes to the spec.

thewtex:
The 0.4 spec does include the identity specification:

image

and ensures it is in the schema.

A scale consisting of 1.0's is not the same -- interpreting software handles it differently. And identity means there are not extra computations.

A zero-length `coordinateTransformation` [is not allowed in the 0.4 schema](https://github.com/ome/ngff/blob/main/0.4/schemas/image.schema#L169).

This backward-compatible update to the schema ensures that the text and the schema are the consistent.

Removing a transformation in the spec is not backwards compatible.

d-v-b:
@thewtex if I understand things correctly, there has _never_ been a way to use the Identity transform in v0.4, because of the text [here](https://github.com/ome/ngff/blob/99b8765c0f1b9aa6072b030c8f15f75c082a1e5e/0.4/index.bs#L370) and [here](https://github.com/ome/ngff/blob/99b8765c0f1b9aa6072b030c8f15f75c082a1e5e/0.4/index.bs#L377). In the two places were the spec describes uses of `coordinateTransformations`, it _explicitly_ says that only `scale` and `translation` are allowed.

To rectify this there are two options. The first option, supported by me and @clbarnes, is to remove references to the Identity transform completely, and update the spec to state that an empty collection of `coordinateTransformations` is to be interpreted as expressing an identity transformation. This would technically not be a breaking change, because it was impossible to use the identity transform anyway. 

The other option is this PR, but this still requires additional changes to the spec: we would need to remove [this sentence](https://github.com/ome/ngff/blob/99b8765c0f1b9aa6072b030c8f15f75c082a1e5e/0.4/index.bs#L370C61-L370C126) to allow use of an explicit identity transform. Also, unlike removing the explicit identity transform, your PR is a breaking change -- in v0.4, only scale and translation are allowed, so after this change, ome-ngff readers will need to be updated to handle datasets using the identity transform.

clbarnes:
> The 0.4 spec does include the identity specification:

Could you find where in the spec you can actually use this transform? I couldn't. The only places coordinateTransforms are used is in the multiscale definition; however, in multiscales, coordinateTransformations MUST either be a scale or a translation, i.e. must not be an identity.

So exactly as Davis says, we know what an identity transform is, but we're not allowed to use it. No current implementation can support identity transforms because there is nowhere they can use them while being compliant with the spec. Our options, then, are

- Allow the use of an identity transform, which necessitates adding it to the schema
    - Why? What purpose does it serve?
    - This requires updating implementations
- Remove the definition of the literally useless transformation which we're already not allowed to use anyway
    - This simplifies the spec, simplifies the schema, and requires no change in implementations

thewtex:
> The transformations are defined according to [[#trafo-md]]. The transformation MUST only be of type `translation` or `scale`.

This line is inconsistent. `[[#trafo-md]]` refers to `identity`, `translation`, and `scale`. 

Identity transforms serve a purpose -- they indicate that the transformation is the identity. This is different than an undefined transformation.

There are inconsistencies with the specification, which this patch looks to resolve. I will push an update so the transformation list is consistent. The spec does include an identity transformation. These lines do exist:

https://github.com/ome/ngff/blob/99b8765c0f1b9aa6072b030c8f15f75c082a1e5e/0.4/index.bs#L337
https://github.com/ome/ngff/blob/99b8765c0f1b9aa6072b030c8f15f75c082a1e5e/0.4/index.bs#L341

And there are implementations that use it, which is why I created this patch. If there are implementations that do not support `identity`, I am happy to create patches for them.

d-v-b:
Thanks @thewtex for removing that inconsistent sentence. 

> Identity transforms serve a purpose -- they indicate that the transformation is the identity. This is different than an undefined or unknown transformation.

I don't think anyone is opposing identity transforms. I think the question is how they should be represented in the spec.  

What are the advantages of an identity transform expressed as an object, vs representing the identity transform implicitly as an empty list? Speaking for myself, I like the idea of keeping the spec simple -- making the identity transform implicit rather than explicit would allow me to delete some code, which is a small win. But perhaps there is some advantage to the explicit identity transform that I'm not aware of?

0.4/index.bs
```diff
@@ -367,7 +367,7 @@ to the current zarr group. The "path"s MUST be ordered from largest (i.e. highes
 
 Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
 Each dictionary in "datasets" MUST contain the field "coordinateTransformations", which contains a list of transformations that map the data coordinates to the physical coordinates (as specified by "axes") for this resolution level.
-The transformations are defined according to [[#trafo-md]]. The transformation MUST only be of type `translation` or `scale`.
+The transformations are defined according to [[#trafo-md]]. The transformation MUST only be of type `identity`, `translation` or `scale`.
```
d-v-b, line 370: What if we remove "The transformation MUST only be of type `identity`, `translation` or `scale`." entirely, and just say "The elements of `coordinateTransformations` are the transformations defined in [[#trafo-md]]"? This way, the transformations section of the spec is the single source of truth.

d-v-b:
This has been dormant for a few months. How can we move this forward?

thewtex:
> What are the advantages of an identity transform expressed as an object, vs representing the identity transform implicitly as an empty list? 

When transformations are components of a multi-modal registration analysis workflow, there are often a chain of transformations that represent stages of registration, e.g. a calibration step, a re-orientation step, a template registration step, an atlas registration step. For some modalities or datasets in an analysis, a step may be the identity transformation, and it is useful to indicate that. An empty list does not capture that.

> Right, but there are 2 ways to resolve the inconsistency. This resolution (adding references to identity in several places) requires changes to spec-compliant implementations

If the spec calls out "identity", which it does, an implementation that does not support "identity" is not "spec-compliant".

0.4/index.bs
thewtex: :+1: yes added in 1a55fa1.

clbarnes:
> If the spec calls out "identity", which it does, an implementation that does not support "identity" is not "spec-compliant".

I feel like we're going around in circles a bit. It is correct to say that the current spec includes the identity transformation. However, the spec does not allow the use of that transformation. Therefore, any implementation which allows the use of the identity transformation is not spec-compliant. So while it's possible for implementations to have an identity transformation defined, if they can parse it from the metadata, or if they can write it into the metadata, they are not compliant with the spec. Therefore, there are zero functional changes associated with removal of the identity transform, it simply allows the removal of code which *must* be unreachable according to the current spec. Meanwhile, allowing the use of the identity transform requires changes which hook an implementation's representation of that transform into the parsing and transforming workflow.

Does this make sense? I feel like I may not be getting my point across very clearly.

I see your point about explicit transformations allowing logic like "In my particular workflow, the 3rd transformation is always the one generated by process X, and therefore if X generates an identity transform, I must be able to represent it in the transformation array". That's totally reasonable! However, I'm not sure it's necessarily compatible with the current spec, which limits the number and order of transformations (one scale, and up to one translation, which must occur after the scale). Unless your registration workflow is specifically designed to fit to the OME-NGFF spec (which it might be) I can see it easily breaking those rules.

thewtex:
@clbarnes I understand that you are trying to assert that the spec does not allow identity transformations, and an implementation with an identity transform is not spec-compliant. This assertion was made repeatedly. I understand your point.

However, it does not make sense that the spec states that transforms supported include the `identity` transform, and an `identity` transform is not allowed.

Clearly, clarifications are required. Perhaps this PR should be integrated into the v0.5 update?

ziw-liu:
TIL we've written 10s of PB of spec-conformant v0.4 data that will fail at the validator due to this inconsistency. Is it possible to limit the scope and move forward with a fix and get a v0.4.2 released? This should be strictly backwards compatible.

will-moore (approved):
We probably should avoid editing `latest/index.bs` at this point, to reduce the chance of conflicts, and it's not needed to fix the validation errors from the schemas.

Otherwise I think it makes sense to merge this. But deferring to @joshmoore on the consequences (this shouldn't need any version numbers changing etc)?

d-v-b:
By adding a new type of valid transformation type to the elements of `multiscales`, this PR will require changes to implementations of the spec, so I wouldn't merge it would getting some feedback on that point.

d-v-b:
the spec currently says 
> The transformations are defined according to [[#trafo-md]]. The transformation MUST only be of type `translation` or `scale`.

This PR changes this statement to say
> The transformations are defined according to [[#trafo-md]]. The transformation MUST only be of types defined in [[#trafo-md]].

This adds a new transform type (`identity`) where previously only `scale` and `translation` were allowed. Any complete implementation of the spec would have to adjust to this change.

will-moore:
Ah, yes I see, thanks.

Unfortunately the existing spec isn't entirely consistent because the `[#trafo-md]` section linked to in that text https://ngff.openmicroscopy.org/0.4/index.html#trafo-md says "The value of "type" MUST be one of the elements of the type column in the table below" - The 'types' in the table are `identity`, `translation` and `scale`.

So yes, agreed it's possible that some implementations may only support `scale` and `translation`.

d-v-b:
unfortunately the [same confusing language is in 0.5](https://github.com/ome/ngff/blob/bd6de5a6a86909f51dc9a793a6517dc33e78b91f/0.5/index.bs#L314).

joshmoore:
Very sorry for having missed this for 0.5. I'd propose we fold it into 0.6. I'll work towards an initial dev release of 0.6 like we did for the challenge (but using the new [repo setup](https://github.com/ome/ngff/pull/282) so we can hopefully start validating, etc.

Issue #153, opened 2022-10-21
Scale array MUST have the same dtype


constantinpape:
The current v0.4 spec does not explicitly require that the scale arrays of one image to have the same dtype.
I think this should be fixed ASAP, since having different datatypes will make the data invalid for many tools. (In particular we know that this is the case for MoBIE, and I assume it's the case for most tools written in a strongly typed language).

We have discussed this before, and we're running into this issue with real data now. I am happy to propose the changes for this, but I didn't have time to keep up with all the latest spec developments; should the PR go to `latest` or is there something to take care of w.r.t. other current PRs?

will-moore:
I think this is a duplicate of https://github.com/ome/ngff/issues/2?
This is a small change and doesn't affect or conflict with other changes so I think it can be included in the v0.5 spec.

Although this isn't something we can validate with a JSON schema, I should be able to get ome-ngff-validator to check this and warn as Invalid for v0.5 spec. Noted at https://github.com/ome/ome-ngff-validator/issues/6

constantinpape:
Closed in favor of #2. (And PR is open #154)

PR #154, opened 2022-10-21
Require same datatype for all arrays in multiscales:datasets

### LINKED ISSUE #2: Constrain dtypes for multiscale arrays
see: https://github.com/tischi/i2k-2020-s3-zarr-workshop/issues/9#issuecomment-732369669

The current spec allows for each resolution level to be a different dtype. This should be restricted with a clause that they "MUST be the same".

constantinpape:
Fixes #2

I think it is important to include this change rather soon. Right  now data with different dtypes across the dimensions is technically valid according to the spec, but will cause issues for several implementations.
(It does not work in MoBIE; I am pretty sure the same is true for all tools based on BigDataViewer, and I suspect any tool implemented in a strongly typed language will have issues with this).

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -153,7 +153,8 @@ For this example we assume an image with 5 dimensions and axes called `t,c,z,y,x
     │   ...                   # which is a folder containing chunk files which compose the array.
     ├── n                     # The name of the array is arbitrary with the ordering defined by
     │   │                     # by the "multiscales" metadata, but is often a sequence starting at 0.
-    │   │
+    │   │                     # All arrays must have the same datatype and the same number of dimensions.
+    |   | 
     │   ├── .zarray           # All image arrays must be up to 5-dimensional
     │   │                     # with the axis of type time before type channel, before spatial axes.
     │   │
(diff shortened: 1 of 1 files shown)


latest/index.bs
```diff
@@ -365,10 +366,11 @@ The order of the entries MUST correspond to the order of dimensions of the zarr
 If there are three spatial axes where two correspond to the image plane ("yx") and images are stacked along the other (anisotropic) axis ("z"), the spatial axes SHOULD be ordered as "zyx".
 
 Each "multiscales" dictionary MUST contain the field "datasets", which is a list of dictionaries describing the arrays storing the individual resolution levels.
-Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
+Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative.
 to the current zarr group. The "path"s MUST be ordered from largest (i.e. highest resolution) to smallest.
 
-Each "datasets" dictionary MUST have the same number of dimensions and MUST NOT have more than 5 dimensions. The number of dimensions and order MUST correspond to number and order of "axes".
+All arrays referenced in "datasets" MUST have the same number of dimensions and MUST have the same datatype. The number of dimensions MUST NOT be more than 5 dimensions.
```
will-moore, line 372: "The number of dimensions MUST NOT be more than 5 dimensions" - I think this rule is due to be removed in v0.5? cc@bogovicj I don't know if that was going to happen in https://github.com/ome/ngff/pull/138 but I don't see that it has yet. So it may need a rebase etc if this is merged first to avoid conflicts

latest/index.bs
```diff
@@ -365,10 +366,11 @@ The order of the entries MUST correspond to the order of dimensions of the zarr
 If there are three spatial axes where two correspond to the image plane ("yx") and images are stacked along the other (anisotropic) axis ("z"), the spatial axes SHOULD be ordered as "zyx".
 
 Each "multiscales" dictionary MUST contain the field "datasets", which is a list of dictionaries describing the arrays storing the individual resolution levels.
-Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative
+Each dictionary in "datasets" MUST contain the field "path", whose value contains the path to the array for this resolution relative.
```
will-moore, line 369: Full stop added accidentally? The sentence continues on the next line

latest/index.bs
```diff
@@ -153,7 +153,8 @@ For this example we assume an image with 5 dimensions and axes called `t,c,z,y,x
     │   ...                   # which is a folder containing chunk files which compose the array.
     ├── n                     # The name of the array is arbitrary with the ordering defined by
     │   │                     # by the "multiscales" metadata, but is often a sequence starting at 0.
-    │   │
+    │   │                     # All arrays must have the same datatype and the same number of dimensions.
+    |   |
```
will-moore, line 157: Minor point: Looks like 2 lines have changed because whitespace was added to the "blank" line. `    |   | ` Just adding 1 line is preferable.

will-moore (commented):
Looks good. A couple of minor comments

clbarnes:
I suppose one place where this might not be wanted would be in label datasaets where the bottom level is just a voxel array and higher levels are variable length as labels get collapsed together? Storing that bottom level as variable length could potentially be costly.

lubianat:
Just got to this PR and issues 
- #2
- #153 
- #221

To avoid bloating the queue for 0.6, I've added it to the 1.0 milestone, but if it is included sooner, that seems good too.

jo-mueller:
@constantinpape I have moved this requirement over to https://github.com/ome/ngff-spec and will close this. Thanks for the PR!

Issue #135, opened 2022-08-16
Allow `.` as the dimension separator


jstriebel:
Since v0.2, `/` is the only allowed dimension separator for OME NGFF chunk paths. To be compatible with Zarr arrays in general, it would be great to also allow `.` as the dimension separator, depending on the Zarr metadata if available.

This allows to add OME metadata to existing Zarr arrays, that already use the `.` dimension separator.

joshmoore:
Thanks for capturing this, @jstriebel. I can't remember if I mentioned to you or to Norman but for everyone: the hard requirement on a single separator was required while the `dimension_separator` addition was being add to the Zarr v2 spec. At this point, I think it's fine to drop. If either of you want to drive that, please feel free. The major caveat will be that the implementations will need to be updated as well.

(There have also been requests to support Zip-based NGFFs which will touch the same code paths.)

normanrz:
We are happy to drive this.
Fwiw, I wouldn't interpret the current spec to have a strict `/` requirement, because it is not explicitly mentioned. Other parts of the Zarr spec are also not re-defined in the OME-NGFF spec, so, I would assume all features of zarr v2 are valid. Most implementations that support OME-NGFF already support `.`, because of their underlying zarr implementations.

joshmoore:
Then we're looking at different implementations which is a good realization itself :smile:

```
 │   └─ t                  # Chunks are stored with the nested directory layout.
```

normanrz:
I see. To me the directory listing seemed more like an example section than a normative section.

dstansby:
As @d-v-b says at https://forum.image.sc/t/webknossos-uses-as-a-dimension-separator/110373/9?u=dstansby, it seems like the spec doesn't disallow "." as a separator? Certainly my reading of version 0.5 is it doesn't. So I think this can be closed?

normanrz:
It is definitely fixed in 0.5.

joshmoore:
The need to hard code it before 0.5 was that for a period Zarr didn't have this metadata recorded anywhere and so assumptions had to be made. The move to Zarr v3, as @normanrz says, definitely corrects it. :+1:

Issue #100, opened 2022-02-16
Misc spec clarifications


constantinpape:
There are a few clarifications that could improve the spec (without changing any content). This could be done for a version 0.4.1 or just rolled into 0.5.

- [ ] Specify json encoding https://github.com/ome/ngff/issues/81
- [ ] Constrain dtypes & shape for multiscale arrays: https://github.com/ome/ngff/issues/2
- [ ] Clarify purpose of multiple multiscales https://github.com/ome/ngff/issues/17
- [ ] Clarify json terminology https://github.com/ome/ngff/issues/66
- [ ] Text vs.  markdown quotes: https://github.com/ome/ngff/issues/38
- [ ] Single version information at top level, see https://github.com/ome/ngff/issues/83. (That would be a change of the metadata, but only superficial)

joshmoore:
Would you consider https://github.com/ome/ngff/issues/22 ("Update label properties to include example: prefix") under this banner as well? Or does that count as a spec _change_ since we would basically be saying "this isn't core"?

cc: @sbesson

constantinpape:
> Would you consider #22 ("Update label properties to include example: prefix") under this banner as well? Or does that count as a spec _change_ since we would basically be saying "this isn't core"?
> 
> cc: @sbesson

I don't have enough background on this and don't understand these parts of the spec (yet). So up to your judgement ;).

constantinpape:
We should also include #105 here.

PR #161, opened 2022-11-10
Allow '.' dimension separator

normanrz:
I am proposing to allow the `.` dimension separator alongside `/`. In `.zarray` there is already the relevant metadata available, which OME-NGFF can defer to.

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -157,7 +157,7 @@ For this example we assume an image with 5 dimensions and axes called `t,c,z,y,x
     │   ├── .zarray           # All image arrays must be up to 5-dimensional
     │   │                     # with the axis of type time before type channel, before spatial axes.
     │   │
-    │   └─ t                  # Chunks are stored with the nested directory layout.
+    │   └─ t                  # Chunks are stored with the flat or nested directory layout. The `dimension_separator` field of the underlying Zarr array must be set accordingly.
     │      └─ c               # All but the last chunk element are stored as directories.
     │         └─ z            # The terminal chunk is a file. Together the directory and file names
     │            └─ y         # provide the "chunk coordinate" (t, c, z, y, x), where the maximum coordinate


joshmoore:
:+1::

 - closes https://github.com/ome/ngff/issues/135
 - see https://github.com/ome/ngff/issues/100

I think there are two remainingissues:
 * The primary one is what version of NGFF will this match. (Proposals needed.)
 * The other is that there are almost certainly datasets in the wild which use the non-default separator and which do NOT set the `dimension_separator` property. We might should warn as to this situation and perhaps (if possibly) validate it. cc: @will-moore

normanrz:
> * The primary one is what version of NGFF will this match. (Proposals needed.)

This could be a 0.4.2?


> * The other is that there are almost certainly datasets in the wild which use the non-default separator and which do NOT set the `dimension_separator` property. We might should warn as to this situation and perhaps (if possibly) validate it.

We could specify `/` as default in the OME-NGFF spec. Is there a default on the Zarr side?

normanrz:
> > Is there a default on the Zarr side?
> 
> If there is no metadata in .zarray, then `.` is used. The problem stems from the fact that before `dimension_separator` was introduced, `NestedDirectoryStore` and friends wrote file sets with `/` without setting any metadata (which should be considered invalid/broken but requires a heuristic to detect.)

Do you think there is any v0.4 data out there that matches this description? That could be an argument for making this a v0.5 change.

will-moore:
> The other is that there are almost certainly datasets in the wild which use the non-default separator and which do NOT set the dimension_separator property

If there are such files, these would already be considered invalid zarr files?
Or not valid v0.4 files anyway?

So, if we were to state "v0.4+ files with missing `dimension_separator` are invalid" that wouldn't be a breaking change?

joshmoore:
> If there are such files, these would already be considered invalid zarr files?
> Or not valid v0.4 files anyway?

Not valid NGFF before this PR, exactly. They _are_ valid V2 files, but essentially the addition of `dimension_separator` was like a subtle `v2.1` to fix the fact that it wasn't possible to detect them.

will-moore:
We already check that dimension separator is '/' for v0.2 -> v0.4: https://github.com/ome/ome-ngff-validator/pull/16/files#diff-b06a22b09da1dc9eba4472e0983b0c06000bb32a69693e8d8b86275a3034dd85R58

normanrz:
> We already check that dimension separator is '/' for v0.2 -> v0.4: https://github.com/ome/ome-ngff-validator/pull/16/files#diff-b06a22b09da1dc9eba4472e0983b0c06000bb32a69693e8d8b86275a3034dd85R58

If I read the code correctly, it will yield an error if the `dimension_separator` field is not present.

will-moore:
@normanrz Yes. I didn't test it but I think it should do that (which is what we want, right)?
Also, it only checks for v0.2 - v0.4, so if we allow `.` in v0.5 it won't need updating.

normanrz:
I don't have a strong opinion about whether this should go into v0.4.2 or v0.5. However, i'm not sure if this small change is worth to trigger a v0.5 and/or to wait for bigger changes to reach consensus.

will-moore:
I'm not really sure what a `v0.4.2` means for schema validation etc.
Would it mean that previously generated 'v0.4' that use `.` dimension separator are now valid?
E.g. this is currently reported as invalid, but would become valid (except for the axes order):
https://ome.github.io/ome-ngff-validator/?source=https://data-humerus.webknossos.org/data/zarr/scalable_minds/skin/color

Screenshot 2022-11-11 at 13 07 00

It would also need the fix at https://github.com/ome/ome-zarr-py/pull/235

normanrz:
> I'm not really sure what a `v0.4.2` means for schema validation etc.
> Would it mean that previously generated 'v0.4' that use `.` dimension separator are now valid?

Yes, they would become valid.

> It would also need the fix at [ome/ome-zarr-py#235](https://github.com/ome/ome-zarr-py/pull/235)

Yep. Not sure what other tools would need adapting. My guess is, most already support `.` because of the underlying zarr implementation.

joshmoore:
@will-moore: after our recent discussion, thoughts on getting this rolled into 0.5?

cc: @kevinyamauchi

joshmoore:
@normanrz, are you still interested in pursuing this or would you prefer to focus on the move to V3?

normanrz:
> @normanrz, are you still interested in pursuing this or would you prefer to focus on the move to V3?

No, closing this in favor of Zarr v3 support in #227.

Issue #108, opened 2022-03-15
"labels" metadata section is confusing (and wrong?)


constantinpape:
I am pretty confused by the ["labels" metadata section](https://ngff.openmicroscopy.org/0.4/#labels-md):
It looks like this specifies the path to the *first scale array*. However, it would make much more sense to me if this specified the *group with multiscales metadata* for the label image.
I.e. instead of `{"labels": ["orphaned/0"]}` I would expect `{"labels": ["orphaned"]}`. Is this maybe just wrong?

On a sidenote: why do you call it `orphaned`? I think for labels something like `mask` or `my-fancy-mask` would make more sense as an example name.

sbesson:
I agree that the `labels` dictionary should point at the `multiscales` group rather than the first array of the group. Looking at the  OME-NGFF samples generated by the OME team (see [this blog post](https://www.openmicroscopy.org/2020/11/04/zarr-data.html) and [this blog posts[(https://www.openmicroscopy.org/2021/12/16/ome-ngff.html)) that contain labels, it looks like this is also how this was implemented.

```
(base) sbesson@Sebastiens-MacBook-Pro ~ % curl https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.3/idr0079A/9836998.zarr/labels/.zattrs
{
    "labels": [
        "0"
    ]
}%                                                                              (base) sbesson@Sebastiens-MacBook-Pro ~ % curl https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.3/idr0052A/5514375.zarr/labels/.zattrs
{
    "labels": [
        "Cell",
        "Chromosomes"
    ]
}%                                                                              (base) sbesson@Sebastiens-MacBook-Pro ~ % curl https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr/labels/.zattrs
{
    "labels": [
        "masks"
    ]
}%  
```

I suspect this is an oversight and if so, it should be rectified at earliest. 👍 from my side for a PR updating the specification.
On the naming, I agree that `orphaned` leads to confusion. Maybe this should be unified with the label name used in the [example layout](https://ngff.openmicroscopy.org/0.4/#image-layout) for consistency?

constantinpape:
[...]

PR #165, opened 2023-02-09
Labels section rewrite

virginiascarlett:
Hello all,

@bogovicj and I have co-written a new “labels” section to the OME-NGFF spec that combines the `"labels" metadata` and `"image-label" metadata` sections and hopefully improves their readability. We did our best to keep the actual spec the same, and just improve the language explaining it. 

The new “labels” JSON example is simpler than the one discussed in [issue #108](https://github.com/ome/ngff/issues/108). Hopefully the ambiguity of the previous example (it was unclear whether it was pointing to an image or a scale level) is cleared up. We wrote: “All images in the "labels" group SHOULD be listed here.” This was our interpretation of the existing statement that “Unlisted groups MAY be labels.” (BTW, any reason this is a MAY/SHOULD and not a MUST?)

Welcoming your feedback. 

P.S. – For those who don’t know, I recently joined HHMI Janelia and I am new to the bioimaging field. This is my first PR ever!

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -676,6 +676,11 @@ Version History {#history}
   </thead>
   <tr>
     <td>0.4.1</td>
+    <td>2023-02-09</td>
+    <td>expand on "labels" description</td>
+  </tr>
+  <tr>
+    <td>0.4.1</td>
     <td>2022-09-26</td>
     <td>transitional metadata for image collections ("bioformats2raw.layout")</td>
   </tr>
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -3,9 +3,9 @@ Title: Next-generation file formats (NGFF)
 Shortname: ome-ngff
 Level: 1
 Status: LS-COMMIT
-Status: w3c/CG-FINAL
+Status: w3c/ED
 Group: ome
-URL: https://ngff.openmicroscopy.org/0.4/
+URL: https://ngff.openmicroscopy.org/latest/
 Repository: https://github.com/ome/ngff
 Issue Tracking: Forums https://forum.image.sc/tag/ome-ngff
 Logo: http://www.openmicroscopy.org/img/logos/ome-logomark.svg
@@ -16,11 +16,13 @@ Markup Shorthands: markdown yes
 Editor: Josh Moore, University of Dundee (UoD) https://www.dundee.ac.uk, https://orcid.org/0000-0003-4028-811X
 Editor: Sébastien Besson, University of Dundee (UoD) https://www.dundee.ac.uk, https://orcid.org/0000-0001-8783-1429
 Editor: Constantin Pape, European Molecular Biology Laboratory (EMBL) https://www.embl.org/sites/heidelberg/, https://orcid.org/0000-0001-6562-7187
+Text Macro: NGFFVERSION 0.5-dev
 Abstract: This document contains next-generation file format (NGFF)
 Abstract: specifications for storing bioimaging data in the cloud.
 Abstract: All specifications are submitted to the https://image.sc community for review.
-Status Text: This is the 0.4 release of this specification. Migration scripts
-Status Text: will be provided between numbered versions. Data written with the latest version
+Status Text: The current released version of this specification is
+Status Text: <a href="../0.4/index.html">0.4</a>. Migration scripts
+Status Text: will be provided between numbered versions. Data written with these latest changes
 Status Text: (an "editor's draft") will not necessarily be supported.
 </pre>
 
@@ -328,8 +330,9 @@ Conforming readers:
 - MAY choose to show all images within the collection or offer the user a choice of images, as with <dfn export="true"><abbr title="High-content screening">HCS</abbr></dfn> plates;
 - MAY ignore other groups or arrays under the root of the hierarchy.
 
+
 "coordinateTransformations" metadata {#trafo-md}
--------------------------------------
+------------------------------------------------
 
 "coordinateTransformations" describe a series of transformations that map between two coordinate spaces (defined by "axes").
 For example, to map a discrete data space of an array to the corresponding physical space.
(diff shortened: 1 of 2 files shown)


joshmoore:
Hi @virginiascarlett. Thanks for this! (To @bogovicj as well). It looks like the edits were done on a copy of `0.4/index.bs` file but then committed as `latest/index.bs` which gives us a substantial number of, I think unintended, changes: https://github.com/ome/ngff/pull/165/files Any chance of getting a more minimal set?

virginiascarlett:
Oops! Yes, I had somehow copy-pasted text from the 0.4/ index into the latest/ index, causing quite a mess. I think it should be fixed now? Thanks for helping me learn.

joshmoore:
It does looked fixed, thanks! And no worries at all.

latest/index.bs
```diff
@@ -444,59 +444,65 @@ for more information.
...
   "labels": [
-    "orphaned/0"
+    "cell_space_segmentation/"
   ]
 }
 ```
 
-Unlisted groups MAY be labels.
+The `.zattrs` file within an image within a "labels" group MUST contain a `multiscales` JSON object. Within the `multiscales` object, the JSON array 
+associated with the `datasets` key MUST have the same number of entries (scale levels) as the original unlabeled image.
```
will-moore, line 470: This `MUST` seems a bit more strict than necessary? E.g. I may have scale levels down to ~100 pixels for thumbnails of my original images, but there may be no use-case for generating any downsampling of labels or vice versa.
E.g. https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0062A/6001240.zarr has different numbers of scale levels for original image (3) and labels (4).

will-moore (commented):
Looks good, thanks.
Just one comment added...

latest/index.bs
virginiascarlett: This was our interpretation of the line in the current spec that reads, "`image-label` groups MUST also contain multiscales metadata and the two "datasets" series MUST have the same number of entries." 

Your suggestion sounds fine to me. Maybe chage the new line to, "Within the `multiscales` object, the JSON array associated with the `datasets` key need not have the same number of entries (scale levels) as the original unlabeled image."?

will-moore: My apologies - I didn't remember that restriction from the current spec (and didn't check it before I commented).
I still feel that we shouldn't enforce the same number of `datasets`, and your suggestion sounds good, but I'd want to be sure that others are OK with the removal of that restriction (in case there's a reason that it was added that I'm not aware of)? cc @joshmoore?

sbesson: Seconding @virginiascarlett's comment, this is definitely a section where the current specification is explicit about the requirement. Personally, I am not opposed to loosening this aspect of the specification but I think the details matter. While the example above brings an interesting perspective about the requirements for lower resolution levels, would we not at least recommend that the dimensions of the largest resolution should match the original image?

As this particular point differs the rest of this proposal which aims to clarifying the existing specification without changing it, I would suggest we migrate this conversation in a follow-up PR rather than burying a spec change discussion in the middle of a larger set of changes.

sbesson (commented):
Overall the new section reads well to me. Apart from the spec change which I would propose to move elsewhere, I put inline comments for the two main points which I believe deserve clarification/consensus:

- the rules for individual sections within this specification
- the requirement level for the `image-label` key

virginiascarlett:
Thanks for your feedback! Here's how I propose we move forward.

First, I will make a new commit with the following changes:
- Correct version for the color_properties file (Should be 0.5-dev. My silly mistake.).
- Clearer language emphasizing three of Sebastien's points:

> - the label image SHOULD be registered within the labels metadata
> - the label image MUST implement the multiscales specification
> - the pixel type MUST be integer data types

Then, I will create a NEW PR with the following changes:
- The label image MAY have the same number of scale levels as the original image. Therefore, it is acceptable for the label image to have fewer scale levels than the original image.
- The label image MUST be stored within the labels group. 
- The label image .zattrs file MUST contain the image-label key.

In this way, I believe we can separate superficial changes to the language from substantive changes to the specification itself.

latest/index.bs
```diff
@@ -444,59 +444,65 @@ for more information.
 "labels" metadata {#labels-md}
```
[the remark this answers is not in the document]
virginiascarlett: Perhaps this warrants a Github Discussion?

sbesson (approved):
Barring the discussion on the section merging/restructuring, the proposed changes make sense to me. Thanks for the contribution

latest/index.bs
```diff
@@ -444,59 +444,65 @@ for more information.
...
+but these MUST NOT contain metadata. Names of the images in the "labels" group are arbitrary.
+
+The `.zattrs` file associated with the "labels" group MUST contain a JSON object with the key `labels`, whose value is a JSON array of paths to the 
+labeled multiscale image(s). All label images SHOULD be listed within this metadata file. For example:
 
 ```json
 {
   "labels": [
-    "orphaned/0"
+    "cell_space_segmentation/"
```
will-moore, line 464: This probably shouldn't have a trailing `/`

will-moore (approved):
Just the `/` fix I think

joshmoore (approved):
Thanks, everyone. Merging.

A few last minute thoughts:
 - @virginiascarlett: we haven't had huge success with the GH discussion functionality yet. (Neither here nor on the Zarr repo.) You are more than welcome to give it a try, but if it doesn't get the buy-in you want, I'd suggest an issue or image.sc.
 - For the follow-on issue, do note there are several other label issues out there. If we start to get into larger changes (e.g. that would change the spec version) we might want to take a step back and consider other options. Another possibility would be to update the schema to capture the current state and then let 0.5 progress as planned.

sbesson:
Incidentally coming back to this, I realise this change has been added to the `Version History` under `0.4.1` but is only published on https://ngff.openmicroscopy.org/latest/index.html#labels-md.

Should it also be ported to https://ngff.openmicroscopy.org/0.4/index.html#labels-md ?

PR #170, opened 2023-03-07
Resolving misc issues and ambiguities in labels section

virginiascarlett:
Following up on my last PR. Previously, we added a lot more explanation to the 'labels' section of the spec. Now I am not just editing the language, but making some substantive, though minor, changes. Here's a quick summary of what I changed:
- Added for clarity: "Each label image MUST be stored within a "labels" group."
- One MUST --> MAY: "the datasets key MUST have the same number of entries (scale levels) as the original" --> "the datasets key MAY have the same number of elements (scale levels) as the original"
- One SHOULD --> MUST + clearer language: "In addition to the multiscales key, the JSON object in this image-level .zattrs file SHOULD contain another key, image-label, whose value is also a JSON object." --> "The label image .zattrs file MUST also contain the image-label key, whose value is a JSON object."
- Changed for clarity: "This array contains one JSON object for each unique custom label" --> "This array SHOULD contain one JSON object for each unique custom label."

Thanks!

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -450,6 +450,7 @@ This new image is composed of integer values corresponding to certain labels wit
 if the corresponding pixel in the original image represents cellular space or intercellular space, respectively. 
 Such an image is referred to in this specification as a 'label image'. 
 
+Each label image MUST be stored within a "labels" group. 
 The "labels" group is nested within an image group, at the same level of the Zarr hierarchy as the resolution levels for the original image. 
 The "labels" group is not itself an image; it contains images. The pixels of the label images MUST be integer data types, i.e. one of 
 [`uint8`, `int8`, `uint16`, `int16`, `uint32`, `int32`, `uint64`, `int64`]. Intermediate groups between "labels" and the images within it are allowed, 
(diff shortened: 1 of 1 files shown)


latest/index.bs
```diff
@@ -467,15 +468,16 @@ labeled multiscale image(s). All label images SHOULD be listed within this metad
...
 
 The `.zattrs` file for the label image MUST implement the multiscales specification. Within the `multiscales` object, the JSON array 
-associated with the `datasets` key MUST have the same number of entries (scale levels) as the original unlabeled image. 
+associated with the `datasets` key MAY have the same number of elements (scale levels) as the original unlabeled image. 
+Therefore it is acceptable for a label image to have fewer scale levels than the original image.
 
-In addition to the `multiscales` key, the JSON object in this image-level `.zattrs` file SHOULD contain another key, `image-label`, 
-whose value is also a JSON object. The `image-label` object stores information about the display colors, source image, and optionally, 
+The label image .zattrs file MUST also contain the `image-label` key, whose value is a JSON object. 
+The `image-label` object stores information about the display colors, source image, and optionally,
```
will-moore, line 475: Since there are no mandatory contents of the `image-label` dictionary, it would be valid to have `image-label: {}`.
It seems that making the spec more strict (a breaking change) to require the `image-label` dict is probably not worth it in that case?

will-moore (commented):
Just one comment. All the other changes look great!

latest/index.bs
sbesson: Thanks for the ping @virginiascarlett. Trying to justify, the rationale behind my original statement, I see a label image as a specialized type of multiscales image. At the moment, the specification enforces that such data must be stored within a well-defined `labels` hierarchy but moving forward, I could certainly imagine a relaxation of this constraint.

A typical use case that comes immediately to mind is the one where segmentation / classification is performed against a read-only Zarr dataset e.g. public data and the output of this process needs to be stored as a new dataset. At the moment, the structure which is the most compliant with the spirit of the specification is create an artificial `labels/<label_name>/` hierarchy under the root even if there is no multiscales image. Assuming we relaxed this constraint to allow label images to be stored at the root of the Zarr dataset, I would argue the `image-label` metadata would become a critical element to identify what we are dealing with.

That being said, Will's point makes sense and if this specification change simply leads to most implementations adding `image-label: {}` in the metadata, I don't find this is particularly helpful.
Happy to separate this discussion from your proposal, revert the requirement level back to the SHOULD level and come back to this discussion if we decide to specify standalone label images.

will-moore (approved):
Looks good, thanks 👍

jni:
> Added for clarity: "Each label image MUST be stored within a "labels" group."

I would like to point to [this imagesc discussion](https://forum.image.sc/t/save-a-single-labels-dataset-into-an-ome-zarr/93505) where people are quite unanimous that this is a bad idea — can we remove this line?

latest/index.bs
lubianat: Getting here now, I think @sbesson's user story here is currently a part of RFC-8 
https://ngff.openmicroscopy.org/rfc/8/versions/v1-2026-08/index.html#grouping-together-remote-images

@normanrz would you know if there is a github issue (or Zulip thread) collecting user stories for RFC-8?

PR #171, opened 2023-03-08
Add ITKIOOMEZarrNGFF to implementations

dzenanz:


will-moore:
Hi - thanks for the suggestion and apologies for the lack of response.

Please could you add this to the `libraries` list at https://github.com/ome/ngff/blob/main/tools/index.md#libraries,
which should now be displayed at https://ngff.openmicroscopy.org/tools

dzenanz:
Done. I also rebased on top of `main`.

joshmoore:
Thanks for this, @dzenanz! We'll likely drop the entire list in `latest/index.bs` and just link to the new /tools location, but :+1: for having it in all the appropriate places for the moment.

PR #175, opened 2023-03-14
Add CITATION.cff

joshmoore:
Created using https://citation-file-format.github.io/cff-initializer-javascript/#/

Open questions/issues:
 * [ ] list all spec authors here as well as in the specs?
 * [ ] list all versions here as well as in the specs?
 * [ ] use the paper as a preferred-citation?
 * [ ] apparently the tags in this repo need updating
 * [ ] as does the license file

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,110 @@
+# This CITATION.cff file was generated with cffinit.
+# Visit https://bit.ly/cffinit to generate yours today!
+
+cff-version: 1.2.0
+title: Nex-generation file format (NGFF)
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
+  - given-names: Josh
+    family-names: Moore
+    email: josh@openmicroscopy.org
+    affiliation: German BioImaging e.V.
+    orcid: 'https://orcid.org/0000-0003-4028-811X'
+  - given-names: Sébastien
+    family-names: Besson
+    email: sbesson@glencoesoftware.com
+    affiliation: 'Glencoe Software, Inc.'
+    orcid: 'https://orcid.org/0000-0001-8783-1429'
+  - given-names: Constantin
+    family-names: Pape
+    email: constantin.pape@informatik.uni-heidelberg.de
+    affiliation: 'Georg-August Universität Göttingen, Germany'
+    orcid: 'https://orcid.org/0000-0001-6562-7187'
+identifiers:
+  - type: doi
+    value: 10.1038/s41592-021-01326-w
+    description: Original Nature Methods article (2021)
+  - type: doi
+    value: 10.1101/2023.02.17.528834v2
+    description: OME-Zarr preprint (2023)
+repository-code: 'https://github.com/ome/ngff'
+url: 'https://ngff.openmicroscopy.org'
+abstract: >-
+  A growing community is constructing a next-generation file
+  format (NGFF) for bioimaging to overcome problems of
+  scalability and heterogeneity. Organized by the Open
+  Microscopy Environment (OME), individuals and institutes
+  across diverse modalities facing these problems have
+  designed a format specification process (OME-NGFF) to
+  address these needs. Those community members have built
+  the format itself – OME-Zarr – along with tools and data
+  resources available today to increase FAIR access and
+  remove barriers in the scientific process. The current
+  momentum offers an opportunity to unify a key component of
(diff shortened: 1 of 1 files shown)


joshmoore:
FYI from https://citation-file-format.github.io/ 
image

CITATION.cff
```diff
@@ -0,0 +1,110 @@
+# This CITATION.cff file was generated with cffinit.
+# Visit https://bit.ly/cffinit to generate yours today!
+
+cff-version: 1.2.0
+title: Nex-generation file format (NGFF)
```
will-moore, line 5: typo `Nex-`
Should the title include `OME`?

joshmoore: Typo fix pushed. Re: "OME" good question. It's hard to do the expansion, then. Maybe let's see how it looks when we start using it.

CITATION.cff
```diff
@@ -0,0 +1,110 @@
...
+message: >-
+  If you use this software, please cite it using the
```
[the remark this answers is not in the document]
joshmoore: > This repo isn't really software.

Updated to "Specification" (but "type" must remain "Software" since the only other option is "Dataset").

> Should you cite this repo if you use ome-zarr-py?

I wouldn't say it's a must.

joshmoore:
Re-opened to trigger RTD build.

jo-mueller:
Coming back to this after reading issue #315 . Maybe a small statement in the [Contributions](https://github.com/ome/ngff/blob/main/contributing/index.md) guide would be nice for PR authors to add themselves there if they haven't yet?

jo-mueller:
Ok, suggestion for attribution:

- **Authorship**: Since we'll break out the spec document into the ngff-spec repo authorship could simply go into the project metadata of the built docs.
- **Citation**: We currently point to the [Zenodo view on the spec text](https://zenodo.org/records/4282107) as preferred citation. I think this works in general. At least it reproduces the current recommendation as per ngff.openmicroscopy.org
- **License**: The MyST metadata standard allows a [few common licensing options ](https://mystmd.org/guide/frontmatter#licenses) to be embedded into the project's metadata. Maybe that would help? I think right now, the license contains references to a few different components of this repo. Maybe this could be simpler for only the spec text?
- **Tags**: A tag in the ngff-repo would then be authoritative and do the following things:
  - Trigger a Python script that builds a `citation.cff` from the metadata of the spec documents
  - Trigger the upload to Zenodo with the correct author information

### Open question

- Should all authors be displayed in the document's frontmatter?
- How should they be listed? Alphabetically based on the last name?

### Next steps

@joshmoore was so kind to print me a list of all contributors to the spec. I would ping all of them here and ask them to add their name/affiliation/orcid/github to [this branch](https://github.com/jo-mueller/ngff-spec/tree/include-all-authors) over at ngff-spec under `./ngff_spec/myst.yml`

@dstansby @joshmoore @will-moore LMKWYT :)

will-moore:
Where do we have citation instructions? Ah - I see https://ngff-spec.readthedocs.io/en/latest/citing.html but nothing from https://ngff.openmicroscopy.org/.
I guess the zenodo link will need updating as part of what's triggered by a new Tag?

Don't know about "frontmatter" or where/how it's displayed?
Listing Alphabetically based on the last name seems best, apart from maybe Josh (as Editor)

jo-mueller:
>but nothing from https://ngff.openmicroscopy.org/.

Yes. I could see this page becoming more of a hub for all sorts of resources beyond just the specification. I think it makes sense to move citation information along with the spec if someone wanted to cite the spec itself.

> I guess the zenodo link will need updating as part of what's triggered by a new Tag?

Or maybe it could be removed entirely? For once, readthedocs shows you the version of the document you are looking at anyway. Secondly, the link to Zenodo *could* point to the latest version and simply say: "All past versions of the spec are available on Zenodo (link)" 

It's probably possible to find out the DOI on Zenodo on build and insert it in the markdown documents but maybe that's overkilling it ^^"

> Don't know about "frontmatter" or where/how it's displayed?

Essentially, every markdown document in the page gets a header-like part (the frontmatter), the content of which can be defined from the project metadata (the `myst.yml`) or overwritten in every document by inserting this at the top:

```markdown
---
title: Whatever
foo: bar
---
```

If nothing is displayed, then the result would look something [like this example](https://heftie-textbook.readthedocs.io).

> Listing Alphabetically based on the last name seems best, apart from maybe Josh (as Editor)

👍

PR #186, opened 2023-03-22
Version required

### LINKED ISSUE #185: Change `version` field(s) from SHOULD to MUST
Being able to rely on the `version` (either at top-level https://github.com/ome/ngff/issues/83) or within the individual metadata fields would make implementation of parsers / readers easier - avoiding the need to guess which schema version a particular OME-NGFF or metadata field conforms to.

This seems particularly important for allowing potential semantic changes between spec versions that might not be able to be inferred from data types (e.g., based on whether `axes` elements are strings vs. objects)

will-moore:
Fixes #185.

### DIFF
--- a/latest/index.bs
+++ b/latest/index.bs
@@ -546,7 +546,7 @@ contain only alphanumeric characters, MUST be case-sensitive, and MUST NOT be a 
 other `name` in the `rows` list. Care SHOULD be taken to avoid collisions on
 case-insensitive filesystems (e.g. avoid using both `Aa` and `aA`).
 
-The `plate` dictionary SHOULD contain a `version` key whose value MUST be a string specifying the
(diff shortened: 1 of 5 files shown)


sbesson (approved):
Until a 1.0 version of the OME-NGFF specification is published, producers/consumers should definitely expect iterations of the spec containing breaking changes. Being able to unambiguously associate a dataset to a given version is critical both when writing and reading. In that sense, enforcing the `version` field feels like a very reasonable requirement at least from my side.

Assuming this gets included (is there a formal mechanism for accepting these types of changes), is there any reason to limit this change to the `multiscales` specification or are you also considering making the field mandatory in other locations?

will-moore:
Added `version` requirement to `plate`. @sbesson OK with those changes? thx

sbesson (approved):
No objections to requiring `version` in both `plate` and `image` from my side

PR #191, opened 2023-04-27
improve description of omero metadata

giovp:
Currently, the omero metadata section in the specs could be a bit more descriptive, this is a small attempt.

This info was kindly summarized by @will-moore here: https://github.com/ome/ome-zarr-py/pull/261#issuecomment-1525658925

EDIT: while this PR is purely to improve documentation, might be related to potential relaxation of omero metadata as discussed in https://github.com/ome/ngff/issues/192

### DIFF
--- a/0.4/index.bs
+++ b/0.4/index.bs
@@ -438,6 +438,12 @@ can be found under the "omero" key in the group-level metadata:
 See https://docs.openmicroscopy.org/omero/5.6.1/developers/Web/WebGateway.html#imgdata
 for more information.
 
+The "omero" metadata is optional, but if present it MUST contain the field "channels", which is an array of dictionaries describing the channels of the image.
+Each dictionary in "channels" MUST contain the field "color", which is a string of 6 hexadecimal digits specifying the color of the channel in RGB format.
+Each dictionary in "channels" MUST contain the field "window", which is a dictionary describing the windowing of the channel.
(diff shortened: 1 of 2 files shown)


joshmoore:
Thanks, @giovp! (https://ngff--191.org.readthedocs.build/latest/index.html#omero-md)

If no one else has any comments, I'll move to get this in.

will-moore:
Looks good.

Although we normally wouldn't make changes to a spec after release, this extra info *could* be added back to the `v0.4` spec and might be useful?

But not a blocker!

will-moore:
Also with https://github.com/ome/ngff/issues/192 proposed, this change in latest would be updated again before release, so it would be good to have this description in the v0.4 spec, where it won't change.

giovp:
sorry for radio silence @will-moore @joshmoore , I have time now, shall I just add the info as optional as discussed in #192 ? I yes, I can then submit changes to IO in `ome-zarr-py` . Let me know!

will-moore:
I would copy what you have now to the v0.4 spec - since it describes that spec and is worth having there. Then (probably in #192) update it in the in `latest/index.bs` according to the proposed changes there. 
Probably easiest to keep them as separate PRs since this one just documents current state, but #192 is a change. I don't expect that to take a whole lot longer but you never know!

giovp:
done, added it to 0.4

will-moore:
Great, thanks!

PR #198, opened 2023-05-12
Add 'Edit on GitHub' link (fix #195)

### LINKED ISSUE #195: Add rtd metadata to enable doc contributions
Sphinx metadata should be set so that a GitHub link is present so users can submit additions to the /tools and /data pages (as stated in the OME-Zarr preprint)

image

cc: @will-moore @jburel

joshmoore:


joshmoore:
Doesn't look like this is necessary. The default webpage has this:

image


when clicking on the version name at the bottom. To get something more prominent is likely a question for the theme.

Issue #192, opened 2023-04-27
relax omero metadata


giovp:
### Motivation
With this issue, I would like to discuss whether ngff could relax the omero metadata requirements by making "visualization specific" parameters (["window"](https://github.com/ome/ngff/blob/15565cb8301e14fc9ba2fc63b6332a97534cee67/0.4/schemas/image.schema#L98) and ["color"](https://github.com/ome/ngff/blob/15565cb8301e14fc9ba2fc63b6332a97534cee67/0.4/schemas/image.schema#L91) ) optional.

As discussed in https://github.com/ome/ngff/issues/155 , there seems to be agreement that the channel name information (e.g. in a microscopy image, the name of the fluorescent markers/antibodies such as "DAPI", "Actin" etc.) should be saved in omero metadata. This information would be stored under "label" according to [current spec](https://ngff.openmicroscopy.org/latest/#omero-md). This type of information is critical to be stored especially in highly-multiplexed imaging techniques such as CyCIF, Codex, Imaging Mass Cytometry etc.

Users might want to save the channel name information without specifying "color" or "window" information. Currently, it is not possible, as there are two required arguments by the omero specs which are, imho, purely for "visualization" purposes. Users would therefore have to specify "arbitrary" information on aesthetics that is not necessarily useful/actionable. 

### Potential solution
Make the current required parameters ("color" and "window") as optional. This would enable the use case of only saving channel name information without having to specify the remaining visualization parameters.

will-moore:
Sounds good 👍

giovp:
sorry for radio silence @will-moore I will have time to finish this now. Shall I just add this to #191 ?

PR #202, opened 2023-07-06
remove "windows" and "color" from omero metadata as required

giovp:
ref #192 . I left the "windows" parameters as "required" but I'm not sure if it makes sense.


will-moore (approved):
Spec change looks good to me.
I don't see that this relaxation needs any changes to the tests

will-moore:
I think the failing test above is due to the trailing comma on the previous line that needs to be removed

giovp:
bump

will-moore:
Thanks!

PR #204, opened 2023-07-13
Update citation to point to latest zenodo

Czaki:
When going through the documentation I found that link to zenodo points to release 0.1 not the current one 0.4.1.

As you do not have `zenodod.json` and do not provide bibtext entry are you thinking that add one of these things could be a good idea (as bibtex generated by zenodo has different content to example citation string in the documentation)?

### DIFF
--- a/0.4/index.bs
+++ b/0.4/index.bs
@@ -657,10 +657,10 @@ Citing {#citing}
 ================
 
 [Next-generation file format (NGFF) specifications for storing bioimaging data in the cloud.](https://ngff.openmicroscopy.org/0.4)
-J. Moore, *et al*. Editors. Open Microscopy Environment Consortium, 8 February 2022.
+J. Moore, *et al*. Editors. Open Microscopy Environment Consortium, 25 may 2023.
 This edition of the specification is [https://ngff.openmicroscopy.org/0.4/](https://ngff.openmicroscopy.org/0.4/]).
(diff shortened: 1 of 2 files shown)


joshmoore:
> When going through the documentation I found that link to zenodo points to release 0.1 not the current one 0.4.1.

Thanks for the bump, @Czaki!

> As you do not have zenodod.json and do not provide bibtext entry are you thinking that add one of these things could be a good idea (as bibtex generated by zenodo has different content to example citation string in the documentation)?

zenodod.json is news to me. I had previously looked into CITATION.cff (https://github.com/ome/ngff/pull/175). Do you have a recommendation?

Czaki:
`CITATION.cff` is for me something like just list of author, when `.zenodo.json` allow to define much more detailed metadata.

Also in [zenodo FAQ](https://help.zenodo.org/faq/) there is:

> Do you support other software metadata files, like CITATION.cff or codemeta.json?
> 
> Part of our plans and ongoing work is to take into account or improve our integration of other software metadata schemas such as the [Citation File Format (CFF)](https://citation-file-format.github.io/) and [CodeMeta](https://codemeta.github.io/). We are working closely with our users and scholarly software and citation communities to figure out exactly when, in what form, and how all these different metadata schemas and their files will be integrated into Zenodo.

Also other responses suggest that `CITATION.cff` is not fully supported. 

I'm not sure what is need for this project. So I'm not sure if ability to fill all zenodo metadata from code is required for project (like license, funding etc). To read about metadata that could be stored in `.zenodo.json` see https://developers.zenodo.org/#github

PR #206, opened 2023-07-20
Zarr v3

normanrz:
This PR proposes to adopt the version 3 of the Zarr format for OME-Zarr.

Main changes:
- Only Zarr v3 is allowed for upcoming versions of OME-Zarr.
- Zarr v3 stores its metadata in `zarr.json` files. This is the same for arrays and groups. Attributes are now stored in a `attributes` object within the `zarr.json` files.
- The `chunk_key_encoding` of the Zarr arrays are respected instead of mandating the `/` dimension separator. Basically all Zarr features are available in OME-Zarr as they get approved through the ZEP process.
- Add a `ome` namespace within the `zarr.json` attributes.
- Move `version` fields from `multiscale`, `plate`, `well` etc. up into the `ome` object.
- Move the registration of labels from the `labels` group into the `multiscale` group.
- Rename the document to `OME-Zarr specification`

This PR is currently in a draft status and builds upon the #138 PR by @bogovicj.
Check this link for a more convenient review: https://github.com/ome/ngff/pull/206/files/b92f540dc95440f2d6b7012185b09c2b862aa744..HEAD


d-v-b:
> * Add a `ome` namespace within the `zarr.json` attributes.
> 
> * Move `version` fields from `multiscale`, `plate`, `well` etc. up into the `ome` object.

Thank you for this!

joshmoore:
After a quick read, @normanrz, here are my initial thoughts (where not simply a big :+1:):

> Add a ome namespace within the zarr.json attributes.
> Move version fields from multiscale, plate, well etc. up into the ome object.

These two together make me wonder whether or not the key couldn't be something more sophisticated, like the version itself:

```
"https://ngff.openmicroscopy.org/0.5": {
   ...
}
```
It's somewhat non standard to use a more complicated key like that, but it's a value that we will have in memory anyway to load the schema files, etc.

> Move the registration of labels from the labels group into the multiscale group.

The `labels` "object" was definitely poorly conceived originally and so improvements welcome, but I'll note there are a handful of open issues as well as integration with the upcoming SpatialData work that might be worth considering. If it becomes more of an issue, I'd suggest splitting this out of your proposal here (if possible).

> Rename the document to OME-Zarr specification

I know this is a constant source of confusion/frustration, but I think we might want to reconsider this.

normanrz:
> > Add a ome namespace within the zarr.json attributes.
> > Move version fields from multiscale, plate, well etc. up into the ome object.
> 
> These two together make me wonder whether or not the key couldn't be something more sophisticated, like the version itself:
> 
> ```
> "https://ngff.openmicroscopy.org/0.5": {
>    ...
> }
> ```
> 
> It's somewhat non standard to use a more complicated key like that, but it's a value that we will have in memory anyway to load the schema files, etc.

Fine with me. It is a bit more verbose, but we could get rid of the additional `"version"` attribute.



> > Move the registration of labels from the labels group into the multiscale group.
> 
> The `labels` "object" was definitely poorly conceived originally and so improvements welcome, but I'll note there are a handful of open issues as well as integration with the upcoming SpatialData work that might be worth considering. If it becomes more of an issue, I'd suggest splitting this out of your proposal here (if possible).

Also, fine with me to split it out. But I really think label images in OME-Zarr need to be reworked.



> > Rename the document to OME-Zarr specification
> 
> I know this is a constant source of confusion/frustration, but I think we might want to reconsider this.

That is certainly up for debate. But my preference would be to go with `OME-Zarr` as title of the document, because that is what it is about.

latest/index.bs
```diff
@@ -331,24 +338,1010 @@ Conforming readers:
...
+</div>
+
+The axes names of an array MAY be duplicated in the Zarr metadata key `dimension_names`. 
+Implementations MUST give precedence to the `axes` metadata over the `dimenions_names`.
+
+
+### Array coordinate systems
+
+Every array has a default coordinate system whose parameters need not be explicitly defined.  Its name is the path to the array
+in the container, its axes have `"type":"array"`, are unitless, and have default "name"s. The ith axis has `"name":"dim_i"`
```
jbms, line 492: Shouldn't the default names be those specified by the zarr v3 `dimension_names` property?

jbms: I see this is kind of addressed in the paragraph above, but it still might be helpful to clarify here.

latest/index.bs
```diff
@@ -331,24 +338,1010 @@ Conforming readers:
 - MAY ignore other groups or arrays under the root of the hierarchy.
 
 
+"coordinateSystems" metadata {#coord-sys-md}
```
jbms, line 341: It would help to clarify exactly where this is specified within the zarr attribute structure. The example could include the top-level "ome" key to clarify.   If this isn't intended as a top-level member of the "ome" object, then perhaps it should be described in a later section, after describing its "parent" object, in order to make the specification clearer.

normanrz: I think this could be addressed in #138, which this PR is based on.

normanrz:
Closing this in favor of #227. Discussion will continue there.

PR #210, opened 2023-08-01
Clarify `version` field

d-v-b:
For the various metadata objects with a `version` field, this PR makes explicit that `version` is a string, and that it must be the current version.


normanrz:
I wonder if you'd want to integrate that with https://github.com/ome/ngff/pull/206?

d-v-b:
Definitely. I see that you are building on #138 -- is there a timeline for getting that finalized (it's been nearly a year)?

joshmoore:
Agreed.

normanrz:
This is now clarified in [0.5](https://ngff.openmicroscopy.org/0.5/index.html#metadata)
> The version of the OME-Zarr Metadata is denoted as a string in the version attribute within the ome namespace.

normanrz:
Should we close this?

PR #214, opened 2023-08-23
Update index.md to add stack_to_multiscale_ngff

AlanMWatson:
Added pointer to stack_to_multiscale_ngff, a python package for high-performance conversion of image stacks to ome-ngff


joshmoore:
Thanks, @AlanMWatson.

PR #224, opened 2024-01-15
Attempt to fix calendar timezone

joshmoore:
Apparently the `<script/>` block needs to be on the same link with the `<iframe/>` so that the timezone shown to readers will be their own.


joshmoore:
Got a 👍🏽 from @jburel

PR #226, opened 2024-02-12
Add codespell to detect new typos and fix existing ones (even in prior versions since why not?)

yarikoptic:
See https://github.com/codespell-project/codespell for more info about this handy tool


joshmoore:
Thanks, @yarikoptic!

Issue #207, opened 2023-07-24
Single-scale images


clbarnes:
Images are not always scale pyramids, but single-image arrays also benefit from some of the metadata we apply to multiscale groups. I suppose these are implicitly supported in the spec ("just add a `coordinateTransformations` and `axes` specified elsewhere") but IMO it would be useful to define single-scale datasets as their own valid type. The use case for me is extracting a small ROI from a single scale level of a larger volume.

You can, of course, define a multiscale group with only one dataset in it, but then it becomes quite verbose, as well as having to bounce between the group and array metadata.

This is sort of the opposite of https://github.com/ome/ngff/issues/187 : that asks to define some image metadata in a group above the containing group, where this asks to define some image metadata on the array itself.

d-v-b:
Ideally, a single resolution image would "just" be a multiscale image with one scale level. This would be nice for a lot of reasons -- OME-NGFF viewers will know how to deal with it, and if you later decide that you do need multiple levels of resolution, you can drop them in without reworking your metadata. You refer to a small ROI, but there's no guarantee that the next one is small :) .

But as you note (and as I have experienced), the current multiscale metadata spec is unwieldy, both because it is needlessly verbose (specifying `coordinateTransformations` at the top-level _and_ per-dataset, when only per-dataset is needed) and because it keeps the spatial metadata for an image away from the image itself, which leads to parsers ping-ponging between the image you want to load and the group that defines how the interpret that image.

Purely hypothetically, consider the following minimized multiscale metadata scheme:

Group metadata:
```
ome: {
    multiscales: {
        metadata: {...}, # description of the multiscale pyramid 
        datasets: ["s0"] # name(s) of array(s) contained in this group
        }
}  
```

`multiscales` is drastically simplified. it no longer a list, and it just contains `metadata` and a list of strings which MUST be names of arrays contained in the group bearing the `multiscales` metadata. 

Array metadata (e.g., for `s0`)
```
ome: {
    axes: [...],
    transforms: {
         scale: [...],
         translation: [...]
          }
}
```
array metadata contains the information required to embed the array in physical space. seems legit.

I'm namespacing all the ome-ngff stuff under a `ome` keyword, inspired by #zarrv3

if we had something this compact for multiscale metadata, would you still want a separate single-scale schema @clbarnes ?

clbarnes:
[...]

PR #227, opened 2024-02-15
RFC: Zarr v3

normanrz:
This is an RFC proposal for adopting Zarr v3 as the new storage format for OME-Zarr.

It is a followup to the discussions in #206 and on [image.sc](https://forum.image.sc/t/adopt-zarr-v3-in-ome-zarr/84786).

Briefly, this proposal aims to adopt Zarr v3 as the new format for the next version of OME-Zarr. That unlocks the new features of Zarr v3 including sharding. Zarr v2 would not be allowed anymore (only through older versions of OME-Zarr). Additionally, there are some small changes to the OME-Zarr metadata to improve namespacing and versioning.

This RFC is currently in draft status with the goal of clarifying questions before the full review. Additional endorsements are also welcome.

Check this link for a review: https://ngff--227.org.readthedocs.build/rfc/2/index.html


kevinyamauchi:
I think moving to zarr v3 is a great step. I would be happy to "endorse" this RFC.

Thanks, @normanrz !

clbarnes:
Does the sharding codec need to be detailed here or can we just name-drop it as an advantage of zarr v3 and then link to zarr's information about it?

There are a couple of things which could change in NGFF to make it more zarr-y. They're not blockers to adoption, just something to reduce downstream users' headaches, which IMO will never be addressed if not now.

One is the the case convention: zarr uses `snake_case`, NGFF uses `camelCase`.

The other is how enums are tagged: zarr uses adjacently tagged (`{"name": "something", "config": {...}}`) where NGFF uses internally tagged (`{"name": "something", "somekey": ..., "anotherkey": ...}`).

d-v-b:
> One is the the case convention: zarr uses `snake_case`, NGFF uses `camelCase`.
> 
> The other is how enums are tagged: zarr uses adjacently tagged (`{"name": "something", "config": {"somekey": ..., "anotherkey": ...}}`) where NGFF uses internally tagged (`{"name": "something", "somekey": ..., "anotherkey": ...}`).

These are both very good points -- on the latter, you might want to weigh in over at https://github.com/ome/ngff/pull/138, where a lot of new enums are being minted.

normanrz:
> Does the sharding codec need to be detailed here or can we just name-drop it as an advantage of zarr v3 and then link to zarr's information about it?

I think sharding is a major motivation to move to v3. That is why it gets so much space in my proposal. It is intended as background information and, in the end, has no direct implications on the OME-Zarr spec.

> There are a couple of things which could change in NGFF to make it more zarr-y. They're not blockers to adoption, just something to reduce downstream users' headaches, which IMO will never be addressed if not now.
> 
> One is the the case convention: zarr uses `snake_case`, NGFF uses `camelCase`.
> 
> The other is how enums are tagged: zarr uses adjacently tagged (`{"name": "something", "config": {"somekey": ..., "anotherkey": ...}}`) where NGFF uses internally tagged (`{"name": "something", "somekey": ..., "anotherkey": ...}`).

Pinging @joshmoore for advice on whether that should be a separate RFC. I think it is possible to bundle multiple RFCs into one new version of the spec.

bogovicj:
Thanks @normanrz ! I'm also happy to endorse this RFC! :

matthewh-ebi:
Also happy to endorse, this would be very beneficial for our use cases, huge thanks @normanrz !

tischi:
Thanks @normanrz, I am also happy to endorse this RFC.

jluethi:
Thanks a lot @normanrz , I am also happy to endorse this RFC! It will be great to get sharding for OME-Zarrs.

constantinpape:
I also endorse this RFC!

will-moore:
I'm happy to endorse this RFC! 👍

jni:
Obviously I'm all in favour of supporting v3, but:

> Zarr v2 would not be allowed anymore (only through older versions of OME-Zarr).

What is the motivation for this? Why should we couple ome-zarr and zarr so tightly? If someone has ome-zarr v0.5 (or whatever) metadata at the root of a v2 zarr folder, why should that be forbidden?

jni:
Gaaaah, don't comment before reading the article! 😅 Just read:

> The metadata of Zarr v3 arrays are not backwards compatible with Zarr v2.

which explains it. However, it does still seem lightweight to support both on the ome side?

normanrz:
> However, it does still seem lightweight to support both on the ome side?

While it is easy to support both versions in the OME spec document, I'm concerned with the complexity burden for implementations. I'd rather not add one dimension to the compatibility matrix.
OME-Zarr implementations that build upon libraries probably have good support for v2 and v3 at the moment. However, this might change in the future (anyone remember what happened to Zarr v1?). For example, in zarr-python, we are working on a refactoring that is v3-first. It is not unlikely that in the future v2 support will become deprecated. 
Also, there are implementations that roll their own Zarr stack that would need to add support for v2 and v3. 

From a OME-Zarr user perspective, the hard cut also makes things simpler: ≤ 0.5 => Zarr v2; > 0.5 => Zarr v3 (or whatever the version number will be). If users wish to upgrade their data from one OME-Zarr version to another, it would be easy to also migrate the core Zarr metadata to v3. This is a fairly cheap operation, because only json files are touched. Zarr v2 and v3 metadata could even live side-by-side in the same hierarchy. There are functions available that can migrate the metadata automatically (e.g. in [zarrita](https://github.com/scalableminds/zarrita/blob/async/zarrita/array_v2.py#L452-L559) and soon zarr-python).

jni:
Sure, I guess it is indeed easier as a user to know if you have an ome-zarr v0.5 file that all readers would support it, rather than have to understand whether your reader supports zarr v2. It would indeed be easy to get into a situation like "does this USB cable support data transfer and at what speed?" 😅 

Anyway, please take my question as more for my own information rather than as a blocker: I too am happy to endorse this plan. 😊

joshmoore:
> [normanrz](https://github.com/normanrz) commented [4 days ago](https://github.com/ome/ngff/pull/227#issuecomment-1948298152)
> Pinging @joshmoore for advice on whether that should be a separate RFC. I think it is possible to bundle multiple RFCs into one new version of the spec.

The current plan is definitely to collect multiple RFCs into a single spec version, but that being said, I personally don't think sharding needs a separate RFC. With the move to v3, we have the chance to decouple the NGFF specific from specifics of the backend (looking at you, dimension separator), so I agree with @clbarnes that should be more referencing an existing spec (ZEP1 & ZEP2) *but* we do need to make users & implementers aware of the trade-offs that a given backend provides them. It's going to be a fine balance.

> [normanrz](https://github.com/normanrz) commented [3 hours ago](https://github.com/ome/ngff/pull/227#issuecomment-1953676115)
> > However, it does still seem lightweight to support both on the ome side?
> 
> While it is easy to support both versions in the OME spec document,

I still need to go through the text, but a :+1: for including any explanations you give here in the main text if they are not already there.

joshmoore:
> [joshmoore](https://github.com/joshmoore) commented [19 hours ago](https://github.com/ome/ngff/pull/227#issuecomment-1954338710)]
> ... so I agree with @clbarnes that should be more referencing an existing spec (ZEP1 & ZEP2)...

@normanrz gently pointed out that I had misunderstood his question. The point was whether or not this RFC should include issues about camelCasing, etc. I would think *not*. Judging simply be the amount of endorsers this has already received, adding more issues especially ones that are prone to bike shedding can only make things *less* clear. (And for such topics, it likely makes sense to do some consensus building *outside* of the RFC before bringing it for review (D2 "gather support") but that's beyond the scope of this PR. :smile:)

normanrz:
Thanks for the endorsements and feedback! I moved the proposal to draft state D3, which is intended to clarify questions before the review. Any feedback and questions are appreciated. More endorsements are also very welcome!

Based on the feedback, I moved the sections about sharding to "Background" (because it is not really part of this RFC; just illustrates a motivation for adopting v3) and added the motivation for dropping v2 support in the text.

rfc/2/index.md
```diff
@@ -0,0 +1,458 @@
...
+  - [crc32c](https://zarr-specs.readthedocs.io/en/latest/v3/codecs/crc32c/v1.0.html) for decorating chunks with a checksum
+  - [sharding_indexed](https://zarr-specs.readthedocs.io/en/latest/v3/codecs/sharding-indexed/v1.0.html) for storing sharded arrays (see below).
+
+The Zarr specification does not prescribe the support stores for Zarr hierarchies.
+HTTP(S), File system, S3, GCS, and Zip files are commonly used stores.
+
+With this proposal all features of the Zarr specification are allowed in OME-Zarr.
+In the future, the OME-Zarr community MAY decide to restrict the allowed feature set.
+
+### Changes to the OME-Zarr metadataa
```
yarikoptic, line 124: minor typo

```suggestion
### Changes to the OME-Zarr metadata
```

perlman:
+1 to the chorus of endorsements.

I am looking forward to using Zarr v3 with several of my hats.

rfc/2/index.md
```diff
@@ -0,0 +1,459 @@
...
+                  "name": "z",
+                  "type": "space",
+                  "unit": "nanometer"
+                }
+              ]
+            }
+          ],
+
+          "datasets": [
+            {
```
ziw-liu, line 287: The current (v0.4) version specifies that image paths `MUST` be alphanumeric. Is it within the scope of this RFC to relax that constraint?

https://github.com/ome/ngff/blob/b43000c65fd4a0cde67a2f32da447deb856b774b/0.4/index.bs#L582-L588

ziw-liu: Sorry I misread this to be the HCS paths...

rfc/2/index.md
```diff
@@ -0,0 +1,459 @@
...
+          }
+        ],
+        "index_codecs": [{ "name": "bytes" }, { "name": "crc32c" }],
+        "index_location": "end"
+      },
+      "name": "sharding_indexed"
+    }
+  ],
+  "attributes": {},
+  "dimension_names": ["c", "x", "y", "z"]
```
ziw-liu, line 392: The `"dimension_names"` field is new to Zarr v3. Will this RFC define the relationship between this field and the `"coordinateSystem"` metadata? For example, it may be specified that their values `MUST` be consistent.

ziw-liu:
Thanks @normanrz! I endorse this RFC.

rfc/2/index.md
```diff
@@ -0,0 +1,459 @@
...
+Sharding provides a mechanism to store multiple chunks within one file/object.
+This can greatly reduce the number of files/objects that are required for large Zarr arrays, while preserving fast access to individual chunks and parallel writing capabilities.
+Sharding solves some pain points that are also greatly felt by users in the OME-Zarr community.
+Adopting Zarr v3 in OME-Zarr is a precondition for using sharding.
+
+Library support for Zarr v3 is already available for several languages:
+
+- [tensorstore (C++/Python)](https://github.com/google/tensorstore)
+- [zarrita (Python)](https://github.com/scalableminds/zarrita)
+- [zarrita.js (JS)](https://github.com/manzt/zarrita.js/)
```
yarikoptic, line 54:
```suggestion
- [zarrita.js (JS)](https://github.com/manzt/zarrita.js)
```
Just for consistency

rfc/2/index.md
```diff
@@ -0,0 +1,459 @@
...
+Each shard contains an index that contains references to the inner chunks that are stored within a shard.
+Inner chunks are compressed individually, if such a codec is specified.
+Implementations can read inner chunks individually.
+Depending on the choice of codecs and the underlying storage backends, it may be possible to write inner chunks individually.
+However, in the general case, writing is limited to entire shards.
+
+## Proposal
+
+This RFC proposes to adopt version 3 of the Zarr format for OME-Zarr.
+Version 2 will no longer be supported.
```
yarikoptic, line 88: In what sense? What will happen with data already in zarr v 2 - will it remain at least readable in foreseeable future?

rfc/2/index.md
```diff
@@ -0,0 +1,459 @@
...
+HTTP(S), File system, S3, GCS, and Zip files are commonly used stores.
+
+With this proposal all features of the Zarr specification are allowed in OME-Zarr.
+In the future, the OME-Zarr community MAY decide to restrict the allowed feature set.
+
+### Changes to the OME-Zarr metadata
+
+While the adoption of Zarr v3 does not strictly require changes to the OME-Zarr metadata, this proposal contains changes to align with community conventions and ease implementation:
+
+- OME-Zarr metadata will be stored under a dedicated key in the Zarr array or group attributes. The key will be a well-known URI of the OME-NGFF specification with a version number, e.g. `https://ngff.openmicroscopy.org/0.6`.
```
yarikoptic, line 129: "E.g" or that specific one for ever? Or it would be over starting with that specific uri and version component possibly changing?

Why not just "ngff" and have dedicated and clearly defined field "Version"? That would simplify accessing such key (instead of matching by leading uri if accommodating multiple versions).

rfc/2/index.md
normanrz: I think that would be a reasonable requirement. I'll add it to the text.

rfc/2/index.md
normanrz: That depends on the implementations. If libraries continue to support OME-Zarr ≤ 0.5, Zarr 2 will remain accessible. As I noted in another section of the proposal, it is fairly easy to migrate Zarr v2 to v3 without rewriting the chunk data.

rfc/2/index.md
normanrz: The idea was to reference the version in the key. Yes, that would require a prefix scan, which I don't think is a big problem because most implementation parse the entire json, anyways. The upside is that you could have multiple OME metadata versions in the same document, e.g. `https://ngff.openmicroscopy.org/0.6` and `https://ngff.openmicroscopy.org/1.0`.
@joshmoore suggested this approach in favor of having a generic "ome" key with an additional "version" key in https://github.com/ome/ngff/pull/206#issuecomment-1746870625

rfc/2/index.md
yarikoptic: Wording remains unclear -- "no longer be supported" by what? by libraries? by standard itself?

if "easy" -- why not to be instructed/recommended to be done on the fly by the libraries?

My motivation: on dandiarchive.org we already have many TBs of OME-Zarr v2.  Even if it is "easy" to convert, once data is published, and DOI is minted, data must not mutate. 
Ideally a standard should remain stable enough to allow access to data for years coming. Otherwise it would be hard(er) to convince people to place data in such a "standard form" which would not be accessible within just a handful of years.

rfc/2/index.md
yarikoptic: That's "cute" -- thanks for the explanation/reference!  I wonder if use of a versioned URI to "allocate" a namespace here was ra- by a wider Zarr community?  not that it would not be permitted but rather since I believe it is first/unique use there (IIRC it standed out whenever I saw it among other keys) and wider Zarr might either like to adopt it wider or raise some concerns.

My immediate question is: what should be a "standard" behavior when multiple versions are encountered? could be "the latter supported is taken".  And either such behavior also should be formalized at Zarr or NGFF levels. Might be worth spelling out as well ahead of time if allowing for such multi-version specification.

rfc/2/index.md
normanrz: > Wording remains unclear -- "no longer be supported" by what? by libraries? by standard itself?

This means that data that uses OME-Zarr version 0.6 (or whatever the version number ends up being for this RFC) or higher will have to use Zarr v3 instead of v2. Data that uses earlier versions of OME-Zarr are not affected by this requirement.

> if "easy" -- why not to be instructed/recommended to be done on the fly by the libraries?

My hope is that implementations still continue to support a wide range of OME-Zarr versions, including the current 0.4. In the end that is up to each implementation to decide. I think [this table](https://ome.github.io/ome-ngff-tools/) should get another dimension for the OME-Zarr version to track implementation support. I don't think it is in scope for this RFC to make recommendations about support for earlier versions of the spec, though.

> My motivation: on dandiarchive.org we already have many TBs of OME-Zarr v2. Even if it is "easy" to convert, once data is published, and DOI is minted, data must not mutate.
> Ideally a standard should remain stable enough to allow access to data for years coming. Otherwise it would be hard(er) to convince people to place data in such a "standard form" which would not be accessible within just a handful of years.

Since the data in an archive is immutable, it wouldn't be upgraded to new versions of OME-Zarr. That would mean it is not affected by this RFC. If new data is published under the new version of OME-Zarr, it needs to use Zarr v3 under the hood.

There is a case to be made to abandon version numbers and require backwards compatibility on a spec level. Personally, I think that is too early for OME-Zarr. That is something a later RFC could pick up.

rfc/2/index.md
normanrz: > That's "cute" -- thanks for the explanation/reference! I wonder if use of a versioned URI to "allocate" a namespace here was ra- by a wider Zarr community? not that it would not be permitted but rather since I believe it is first/unique use there (IIRC it standed out whenever I saw it among other keys) and wider Zarr might either like to adopt it wider or raise some concerns.

There is an [open discussion on the Zarr level about metadata conventions](https://zarr.dev/zeps/draft/ZEP0004.html), of which NGFF/OME-Zarr is one. That discussion has not reached a conclusion yet. And, I don't think the OME community necessarily needs to wait for a conclusion.

Generally, URLs are used in other places in the Zarr spec, e.g. for [names of extensions](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#extension-points). So, it does not seem entirely foreign.

> My immediate question is: what should be a "standard" behavior when multiple versions are encountered? could be "the latter supported is taken". And either such behavior also should be formalized at Zarr or NGFF levels. Might be worth spelling out as well ahead of time if allowing for such multi-version specification.

That is currently not defined. I almost think that implementations should decide which metadata they should follow. My thinking is that image data and metadata must match. That is true for every version of the metadata that may exist. So, implementations can choose which version they understand and use that to interpret the image data. If implementations support multiple versions, it would be reasonable to use the most recent one.

For example, you could have an OME-Zarr image that has the 0.4 metadata (based on Zarr v2) and the 0.6 metadata (based on Zarr v3). That is possible, because the chunk data is mostly compatible and just the json files differ. Implementations that don't support Zarr v3 can still read the v2 version and vice versa.

rfc/2/index.md
ziw-liu: Thanks for the update! (I don't have permission to mark this as resolved.)

LDeakin:
I endorse this RFC.

Some comments:
- I like that the OME-Zarr metadata can now also be stored in array attributes. That partially satisfies this request https://github.com/ome/ngff/issues/207. `multiscales` in an array will need additional constraints on the `datasets` part though (e.g., single element with empty path).
- I initially thought there were some unspecified changes to the `coordinateTransformations` metadata in the example, but I see it is based on this draft https://github.com/ome/ngff/pull/138.

normanrz:
> * I like that the OME-Zarr metadata can now also be stored in array attributes. That partially satisfies this request https://github.com/ome/ngff/issues/207. multiscales in an array will need additional constraints on the datasets part though (e.g., single element with empty path).

While I would like that, I didn't intend to change this behavior as part of this RFC.

joshmoore:
As described in https://forum.image.sc/t/ome-ngff-update-postponing-transforms-previously-v0-5/95617/2, merging this to move forward with the first round of reviews. Thanks all for your feedback & endorsements.

mkitti:
Commenting solely in an individual capacity, I endorse the premise that "OME-Zarr should adopt Zarr v3 as the storage format". I have recently deployed Zarr v3 shards within Janelia on behalf of Philip Keller's Lab for use with neuroglancer, tensorstore, and other compatible tools.

The contents of the RFC itself is confusing. It seems to mostly replicate parts of the Zarr v3 specification and highlight key changes. Rather than duplicate such specification content, the RFC should mainly reference the canonical Zarr v3 specification.

Lacking from the RFC are content pertinent to OME-Zarr as aa standard. I would particularly like to see the following.

1. Details about how OME metadata would be integrated into the new storage format.
2. Clarity about the status of image data in Zarr v2. Is Zarr v2 deprecated? Do we plan to maintain Zarr v2 as part of the OME-Zarr specification?
3. Should there be OME-Zarr v2 and OME-Zarr v3 specifications?
4. Guidance on how to transition Zarr v2 content to Zarr v3.

normanrz:
Thanks for your feedback @mkitti!
Apart from just referencing the Zarr v3 spec, I thought it would be useful to highlight some of the v3 features as a motivation for adopting the new version. I didn't want to assume that everybody in this community is aware of all the changes in Zarr v3. I'll make sure to better separate that from the actual changes to OME-Zarr in the next iteration of the RFC after the first round of reviews. I also hope that things will become clearer once I added the changes to the spec document.

> Lacking from the RFC is content pertinent to OME-Zarr as a standard. I would particularly like to see the following.
> 
>  1. Details about how OME metadata would be integrated into the new storage format.

The OME metadata will be stored in the `attributes` of the the groups' `zarr.json` files under a new versioned namespace `https://ngff.openmicroscopy.org/0.5`. This is explained in [this section in the RFC](https://ngff.openmicroscopy.org/rfc/2/index.html#changes-to-the-ome-zarr-metadata). 

>  2. Clarity about the status of image data in Zarr v2. Is Zarr v2 deprecated? Do we plan to maintain Zarr v2 as part of the OME-Zarr specification?
>   3. Should there be OME-Zarr v2 and OME-Zarr v3 specifications?

OME-Zarr <0.5 only supports Zarr v2 and OME-Zarr ≥0.5 will only support Zarr v3. It is recommended that implementation support a number of OME-Zarr versions to support for reading existing data. I think that recommendation is useful not only for this RFC. This is discussed in [this section in the RFC](https://ngff.openmicroscopy.org/rfc/2/index.html#proposal).

>   4. Further guidance on how to transition Zarr v2 content to Zarr v3. For each of the metadata examples in the current specification, how do those appear in the Zarr v3 specification?

There is some mention of migration scripts [in this section of the RFC](https://ngff.openmicroscopy.org/rfc/2/index.html#backwards-compatibility). The next iteration of the RFC after the first round of reviews will also contain the changes to the spec document, with updated examples.

>   5. Is it compliant for an array to contain both Zarr v2 and Zarr v3 metadata files side by side?

Yes, Zarr v2 and v3 metadata files can exist side-by-side. As a consequence, OME-Zarr 0.4 and 0.5 metadata should also be able to exist side-by-side. With the new versioned namespace, this will be even easier for future versions. We should probably add an explicit recommendation that implementations prefer the newest version of the metadata that they support.

PR #230, opened 2024-03-07
RFC1 review

melonora:
This is a review of RFC 1 which is authored by @joshmoore.

co-authored by: m.lucalmer@gmail.com


joshmoore:
Many thanks for your comment. I will merge this and in a follow-up implement suggestions and cross-link your comment from the RFC.

PR #231, opened 2024-03-13
RFC 1 Updates

joshmoore:
This PR updates RFC 1 in the following ways:
 * [x] Lists current reviewers and commenters
 * [x] Adds a legend for symbols like 🕑 and 📂
 * [x] Adds a section on decision-making
 * [ ] Add a changelog to the RFC

It additionally:
 * [x] removes the previous use of the term "Editor" throughout the specification to prevent confusion
 * [ ] clarifies how to suggest someone in particular (including oneself) as a reviewer


thewtex (approved):
Thanks for incorporating this!

PR #242, opened 2024-05-31
Changes to the spec and schemata for RFC-2

normanrz:
This is the companion PR for RFC-2 which adds the changes to the spec document, json schemata, examples and test files. The PR is meant to support the review process of RFC-2 by providing the specifics.

Again, a brief summary of the main changes:
- Zarr v3 is used for OME-Zarr that includes that all metadata moves from `.zattrs` to `zarr.json`
- The OME-Zarr metadata will live under the `ome` key in the `attributes` of the `zarr.json` files
- Renaming the spec doc to OME-Zarr


latest/index.bs
```diff
@@ -192,7 +185,10 @@ keys as specified below for discovering certain types of data, especially images
     - Units for "space" axes: 'angstrom', 'attometer', 'centimeter', 'decimeter', 'exameter', 'femtometer', 'foot', 'gigameter', 'hectometer', 'inch', 'kilometer', 'megameter', 'meter', 'micrometer', 'mile', 'millimeter', 'nanometer', 'parsec', 'petameter', 'picometer', 'terameter', 'yard', 'yoctometer', 'yottameter', 'zeptometer', 'zettameter'
     - Units for "time" axes: 'attosecond', 'centisecond', 'day', 'decisecond', 'exasecond', 'femtosecond', 'gigasecond', 'hectosecond', 'hour', 'kilosecond', 'megasecond', 'microsecond', 'millisecond', 'minute', 'nanosecond', 'petasecond', 'picosecond', 'second', 'terasecond', 'yoctosecond', 'yottasecond', 'zeptosecond', 'zettasecond'
 
-If part of [[#multiscale-md]], the length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
+The "axes" are used as part of [[#multiscale-md]]. The length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
+
+The "dimension_names" attribute in the `zarr.json` of the Zarr array of a multiscale level MUS match the names in the "axes" metadata.
```
will-moore, line 190: typo `MUS`. Is the `dimension_names` required by the Zarr v3 spec? If not, then maybe we don't need to require it be duplicated here?

normanrz: `dimension_names` is not mandatory in the Zarr spec. However, I think it is useful to have for interoperability because the `dimension_names` lives in the array metadata and `axes` in the group metadata.

d-v-b: +1 for making `dimension_names` required for ome-ngff arrays

latest/index.bs
```diff
@@ -24,19 +24,13 @@ Status Text: will be provided between numbered versions. Data written with these
 Status Text: (an "editor's draft") will not necessarily be supported.
```
will-moore, line 24: Row above: `Status Text: <a href="../0.4/index.html">0.4</a>.` looks like it needs manual update to `0.5`?

Same for line 612: `This edition of the specification is [https://ngff.openmicroscopy.org/0.4/](https://ngff.openmicroscopy.org/0.4/]).`

latest/index.bs
```diff
@@ -56,75 +50,66 @@ implementations that are later submitted as a formal specification. (See [[#bf2r
...
+    │                         # metadata as specified in the array's `zarr.json`.
     │
     └── labels
         │
-        ├── .zgroup           # The labels group is a container which holds a list of labels to make the objects easily discoverable
-        │
-        ├── .zattrs           # All labels will be listed in `.zattrs` e.g. `{ "labels": [ "original/0" ] }`
-        │                     # Each dimension of the label `(t, c, z, y, x)` should be either the same as the
+        ├── zarr.json         # The labels group is a container which holds a list of labels to make the objects easily discoverable
+        │                     # All labels will be listed in `zarr.json` e.g. `{ "labels": [ "original/0" ] }`
```
will-moore, line 99: Previously, the `.zattrs` in it's entirety was e.g. `{ "labels": [ "original/0" ] }` but now the `zarr.json` will be e.g:

```
{
    "zarr_format": 3,
    "node_type": "group",
    "attributes": {
        "labels": [ "original/0" ]
    }
}
```
Is that right? If so, this is probably too verbose to include in-line here, but maybe some rewording could describe/summarise that? E.g. 

All labels will be listed within `zarr.json` e.g. `"attributes": { "labels": [ "original/0" ] }`

latest/index.bs
```diff
@@ -148,39 +133,47 @@ A well group SHOULD NOT be present if there are no images in the well.
...
+.
 │
-└── 5966.zarr                 # One plate (id=5966) converted to Zarr
-    ├── .zgroup
-    ├── .zattrs               # Implements "plate" specification
+└── 5966.zarr                 # One OME-Zarr plate (id=5966)
+    ├── zarr.json             # Implements "plate" specification
     ├── A                     # First row of the plate
-    │   ├── .zgroup
+    │   ├── zarr.json
```
will-moore, line 141: Does this `zarr.json` need to be present? I'm not entirely clear from reading the Zarr v3 spec whether you are allowed to have empty directories? (or if the rules are different from Zarr v2 with .zgroup)?

If you aren't allowed empty directories, then does this need a change/clarification to the labels section above where we have:

```
Intermediate folders are permitted but not necessary and currently contain no extra metadata
```
Do we need `zarr.json` to be shown within the `original` directory?

latest/index.bs
will-moore: Ah, I realise this will actually need `"attributes": {"https://ngff.openmicroscopy.org/0.5": { "labels": [ "original/0" ] }}`

latest/index.bs
normanrz: There is an [ongoing discussion](https://github.com/zarr-developers/zarr-specs/issues/291) about "implicit" groups. I think the community is leaning towards disallowing these, i.e. requiring `zarr.json` files for intermediate folders.

will-moore:
Unchanged in this PR but `index.bs` still has:
```
Each "multiscales" dictionary SHOULD contain the field "name". It MUST contain the field "version", which indicates the version of the multiscale metadata of this image (current version is [NGFFVERSION]).
```
Even the example below that text doesn't contain `version`. Also it seems that `version` is no-longer needed since that's provided by the `https://ngff.openmicroscopy.org/0.5` key?

latest/index.bs
```diff
@@ -148,39 +133,47 @@ A well group SHOULD NOT be present if there are no images in the well.
...
+=============================
+
+The "OME-Zarr Metadata" contains metadata keys as specified below 
+for discovering certain types of data, especially images.
 
-The various `.zattrs` files throughout the above array hierarchy may contain metadata
-keys as specified below for discovering certain types of data, especially images.
+The OME-Zarr Metadata is stored in the various `zarr.json` files throughout the above array
+hierarchy. In this file, the metadata is stored under the namespaced key 
+`https://ngff.openmicroscopy.org/0.5` in `attributes`.
```
will-moore, line 165: Do we want to say anything here about whether we SHOULD or MUST use the same version within a hierarchy?
E.g. If Plate is v0.5, then can a reader assume that images within it are the same version?
I'm assuming we wouldn't want to support a Plate where some images are v0.4 (Zarr v2) and some are v0.5 (Zarr v3). This has probably always been an assumption by readers & writers but I'm not sure that the spec forbids it?

will-moore:
In looking to implement support for reading the proposed V0.5 data (in ome-ngff-validator), I am finding the usage of the versioned key `https://ngff.openmicroscopy.org/0.5` is a bit painful when you want to `get_version()` since I have to iterate through a list of potential versions to check if the key exists.
This means that I will always have to update the code to support new versions, instead of retrieving the `version` and using that to automatically pick the correct schema to validate against.

So I find that I am in agreement with various comments on RFC-2 about the concerns of using a version string as a key.

d-v-b:
Instead of using a URL-with-a-version-inside as a key, I think it would be better to pick a name like "ome" or "ome-ngff" as the key for an object, and have a `version` field in that object, and a `schema_url` field in that object. much clearer, and it would allow parsers to check the version of the metadata without knowing the version beforehand.

normanrz:
I updated this PR for the RFC-2 revision. The namespace key is now `ome` and there is a separate `version` attribute.

rfc/2/index.md
```diff
@@ -130,7 +130,7 @@ In the future, the OME-Zarr community MAY decide to restrict the allowed feature
 
 While the adoption of Zarr v3 does not strictly require changes to the OME-Zarr metadata, this proposal contains changes to align with community conventions and ease implementation:
 
-- OME-Zarr metadata will be stored under a dedicated key in the Zarr array or group attributes. The key will be a well-known URI of the OME-NGFF specification with a version number, e.g. `https://ngff.openmicroscopy.org/0.6`.
+- OME-Zarr metadata will be stored under a dedicated key in the Zarr array or group attributes. The key will be a well-known URI of the OME-NGFF specification with a version number, e.g. `https://ngff.openmicroscopy.org/0.5`.
```
will-moore, line 133: This sentence needs updating to the `ome` key now.
Also I wonder if you could update the PR description too, thanks.

normanrz: This change will be superseded by #250 
```suggestion
- OME-Zarr metadata will be stored under a dedicated key in the Zarr array or group attributes. The key will be a well-known URI of the OME-NGFF specification with a version number, e.g. `https://ngff.openmicroscopy.org/0.6`.
```

will-moore: OK, understood, thanks.

latest/index.bs
will-moore: I added `dimension_names` validation into https://github.com/ome/ome-ngff-validator/pull/36

normanrz:
Is there a json schema for the base `zarr.json` where we could plug in the OME-Zarr metadata schema? cc @d-v-b

d-v-b:
> Is there a json schema for the base `zarr.json` where we could plug in the OME-Zarr metadata schema? cc @d-v-b

I'm not aware of one, but we should a) make one b) include it with the zarr v3 spec. Were I to work on this today, I would start by fixing up the rather meager [v3 support in pydantic-zarr](https://github.com/janelia-cellmap/pydantic-zarr/blob/main/src/pydantic_zarr/v3.py), and then use that to generate the schema. But any way of generating such a schema is valid.

latest/index.bs
```diff
@@ -192,7 +189,10 @@ keys as specified below for discovering certain types of data, especially images
     - Units for "space" axes: 'angstrom', 'attometer', 'centimeter', 'decimeter', 'exameter', 'femtometer', 'foot', 'gigameter', 'hectometer', 'inch', 'kilometer', 'megameter', 'meter', 'micrometer', 'mile', 'millimeter', 'nanometer', 'parsec', 'petameter', 'picometer', 'terameter', 'yard', 'yoctometer', 'yottameter', 'zeptometer', 'zettameter'
     - Units for "time" axes: 'attosecond', 'centisecond', 'day', 'decisecond', 'exasecond', 'femtosecond', 'gigasecond', 'hectosecond', 'hour', 'kilosecond', 'megasecond', 'microsecond', 'millisecond', 'minute', 'nanosecond', 'petasecond', 'picosecond', 'second', 'terasecond', 'yoctosecond', 'yottasecond', 'zeptosecond', 'zettasecond'
 
-If part of [[#multiscale-md]], the length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
+The "axes" are used as part of [[#multiscale-md]]. The length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
+
+The "dimension_names" attribute in the `zarr.json` of the Zarr array of a multiscale level MUST match the names in the "axes" metadata.
```
LDeakin, line 194: "dimension_names" are redundant in an OME-Zarr multiscale image, so must they be mandatory? Perhaps this restriction could be relaxed to something like:
 > If the "dimension_names" attribute is specified in the `zarr.json` of the Zarr array of a multilscale level, it SHOULD match the names in the "axes" metadata.

This will enable arrays with undesirable/non-descriptive/missing dimension names to be used in an OME-Zarr hierarchy without any array metadata changes.

normanrz: I am open to discuss this. Weakening this restriction could cause conflicts between the array metadata and the OME-Zarr metadata. We would need to define a precedence order. 
I wonder what the circumstances would be that you can add the OME-Zarr metadata on the group level, but cannot adjust the array metadata to match the "dimensions_names" attribute?

LDeakin: I can always work around this, so it is not essential.

I have legacy multiscale arrays that have been converted from NetCDF that I would *prefer* to keep immutable and a one-to-one mapping to their source. I want to slap OME-Zarr on top with sensible `axes` names (e.g. `z`, `y`, `x`). However, the `dimension_names` of the underlying arrays may encode other information and be inconsistent between scales (e.g. `segmented_x_1.23_um`).

LDeakin: The 0.4 spec has many restrictions on the underlying arrays (consecutively numbered groups, the order of dimensions, nested directory layout, etc.) that have since been addressed by this RFC and [RFC-3](https://github.com/ome/ngff/pull/239). As far as I can see[^1], the `dimension_names` restriction introduced in this RFC is the only remaining restriction that could make arrays incompatible with OME-Zarr metadata (aside from requiring the length of `axes` to match the number of dimensions of the arrays, which is necessary).

[^1]: I've only read the spec for `multiscales` and dependent metadata

normanrz: @d-v-b What do you think about this? I believe you advocated for strictly keeping `dimension_names` in sync.

d-v-b: I think they should be kept in sync, because the alternative is confusing -- how should clients interpret variance between `axes` and `dimension_names`? And if clients are supposed to just ignore `dimension_names`, then why did we add it to the zarr v3 spec in the first place?

joshmoore:
The three approving reviews of RFC-2 have now been merged: https://github.com/ome/ngff/pull/261

Minor changes here to address the above discussions, https://github.com/ome/ngff/issues/259 and any issues of versioning, etc. are welcome.

normanrz:
I changed the JSON schema files to use the `attributes` as a root instead of the root of the zarr.json. I think that composes better because we don't have to redefine the Zarr core metadata in our schema.

I also added `schema_url` as a new property but removed it again because it caused issues with the ome-ngff-challenge. We should discuss whether to add `schema_url` to the OME-Zarr metadata. Personally, I am not convinced that this is necessary because it is trivial to look up the schema in this repository based on the `version` attribute. That makes `schema_url` somewhat redundant and verbose.
cc @joshmoore @d-v-b

joshmoore:
Thanks for giving that URL a try, @normanrz. I keep going back and forth so a few quick thoughts for now:
 - Certainly the impact on the challenge shouldn't be a lasting concern. Happy to look into workarounds if you can tell me what went wrong.
 - I agree that *if* the version is in this repo then everything can be managed.
 - So maybe it comes down to: do we want to centralize / restrict who can define a dev version? 
 - And if we are handling everything via the version string, do we want to define any semantics for dev-like versions as used in the challenge?

normanrz:
> * I agree that _if_ the version is in this repo then everything can be managed.

That will be the case, once we get this in.




> * So maybe it comes down to: do we want to centralize / restrict who can define a dev version?
> 
> * And if we are handling everything via the version string, do we want to define any semantics for dev-like versions as used in the challenge?

I feel like this is out-of-scope of this RFC/PR but should be covered in its own RFC.

joshmoore:
I'm not sure I agree. If we are going to add semantics, either in the version string (e.g. "versions can be URLs") or with an additional field (`schema_url`) later, then it will be a MUST understand, right?

A workaround that might prevent the need to decide the specifics now (which I understand your hesitant to try to do before 0.5) would be: for the examples above, either "if the version does not match `/\d+[.]\d+[.]\d+`" or "`if schema_url is not None`" then do not consider an official dataset.

normanrz:
RFC-2 does not add or change the semantics of versioning in OME-Zarr, apart from bumping it to 0.5.

What you seem to be proposing are new semantics for working with dev versions (which we sorta did in the challenge). Adding a `schema_url` field would be one way of dealing with that. Even if we decide that `schema_url` or other means of handling dev versions is useful, I am arguing that this should not be part of this PR and should not block 0.5 from being shipped. I think that would need its own RFC instead of being integrated into the adoption of Zarr v3.

joshmoore:
> I think that would need its own RFC instead of being integrated into the adoption of Zarr v3.

Alright, very fair. A few final points, largely in my defense:
 * The work on the challenge has certainly (unduly?) influenced me here. The issues of schema versions rear their head very quickly when doing anything other than the standard, single-file release.
 * Pulling the version up to the top certainly made this suggestion more *attainable* in this PR, but it's fine if we simply aren't there yet.
 * The previous suggestion with a URL-based version *would* have achieved it which also probably makes it seem so close (yet still too far). :wink:
 
Thanks for your patience with me, @normanrz.

joshmoore:
From a separate discussion with @normanrz, here are some lingering TODOs towards 0.5 post-merge:

- [ ] merging
- [ ] reviewing the status of latest/
- [ ] reverting anything that was introduced incorrectly
- [ ] moving latest/ to 0.5/
- [ ] creating a new “latest” or making it a redirect
- [ ] reviewing implementation support and test data
- [ ] announcement, adopt, applause 👏🏽 

The major issue will be communicating the state of the underlying libraries (Python, Java, Javascript), i.e., what should everyone's expectations be on adoption? This will not be as clear as an issue with, e.g., 0.4, but having *some* adoption will in turn help drive the stabilization of Zarr v3.

cc: @bogovicj @melissalinkert @manzt _et al._

about/index.md
normanrz, line 1: @joshmoore please review

.github/workflows/publish.yml
```diff
@@ -12,10 +12,10 @@ on:
 jobs:
   main:
     name: Build, Validate and Deploy
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v2
-      - uses: ome/spec-prod@main
+      - uses: normanrz/spec-prod@patch-1
```
normanrz, line 18:
```suggestion
      - uses: ome/spec-prod@main
```

conf.py
```diff
@@ -71,11 +80,14 @@ def bikeshed():
                 run_bikeshed = False
 
         if run_bikeshed:
-            subprocess.check_call(f"bikeshed  spec {index_file} {output_file}", shell=True)
+            subprocess.check_call(
+                f"bikeshed -f spec {index_file} {output_file}", shell=True
```
normanrz, line 84:
```suggestion
                f"bikeshed spec {index_file} {output_file}", shell=True
```

joshmoore:
See https://github.com/speced/bikeshed/issues/2946 for some of the concerns around bikeshed

joshmoore:
As someone of you may have noticed from the large number of commits pushed here, the spec build is currently broken due to an update in the build system (`bikeshed`). This is now breaking other work like #267. I'm in the process of copying the reference data (https://github.com/IDR/ome-ngff-samples/pull/22) and getting implementations other than vizarr and validator (like neuroglancer: https://github.com/google/neuroglancer/pull/670) released. I'm going to move forward with our TODO list (https://github.com/ome/ngff/pull/242#issuecomment-2405032109) and merge this but not undertake the announcement step. That way we can get the data in place and the implementations re-tested before announcing more widely.

@will-moore: please note the new behavior of `/latest/` as a built copy of `/0.5/` in case there is any specific login in the validator that depends on something there. It might be useful to be able to specify an explicit schema location in the validator (either via semi-hidden text box or another URL parameter) to be able to test before a full release.

0.5/index.bs
```diff
@@ -399,25 +395,32 @@ The "labels" group is not itself an image; it contains images. The pixels of the
...
 ```json
 {
-  "labels": [
-    "cell_space_segmentation"
-  ]
+  ...
+  "attributes": {
+    "ome": {
+      "version": "0.5",
+      "schema_url": "https://ngff.openmicroscopy.org/latest/schemas/ome_zarr.schema",
```
sbesson, line 407: Is this on purpose? The `schema_url` key is not defined in the specification and I assume a latest URL means the underlying schema will evolve through time.

normanrz: Not on purpose. This needs fixing.

sbesson: Happy to open a PR to remove it unless it's already part of a batch of changes

PR #252, opened 2024-07-12
Update Fractal link & description

jluethi:
I realized that the link for Fractal on the NGFF tools page was pointing to just the Fractal command line client. We now have a nice [Fractal splash page](https://fractal-analytics-platform.github.io/) that gives the overview of the framework. Would be great to update the link to that => see this PR :)

I also made some minor changes to the Fractal description to better reflect the current scope.


joshmoore:
Thanks, @jluethi! :heart:

PR #255, opened 2024-08-14
RFC-5: Transformations and Coordinate systems

bogovicj:


normanrz:
Thanks for your continued work on the transformations proposal and for writing it up as an RFC. 
OME-Zarr really needs expanded transformation capabilities and I like this proposal! Happy to endorse it.

One issue that I see is the use of absolute paths in the `path` attribute (i.e. starting with `/`). This seems somewhat ill-defined to me. The example [in the RFC](https://ngff--255.org.readthedocs.build/rfc/wip-tform/index.html#coordinatetransformations-metadata) shows that "store.zarr" is the root, which I assume would be the starting point for the absolute paths. However, any group in a Zarr hierarchy is a valid hierarchy on its own and there is no mandate for a `.zarr` suffix to denote a root. Therefore, I would advocate for restricting paths to relative paths, which would be relative to the folder that holds the `zarr.json`.

Also, I want to mention that there is interest in the community to create a "collections" proposal for OME-Zarr. I haven't gotten the time to write an RFC yet but I think that could integrate nicely with this transformations proposal and might provide solutions for the path issue as well.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+
+
+The order of the `"axes"` list matters and defines the index of each array dimension and coordinates for points in that
+coordinate system. For the above example, the `"x"` dimension is the first dimension. The "dimensionality" of a coordinate system
+is indicated by the length of its "axes" array. The "volume_micrometers" example coordinate system above is three dimensional (3D).
+
+The axes of a coordinate system (see below) give information about the types, units, and other properties of the coordinate
+system's dimensions. Axis `name`s may contain semantically meaningful information, but can be arbitrary. As a result, two
+coordinate systems that have identical axes in the same order may not be "the same" in the sense that measurements at the same
+point refer to different physical entities and therefore should not be analyzed jointly. Task that require images, annotations,
```
d-v-b, line 66:
```suggestion
point refer to different physical entities and therefore should not be analyzed jointly. Tasks that require images, annotations,
```

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+regions of interest, etc. SHOULD ensure that they are in the same coordinate system (same name, with identical axes) or can be
```
d-v-b, line 67:
```suggestion
regions of interest, etc., SHOULD ensure that they are in the same coordinate system (same name, with identical axes) or can be
```

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+transformed to the same coordinate system before doing analysis. See the example below.
+
+
+### "axes" metadata
+
+"axes" describes the dimensions of a coordinate systems. It is a list of dictionaries, where each dictionary describes a dimension (axis) and:
+- MUST contain the field "name" that gives the name for this dimension. The values MUST be unique across all "name" fields.
+- SHOULD contain the field "type". It SHOULD be one of "array", "space", "time", "channel", "coordinate", or "displacement" but MAY take other values for custom axis types that are not part of this specification yet.
```
d-v-b, line 75:
```suggestion
- SHOULD contain the field "type". It SHOULD be one of the strings "array", "space", "time", "channel", "coordinate", or "displacement" but MAY take other string values for custom axis types that are not part of this specification yet.
```

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+    - Units for "space" axes: 'angstrom', 'attometer', 'centimeter', 'decimeter', 'exameter', 'femtometer', 'foot', 'gigameter', 'hectometer', 'inch', 'kilometer', 'megameter', 'meter', 'micrometer', 'mile', 'millimeter', 'nanometer', 'parsec', 'petameter', 'picometer', 'terameter', 'yard', 'yoctometer', 'yottameter', 'zeptometer', 'zettameter'
+    - Units for "time" axes: 'attosecond', 'centisecond', 'day', 'decisecond', 'exasecond', 'femtosecond', 'gigasecond', 'hectosecond', 'hour', 'kilosecond', 'megasecond', 'microsecond', 'millisecond', 'minute', 'nanosecond', 'petasecond', 'picosecond', 'second', 'terasecond', 'yoctosecond', 'yottasecond', 'zeptosecond', 'zettasecond'
+- MAY contain the field "longName". The value MUST be a string, and can provide a longer name or description of an axis and its properties.
+
+If part of multiscales metadata, the length of "axes" MUST be equal to the number of dimensions of the arrays that contain the image data.
+
+Arrays are often thought of as containing discrete samples along the continuous variable. Axes representing space and time are
```
d-v-b, line 84:
```suggestion
Arrays are often used to store discrete samples from a continuous variable. Axes representing space and time are
```

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+usually continuous, meaning they can be indexed by real-valued (floating point) numbers whereas discrete axes may be indexed
+only by integers. Arrays are inherently discrete (see Array coordinate systems, below) , but values "in between" discrete
+coordinates can be retreived using an *interpolation* method.  If an axis is continuous (`"discrete" : false`), it indicates
```
d-v-b, line 87:
```suggestion
coordinates can be retrieved using an *interpolation* method.  If an axis is continuous (`"discrete" : false`), it indicates
```

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+that indexing with continuous values is meaningful, and that interpolation may be performed along that axis. Interpolation may
+be performed jointly across axes with the same `type`, and interpolation should not be performed for discrete axes or jointly
+across axes with differing `type`s.  Other non-continuous variables and axis types are also usually discrete, such as
+`channel`s, `coordinate`s, and `displacement`s.
```
d-v-b, line 91: I don't think we should say that interpolation _may_ or _may not_ be performed. Generally, we should keep the normative statements restricted to the content of the metadata, rather than what actions people / software might take. We can instead make a non-normative statement to the effect that, for an axis with `{"discrete": false}` , smoothing and resampling of data values along this axis is a well-defined operation, whereas the same is not true for an axis with `{"discrete": True}`

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+
+Note: The most common methods for interpolation are "nearest neighbor", "linear", "cubic", and "windowed sinc". Here, we refer
+to any method that obtains values at real valued coordinates using discrete samples as an "interpolator". As such, label images
+may be interpolated using "nearest neighbor" to obtain labels at points along the continuum.
+
+
+### Array coordinate systems
+
+Every array has a default coordinate system whose parameters need not be explicitly defined.  Its name is the path to the array
+in the container, its axes have `"type":"array"`, are unitless, and have default "name"s. The ith axis has `"name":"dim_i"`
```
d-v-b, line 101: Why make the default name require information about where the array is stored in the container? This means that accessing the default coordinate system requires accessing the storage layer for the array, which rules out anything that acts exclusively on metadata. 

Perhaps it would be simpler to reserve a string like "default" for the default coordinate system.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+
+The dimensionality of each array coordinate system equals the dimensionality of its corresponding zarr array.  The axis with
+name `"dim_i"` is the ith element of the `"axes"` list. The axes and their order align with the `shape`
+attribute in the zarr array attributes (in `.zarray`), and whose data depends on the byte order used to store
+chunks. As described in the [zarr array metadata](https://zarr.readthedocs.io/en/stable/spec/v2.html#arrays),
+the last dimension of an array in "C" order are stored contiguously on disk or in-memory when directly loaded. 
+
+
+The name and axes names MAY be customized by including a `arrayCoordinateSystem` field in
+the `.zattr` metadata of the array whose value is a coordinate system object. The length of
```
d-v-b, line 113:
```suggestion
the user-defined attributes of the array whose value is a coordinate system object. The length of
```

Avoid overfitting to zarr v2

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+in the discrete array, i.e., the area corresponding to nearest-neighbor (NN) interpolation of that sample.
+The center of a 2d pixel corresponding to the origin `(0,0)` in the discrete array is the origin of the continuous coordinate
+system `(0.0, 0.0)` (when the transformation is the identity). The continuous rectangle of the pixel is given by the
+half-open interval `[-0.5, 0.5) x [-0.5, 0.5)` (i.e., -0.5 is included, +0.5 is excluded). See chapter 4 and figure 4.1 of the ITK Software Guide.
+
+
+### "coordinateTransformations" metadata
+
+"coordinateTransformations" describe the mapping between two coordinate systems (defined by "axes").
+For example, to map an array's discrete coordinate system to its corresponding physical coordinates.
```
d-v-b, line 133:
```suggestion
For example, coordinate transforms can be used to map an array's discrete coordinate system to its corresponding physical coordinates.
```

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,681 @@
...
+
+
+Conforming readers:
+- MUST parse `identity`, `scale`, `translation` transformations;
+- SHOULD parse `mapAxis`, `affine` transformations;
+- SHOULD be able to apply transformations to points;
+- SHOULD be able to apply transformations to images;
+
+Coordinate transformations from array to physical coordinates MUST be stored in multiscales,
+and MUST be duplicated in the atrributes of the zarr array. Transformations between different images MUST be stored in the
```
d-v-b, line 200:
```suggestion
and MUST be duplicated in the attributes of the zarr array. Transformations between different images MUST be stored in the
```

rfc/wip-tform/index.md
bogovicj: Because the coordinate systems of different arrays need to be different by default.

rfc/wip-tform/index.md
bogovicj: https://github.com/bogovicj/ngff/commit/dadff02cde9eb5dd76aa975702d9409a2d190c6e

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+    <td> one of: <br>`"translation":List[number]`, <br>`"path":str`
+    <td>translation vector, stored either as a list of numbers (`"translation"`) or as binary data at a location
+    in this container (`path`).
+  <tr><th>`scale`
+    <td> one of: <br>`"scale":List[number]`, <br>`"path":str`
+    <td>scale vector, stored either as a list of numbers (`scale`) or as binary data at a location in this
+    container (`path`).
+  <tr><th>`affine`
+    <td> one of: <br>`"affine":List[number]`, <br>`"path":str`
+    <td>affine transformation matrix stored as a flat array stored either with json uing the affine field
```
LucaMarconato, line 160: Leftover from the old specs, now the array/list is not flat but 2D/nested

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+    or as binary data at a location in this container (path). If both are present, the binary values at path should be used.
+  <tr><th>`rotation`
+    <td> one of: <br>`"rotation":List[number]`, <br>`"path":str`
+    <td>rotation transformation matrix stored as an array stored either
+        with json or as binary data at a location in this container (path).
+        If both are present, the binary parameters at path are used.
+  <tr><th>`sequence`
+    <td> `"transformations":List[Transformation]`
+    <td>A sequence of transformations, Applying the sequence applies the composition of all transforms in the list, in order.
```
LucaMarconato, line 169: I'd clarify that if the list if `[f0, f1, f2]` the order is `f2 \circ \f1 \circ f0 = f2(f1(f0()))`

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+  <tr><th>`displacements`
+    <td>`"path":str`<br>`"interpolation":str`
+    <td>Displacement field transformation located at (path).
+  <tr><th>`coordinates`
+    <td>`"path":str`<br>`"interpolation":str`
+    <td>Coordinate field transformation located at (path).
+  <tr><th>`inverseOf`
+    <td>`"transform":Transform`
+    <td>The inverse of a transformation. Useful if a transform is not closed-form invertible.  See Forward and inverse for details and examples.
```
LucaMarconato, line 178: typo, double space

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+
+This RFC is currently in stage D2.
+
+|  Role  | Name      | GitHub Handle | Institution | Date       | Status                                 |
+|--------|-----------|---------------|-------------|------------|----------------------------------------|
+| Author | John Bogovic    | @bogovicj       | HHMI Janelia   | 2024-07-30  | Author; Implemented               |
+| Author | Davis Bennett   | @d-v-b          |                | 2024-07-30  | Author; Implemented validation    |                                 
+| Author | Luca Marconato  | @LucaMarconato  | EMBL           | 2024-07-30  | Author; Implemented               |
+| Author | Matt McCormick   | @thewtex       | ITK            | 2024-07-30  | Author; Implemented               |
+| Author | Stephan Saalfeld | @axtimwalde    | HHMI JAnelia   | 2024-07-30  | Author; Implemented (with JB)     |
```
LucaMarconato, line 15: Typo JAnelia

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+</pre>
+
+### Additional details
+
+Most coordinate transformations MUST specify their input and output coordinate systems using `input` and `output` with a string value
+corresponding to the name of a coordinate system. The coordinate system's name may be the path to an array, and therefore may
+not appear in the list of coordinate systems.
+
+Exceptions are if the the coordinate transformation appears in the `transformations` list of a `sequence` or is the
+`transformation` of an `inverseOf` transformation. In these two cases input and output SHOULD be omitted (see below for
```
LucaMarconato, line 235: I'd replace "SHOULD be omitted" with something with the meaning of "could, in some cases, be omitted". Because sometimes they are required.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+    }
+}
+```
+
+Implementations SHOULD be able to compute and apply the inverse of some coordinate transformations when they
+are computable in closed-form (as the [Transformation types](#transformation-types) section below indicates). If an
+operation is requested that requires the inverse of a transformation that can not be inverted in closed-form,
+implementations MAY estimate an inverse, or MAY output a warning that the requested operation is unsupported.
+
+#### Matrix transformations
```
LucaMarconato, line 275: I'd clarify that for rotations, we have a standard matrix vector multiplication, but for affine, we need to add a row of [0, 0, ..., 0, 1] and add a 1 to the bottom vector before doing the matrix vector multiplication, i.e. we need to rewrite the matrix and vector in homogenous coordinates.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+`"shape":[3,4]` has 3 rows and 4 columns). When stored as a 2D json array, the inner array contains rows (e.g. `[[1,2,3],
+[4,5,6]]` has 2 rows and 3 columns).
+
+
+### Transformation types
+
+Input and output dimensionality may be determined by the value of the "input" and "output" fields, respectively. If the value
+of "input" is an array, it's length gives the input dimension, otherwise the length of "axes" for the coordinate
+system with the name of the "input" value gives the input dimension.  If the value of "input" is an array, it's
+length gives the input imension, otherwise it is given by the length of "axes" for the coordinate system with
```
LucaMarconato, line 290: typo

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+the name of the "input".  If the value of "output" is an array, it's length gives the output dimension,
```
LucaMarconato, line 291: typo

LucaMarconato: it's -> its

rfc/wip-tform/index.md
LucaMarconato: imension -> dimension

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+  <dd>  The path to a zarr-array containing the scale parameters.
+	The array at this path MUST be 1D, and its length MUST be `N`.</dd>
+  <dt><strong>scale</strong></dt>
+  <dd> 	The scale parameters stored as a JSON list of numbers. The list MUST have length `N`.</dd>
+</dl>
+
+
+#### <a name="affine">affine</a>
+
+`affine`s are [matrix transformations](#matrix-transformations) from N-dimensional inputs to M-dimensional outputs are represented at `(N)x(M+1)`
```
LucaMarconato, line 344: I think there is an error: if N are the input and M are the output, then the matrix should have M rows and N + 1 columns, hence it should be (M)x(N+1).

LucaMarconato: If (N)x(M+1) is changed to (M)x(N+1) then it needs to be changed also in the text below.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+matrices in homogeneous coordinates. This transformation type may be (but is not necessarily) invertible when `N` equals `M`.
+The matrix MUST be stored as a 2D array either as json or in a zarr array.
+
+<dl>
+  <dt><strong>path</strong></dt>
+  <dd>  The path to a zarr-array containing the affine parameters.
+	The array at this path MUST be 2D whose shape MUST be `N x (M+1)`.</dd>
```
LucaMarconato, line 351: should be changed here.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+  <dt><strong>affine</strong></dt>
+  <dd> 	The affine parameters stored in JSON. The matrix MUST be stored as 2D nested array where the outer array MUST be length
+  `N` and the inner arrays MUST be length `M+1`.</dd>
```
LucaMarconato, line 354: and here.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+    when rendering the transformed moving image. Results such as this may be enclosed in an `inverseOf`
+    transformation. This enables the "outer" coordinate transformation to specify the moving image coordinates
+    as `input` and fixed image coordinates as `output`, a choice that many users and developers find intuitive.
+</div>
+
+
+
+#### <a name="sequence">sequence</a>
+
+A `sequence` transformation consists of an ordered array of coordinate transformations, and is invertible if and only if every
```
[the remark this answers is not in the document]
LucaMarconato: I'd rewrite the text to the following:

A `sequence` transformation consists of an ordered array of coordinate transformations. If all the coordinate transformations of the sequence are invertible, then the sequence transformation is also invertible. The opposite is not always true, but we add the constrain that all the transformations need to be invertible, because implementations may choose to apply a sequence transformation by applying the transformations one at the time.

Asking for feedback @d-v-b wdyt? Not adding this constraint would make implementations more complex because we can't apply a recursive logic to invert a sequence transformation.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+- The `input` and `output` fields MAY be omitted for the following transformation types: 
+    - `identity`, `scale`, `translation`, `rotation`, `affine`, `displacements`, `coordinates`
+- The `input` and `output` fields MAY be omitted for `inverseOf` transformations if those fields may be omitted for the
+    transformation it wraps
+- The `input` and `output` fields MAY be omitted for `bijection` transformations if the fields may be omitted for 
+    both its `forward` and `inverse` transformations
+- The `input` and `output` fields MAY be omitted for `sequence` transformations if the fields may be omitted for
+    all transformations in the sequence after flattening the nested sequence lists.
+- The `input` and `output` fields MUST be included for transformations of type: `mapAxis`, and `byDimension` (see the note
+    below), and under all other conditions.
```
LucaMarconato, line 411: (see the note below), I think this is a leftover text (had I reported it before in the main PR?) and should be removed.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+
+
```
LucaMarconato, line 413: Not clear to me. Also I'd use list instead of array since array is used to represent the component of a single transformation.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+            <li><code>linear</code> (default)</li>
+            <li><code>nearest</code></li>
+            <li><code>cubic</code></li>
+        </ul></dd>
+</dl>
+
+For both `coordinates` and `displacements`, the array data at referred to by `path` MUST define space and coordinate transform metadata:
+
+* Every axis name in the `coordinateTransform`'s `input` MUST appear in the coordinate system
+* The array dimension corresponding to the `coordinate` or `displacement` axis MUST have length equal to the dimensionality of the `coordinateTransform` `output`
```
LucaMarconato, line 457: Not sure, is it `output` or `output` + 1?

LucaMarconato: I didn't understand this point.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+* SHOULD have a `name` identical to the `name` of the corresponding `coordinateTransform`.
+
+For `coordinates`:
+
+* coordinateSystem metadata MUST have exactly one axis with `"type" : "coordinate"`
```
LucaMarconato, line 462: coordinateSystem -> `coordinateSystem`

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+* If a `coordinates`' input space has dimensionality `N`, then the array data at `path` MUST have dimensionality equal to `(N + 1)`.
+* the shape of the array along the "coordinate" axis must be exactly `N`
+
+For `displacements`:
+
+* coordinateSystem metadata MUST have exactly one axis with `"type" : "displacement"`
```
LucaMarconato, line 468: coordinateSystem -> `coordinateSystem`

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+For both `coordinates` and `displacements`, the array data at referred to by `path` MUST define space and coordinate transform metadata:
+
+* Every axis name in the `coordinateTransform`'s `input` MUST appear in the coordinate system
+* The array dimension corresponding to the `coordinate` or `displacement` axis MUST have length equal to the dimensionality of the `coordinateTransform` `output`
+* SHOULD have a `name` identical to the `name` of the corresponding `coordinateTransform`.
+
+For `coordinates`:
+
+* coordinateSystem metadata MUST have exactly one axis with `"type" : "coordinate"`
+* If a `coordinates`' input space has dimensionality `N`, then the array data at `path` MUST have dimensionality equal to `(N + 1)`.
```
LucaMarconato, line 463: typo: extra `'` to remove

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+* the shape of the array along the "coordinate" axis must be exactly `N`
+
+For `displacements`:
+
+* coordinateSystem metadata MUST have exactly one axis with `"type" : "displacement"`
+* If a `displacements`' input space has dimensionality `N`, then the array data at `path` MUST have dimensionality equal to `(N + 1)`.
```
LucaMarconato, line 469: typo: extra `'` to remove

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+
+* Every axis name in the `coordinateTransform`'s `input` MUST appear in the coordinate system
+* The array dimension corresponding to the `coordinate` or `displacement` axis MUST have length equal to the dimensionality of the `coordinateTransform` `output`
+* SHOULD have a `name` identical to the `name` of the corresponding `coordinateTransform`.
+
+For `coordinates`:
+
+* coordinateSystem metadata MUST have exactly one axis with `"type" : "coordinate"`
+* If a `coordinates`' input space has dimensionality `N`, then the array data at `path` MUST have dimensionality equal to `(N + 1)`.
+* the shape of the array along the "coordinate" axis must be exactly `N`
```
LucaMarconato, line 464: I don't understand what this means: * the shape of the array along the "coordinate" axis must be exactly `N`

LucaMarconato: Ah got it, I would maybe rephrase a bit, maybe say the "length of the array for the dimension of the `coordinate` axis". "shape` makes me think of `array.shape`, while here is `array.shape[-1] == n`.

rfc/wip-tform/index.md
LucaMarconato, line 463: isn't it the other way around? for `displacements` , the dimensionality equal to `(N + 1)`, but for coordinates, `N`.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+
+For `displacements`:
+
+* coordinateSystem metadata MUST have exactly one axis with `"type" : "displacement"`
+* If a `displacements`' input space has dimensionality `N`, then the array data at `path` MUST have dimensionality equal to `(N + 1)`.
+* the shape of the array along the "displacement" axis must be exactly `N`
```
LucaMarconato, line 470: Commented above, I think here is `N + 1`, while for coordinates is `N`.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+* `input` and `output` MUST have an equal number of dimensions.
```
LucaMarconato, line 471: Commented above, I think there `output` dimensions = `input` dimensions + 1, while for coordinates we have that `input` and `output` MUST have an equal number of dimensions.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+
+
+#### byDimension {#trafo-byDimension}
+
+`byDimension` transformations build a high dimensional transformation using lower dimensional transformations
+on subsets of dimensions.
+
+<dl>
+  <dt><strong>transformations</strong></dt>
+  <dd>  A list of transformations, each of which applies to a (non-strict) subset of input and output dimensions (axes).
```
LucaMarconato, line 482: I think that maybe these conditions leave some edge cases open:
- the byDimension transformation could have a sub transformation whose output is not in the output axes of the byDimension output. Probably we don't want this.
- there could be multiple subtransformations that share some axes in their input, maybe it's ok, but not sure if we want this.
- there could be some input axes of the parent object that have no corresponding input axis in any of the sub transformations. I think that this is acceptable, but I wanted to highlight it here.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+}
+```
+
+would change to:
+
+```
+{
+    "type": "scale",
+    "parameters": {
+        "scale": [2, 3, 4],
```
LucaMarconato, line 551: typo: one extra number.

rfc/wip-tform/index.md
```diff
@@ -0,0 +1,679 @@
...
+```
+
+In discussions, some authors preferred the latter because it will make the "top-level" keys for transformation
+objects all identical, which could make serialization / validation simpler. One downside is that this change
+is breaking for the existing `scale` and `translation` transformations
+
+```
+{
+    "type": "scale",
+    "scale": [2, 3, 4],
```
LucaMarconato, line 539: typo: one extra number.

LucaMarconato:
Thanks @bogovicj for opening the draft for the RFC, I have added my review. The points are mostly minor and I have only two more small comments, here below. After the above is resolved and the comments are addressed, for me we are good to go to officially submit the RFC.

1. I find this example of multiscale transformations vs "standard" transformations, very useful: https://github.com/bogovicj/ngff/blob/coord-transforms/latest/examples/multiscales_strict/multiscales_example_relative.json. I would stress this example more in the specs text.
2. Some things are not clear to me from the example above:
	2.1 are we allowed to use just `scale`, `scale`+`translation` or any transformation for the multiscale images? I am not sure what's the best here
	2.2 related to the point above is this issue of "what is a multiscale image and what not" https://github.com/ome/ngff/issues/260. I opened it outside the context of this PR, but I think it is still useful to decide how to proceed.
	
Regarding 2.1, I think that we should allow for `scale` + `translation` and not general transformations. ult

rfc/wip-tform/index.md
bogovicj: that apostrophe was actually meant to be a possessive  coordinates' ,
but it is confusing so I'll rephrase instead.

rfc/wip-tform/index.md
bogovicj: (same as above)

rfc/wip-tform/index.md
bogovicj: Thanks, I made this clarification in the affine subsection, mostly through the examples, which I'm omitting from the rfc for brevity

rfc/wip-tform/index.md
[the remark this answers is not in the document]
bogovicj: Thanks, I clarified the text a little (see below)

### the new text
> A `sequence` transformation consists of an ordered array of coordinate transformations, and is invertible if every coordinate transform in the array is invertible (though could be invertible in other cases as well). 

compare to
####  the old text
> A sequence transformation consists of an ordered array of coordinate transformations, and is invertible if and only if every coordinate transform in the array is invertible.

rfc/wip-tform/index.md
bogovicj: Yea, I'm referring to the shape field in the zarr's json (not python/numpy array shape). 
I could use "length" if it is clear enough that it refers to one entry of the shape...

rfc/wip-tform/index.md
bogovicj: This  property is true for both `coordinates` and `displacements`.  I will try to write some more text to clarify.

rfc/wip-tform/index.md
bogovicj: > the byDimension transformation could have a sub transformation whose output is not in the output axes of the byDimension
> output. Probably we don't want this.

While I wouldn't recommend that, I don't think this is a problem - it seems unambiguous how to compute the transformation. e.g. 

```
{
  "coordinateSystems": [
    {
      "name": "in",
      "axes": [
        { "name": "j", "type": "array" },
        { "name": "i", "type": "array" }
      ]
    },
    {
      "name": "out",
      "axes": [
        { "name": "y", "type": "array" },
        { "name": "x", "type": "array" }
      ]
    }
  ]
    {   
      "type": "byDimension",
      "input": "in",
      "output": "out",
      "transformations": [
        {
          "type": "translation",
          "translation": [ 1, 3 ],
          "input": [ "i", "j" ],
          "output": [ "x", "IGNORED" ]
        },
        {
          "type": "scale",
          "scale": [ 2, 3 ],
          "input": [ "i", "j" ],
          "output": [ "IGNORED". "y" ]
        }
      ]

    }
  ]
}
```

> there could be multiple subtransformations that share some axes in their input, maybe it's ok, but not sure if we want this.

Agreed that I would not recommend it, but don't think it is a problem. The above example applies to this as well.


> there could be some input axes of the parent object that have no corresponding input axis in any of the sub transformations. I think that this is acceptable, but I wanted to highlight it here.

Agreed.

rfc/wip-tform/index.md
bogovicj: Described this a bit more that I hope clarifies things.

bogovicj:
Thanks  @LucaMarconato and @d-v-b ,

I address most of your comments.  While there are few left, I'll get to them soon, and they're more minor.  So I'm going to take this PR off of draft status.

joshmoore:
* [x] Assigned `RFC-5`
* [x] Added navigation links and changelog
* [x] Corrected some minor formatting issues

Merging and sending for review once the build is green. If you would like to be added as an **`Endorser`**, please comment below or as always, PRs welcome.

Thanks to everyone involved! :tada:

tischi:
I endorse this RFC

PR #258, opened 2024-08-30
RFC-1: author response

joshmoore:
Just *barely* in time to have met my summer goal of replying to the reviews of [RFC-1](https://ngff.openmicroscopy.org/rfc/1), I'm opening this update to the RFC process proposal. There is likely some need for clarifications before the changes are merged and sent out for re-review (R1). I would ask interested members of the community two favors:

### A. Endorsements (easier)

Since the first version of RFC-1 in #222 did not exist in order to request endorsements, I did not request them. Now that RFC-2 and RFC-3 along with #253 and #254 (a.k.a. RFC-4 and RFC-5) have successfully demonstrated the joy of endorsing RFCs, **I would ask anyone who supports RFC-1 to add an emoji below. (Instructions provided)**

### B. Clarifications (bit more work)

Additionally, if you would like to give a careful read of the document and identify anything that is confusing, illogical, or otherwise to be avoided, I'd appreciate the additional eyes. You are additionally welcome to submit an additional, long-form Comment. (Examples are in the built webpages: https://ngff--258.org.readthedocs.build/rfc/1/comments/)

Reviewers are of course welcome to be involved in the clarifications, but need not do so. 

----

⚠️ Emojis on this PR will be considered for endorsements:
 - 👍🏽 ❤️ 🚀 😄 🎉 -- you will be listed on the RFC as an endorser.
 - 👎🏽 ☹️ -- I will contact you regarding how you would like to express your objection.
 - 👀: no special meaning

PR #261, opened 2024-09-11
RFC-2 Updates

joshmoore:
* [x] Applied styling and new directory structure from  #258 
 * [x] Added @jbms' "accept" review
 * [x] Added @yarikoptic' "accept" review
 * [x] change state

(cc: @normanrz)


joshmoore:
Leaving this open with the hope to get @yarikoptic's re-review in.

sbesson:
If this is changing the URLs, the content of review 1b or RFC-2 might need to be slightly amended as it is linking to specific anchor of the review 1 via hyperlink and these would be broken by the renaming

joshmoore:
Ah, thanks for the heads up, @sbesson. I'll review. My experience with #258 was now that everything is in the markdown TOC we should be able to get good link checking if we omit the `.html` ending. I'll review all the files to this end.

PR #265, opened 2024-10-11
Janelia response to RFC-1 review (2b)

virginiascarlett:
Hello,

It's a busy time of year so not all of the reviewers were able to provide a response. Those who were available have provided their response here.

Thank you very much for your continued efforts on this important project.


joshmoore:
And thank you for your support. I take

> the reviewers hope that their concerns will be kept in mind and revisited if / when these concerns become practical issues.

very much onboard. Looking forward to future conversations!

PR #270, opened 2024-11-21
Clarify: other axis type values should be strings

imagejan:
For axis types that are _not_ `space`, `time` or `channel`, the JSON schema only allows string values:

https://github.com/ome/ngff/blob/e1dc738f5714bb872bd5c623931729bfe0a81ad6/0.4/schemas/image.schema#L155-L160

The language in the spec however didn't specify this so far. Let's explicitly specify that other values must be strings.


joshmoore:
Thanks, @imagejan! Because of the current build failures, I've pulled these changes into #242 (and applied to the upcoming 0.5 version)

Issue #275, opened 2024-11-22
Add a complete and verbose changelog


dstansby:
As far as I can see, there is no complete changelog between versions of the OME-NGFF spec. The closest I can find is https://ngff.openmicroscopy.org/0.5/index.html#history, which only has a very short sentence on what changed in each version.

As a concrete example, in v0.5 axis types were narrowed so they have to be strings. But I have absolutely no easy way of knowing that this changed happened, making it very hard for me to make sure that I have finished updating an implementation of the v0.5 spec.

I propose that a complete, verbose changelog is added to the specification docs that explain **every** change that happens to the spec between versions. This would make it easy for users to understand what new features have been added, and easy for implementers to update their implementations with new versions of the spec.

Issue #276, opened 2024-11-22
Which copy of the SPEC should changes be proposed against?


dstansby:
I want to contribute a minor clarification to the spec, but it's not clear what the process is to do that. In particular, which file in this GitHub repo should I propose a change against?

I suggest that a new "Contributing" (or similar) section is added at the top level to https://ngff.openmicroscopy.org/, which explains how to 1) propose minor changes/clarifications 2) how to propose major additions or changes (ie. explain or link to the RFC process).

joshmoore:
Before @normanrz's 0.5 changes, your changes would have gone into `/latest/`. https://github.com/ome/ngff/issues/271 captured "well, now we need a new mechanism". If you have suggestions on doing that, feel free. Otherwise, you can keep capturing proposals against /0.5/ as you did in gh-277 etc. while this issue (#276+#271) gets worked out.

dstansby:
i've put a proposal at #281 - do you have context on why `latest` was removed (cc @normanrz)? It seems quite logical to me to **always** have a `latest` folder that changes are made against, and that is just copied and 'frozen' in a folder named after each version when a release is made.

normanrz:
We were thinking of migrating to a branch-based scheme of managing versions. Instead of `0.4`, `0.5`, `latest` folders, there would be `0.4`, `0.5`, `latest` branches But we didn't habe the time to fully build it out because we were focussed on getting `0.5` done. Putting a symlink in for `latest` -> `0.5` solved the need to keep existing links functioning.

joshmoore:
See #282 for a first POC

joshmoore:
With #308 merged, I've prepared https://github.com/ome/ngff/compare/v0.5...joshmoore:ngff:dev?expand=1 as a proposal or at least a example of what it would look like to base new spec changes on one of the extracted branches, here `v0.5`:

https://github.com/ome/ngff/compare/v0.5...joshmoore:ngff:dev?expand=1

will-moore:
That compare has "Showing 301 changed files with 3,442 additions and 20,132 deletions."
Hopefully that could be cleaned-up a bit?

joshmoore:
Hmmm.... locally it doesn't look quite as bad:

```
git diff --stat origin/v0.5 origin/dev
 examples/coordSystems/.config.json                            |    3 +
 examples/coordSystems/arrayCoordSys.json                      |   10 ++
 examples/multiscales_strict/multiscales_example_relative.json |   76 ++++++++++++
[...]

PR #277, opened 2024-11-22
Clarify that zarr layout given in spec is an example

dstansby:
This is a minor language clarification to switch from saying that this example is the "expected" layout to saying that it is a valid example.

I have made this change because I think "expected" could be interpreted as "this is what the spec says your layout should be", but I think the intent here was actually "here's a useful example of what a typical layout would be". 

Note that I also added a new changelog section to the spec, because changes like this should get a changelog entry (see https://github.com/ome/ngff/issues/275). This is just so I don't forget to include a changelog somewhere, and not an assumption on what the final location/format of that changelog will be, which can be discussed in https://github.com/ome/ngff/issues/275.

### DIFF
--- a/0.5/index.bs
+++ b/0.5/index.bs
@@ -652,6 +652,15 @@ Version History {#history}
   </tr>
 </table>
 
+Changelog {#changelog}
+======================
+
+Version 0.6
+-----------
+
+- Clarified that the example file layout given in section 2.1 is
+  an example of a valid layout, and not *the* expected layout.
+
 
 <pre class="biblio">
 {
--- a/0.5/index.bs
+++ b/0.5/index.bs
@@ -65,7 +65,7 @@ like S3 or GCS.
 Images {#image-layout}
 ----------------------
 
-The following layout describes the expected Zarr hierarchy for images with
+The following layout shows an example valid Zarr hierarchy for images with
 multiple levels of resolutions and optionally associated labels.
 Note that the number of dimensions is variable between 2 and 5 and that axis names are arbitrary, see [[#multiscale-md]] for details.
 


0.5/index.bs
```diff
@@ -652,6 +652,15 @@ Version History {#history}
   </tr>
 </table>
 
+Changelog {#changelog}
+======================
+
+Version 0.6
```
joshmoore, line 658: I understand this to be a collection of more fine-grained changes between versions. If so, can we move it to the top-level? This could lead to confusion as to the state of "0.6" if in the 0.5 spec which could lead to confusion.

joshmoore (changes requested):
Thanks for the clarification, @dstansby. Think we need to work on explaining what's changed where though.

0.5/index.bs
dstansby: So in my head 0.5 is released and therefore fixed, because it is not versioned at a more granular level (e.g., 0.5.0, 0.5.1, 0.5.2). See my reasoning for why I think 0.x should be fixed on release in https://github.com/ome/ngff/pull/278#issuecomment-2493728701.

> If so, can we move it to the top-level?

I don't see the harm (and I think it makes it easier to understand) in calling out which version a change happened in (and coversely, what list of changes happened in a given version)

This all ties back to https://github.com/ome/ngff/issues/276 it seems - shouldn't there be an "active development" version of the spec, which changes are made against, which will become the version after the most recent versioned release?

joshmoore: Hmm.... to be clear: did you edit in `/latest/` or in `/0.5/`? @normanrz's PR may have moved your target.

dstansby: Latest is currently symlinked to /0.5/: https://github.com/ome/ngff/commit/72c443041f49c0ab4c414d5d40f196143c612fd1, so I don't think there is a latest copy at the moment?

joshmoore: Nope. That was what I was trying to figure out. If you had made your change on `/latest/` then @normanrz's change automagically integrated it.

PR #278, opened 2024-11-22
Clarify that axes MUST be a list of dictionaries

dstansby:
I think the intent with "It is a list of dicionaries" was a tight constraint, so I've replaced it with "MUST".

### DIFF
--- a/0.5/index.bs
+++ b/0.5/index.bs
@@ -179,7 +179,7 @@ The OME-Zarr Metadata version MUST be consistent within a hierarchy.
 "axes" metadata {#axes-md}
 --------------------------
 
-"axes" describes the dimensions of a physical coordinate space. It is a list of dictionaries, where each dictionary describes a dimension (axis) and:
+"axes" describes the dimensions of a physical coordinate space. It MUST be list of dictionaries, where each dictionary describes a dimension (axis) and:
 - MUST contain the field "name" that gives the name for this dimension. The values MUST be unique across all "name" fields.
 - SHOULD contain the field "type". It SHOULD be one of "space", "time" or "channel", but MAY take other string values for custom axis types that are not part of this specification yet.
 - SHOULD contain the field "unit" to specify the physical unit of this dimension. The value SHOULD be one of the following strings, which are valid units according to UDUNITS-2.


0.5/index.bs
```diff
@@ -179,7 +179,7 @@ The OME-Zarr Metadata version MUST be consistent within a hierarchy.
 "axes" metadata {#axes-md}
 --------------------------
 
-"axes" describes the dimensions of a physical coordinate space. It is a list of dictionaries, where each dictionary describes a dimension (axis) and:
+"axes" describes the dimensions of a physical coordinate space. It MUST be list of dictionaries, where each dictionary describes a dimension (axis) and:
```
joshmoore, line 182:
```suggestion
"axes" describes the dimensions of a physical coordinate space. It MUST be a list of dictionaries, where each dictionary describes a dimension (axis) and:
```

joshmoore (commented):
Thanks, @dstansby.

joshmoore:
A few additional thoughts:
 - This and maybe a number of your other changes might also be good in 0.4. If you are interested, you might consider applying to both.
 - Any objections to batching them somewhat?
 - As we rightly add more MUSTs, is it worth having them all at the same level:

``` 
A foo is a $definition that:
 - MUST
 - MUST
 - SHOULD
 - SHOULD
```
etc.

dstansby:
> This and maybe a number of your other changes might also be good in 0.4

I'm personally pretty strongly 👎 this, because a) updates to 0.x are not versioned and b) it becomes more work (even if updates are versioned) for implementations to target several moving versions of the spec, instead of one moving one (the latest version). So actually I think this change should go into 0.6, not in 0.5.

> Any objections to batching them somewhat?

Nope, this is a great idea. I'll switch this to draft, go through and find similar updates, and then 

> As we rightly add more MUSTs, is it worth having them all at the same level:

👍

joshmoore:
> So actually I think this change should go into 0.6, not in 0.5.

:+1:

> Nope, this is a great idea. I'll switch this to draft, go through and find similar updates, and then

:+1:

Thanks.

Issue #271, opened 2024-11-21
spec diffs can be hard to read 


joshmoore:
In https://github.com/ome/ngff/issues/249#issuecomment-2165869299, @d-v-b pointed out that the current strategy can lead to situations where there's no clear diff for a spec change (especially related to dev versions). For 0.5, @normanrz and I are going to proceed with the current 0.5/ directory to reduce churn, but for the release of 0.6 it might make sense to collapse the versions down. A priority will be maintaining the existing URLs, though a redirect would likely be acceptable.

joshmoore:
Depending on the solution, also:
 - evaluate dropping `latest`
 - or symlinking (might require a more substantial web server)
 - dropping bikeshed

normanrz:
I just want to add that both the docs and the specs should have a favicon. Right now they have the default RTD and W3C icons.

normanrz:
I think the best way of organizing versions would be through git branches/tags. That adds a bit more complexity to the rendering. I don't know what to do about bikeshed.

joshmoore:
A fairly easy mechanism would be to retroactively attach them as builds which get downloaded from GitHub for rendering.

PR #281, opened 2024-11-29
Add instructions on how to contribute a minor change

dstansby:
This:

1. Creates a latest copy of the specification again, which can be used to propose changes against.
2. Adds a new contributing section, explaining that minor changes or clarifications should be opened as a pull request directly to the `latest` folder. This page should also explain how to open an RFC (and broadly the line between not-an-RFC and RFC change), but that's out of scope for this PR.
3. Adds a changelog to the bottom of the spec, so we have a tracked and easy to read list of changes to the spec over time.

I'm hoping this makes it clearer and easier to contribute changes to the specification?

Fixes https://github.com/ome/ngff/issues/271, fixes https://github.com/ome/ngff/issues/276


joshmoore:
Big :+1: for the additional page. Thanks, @dstansby. Can you take a look at the strategy in #282 to see if you think it could also be (made) new contributor friendly?

dstansby:
I think it's worth rolling this in with https://github.com/ome/ngff/pull/282, adding a contributing section in this PR. It can be copied from this PR to start with, but I think explaining how to work with the proposed layout at https://github.com/ome/ngff/pull/282 will help evaluate it (instead of building some versioning machinery in a documentation free vacuum).

I'll take a look and give some feedback over in https://github.com/ome/ngff/pull/282.

PR #284, opened 2024-12-09
ngff zarr 0.5

thewtex:
- tools: note that File Reading/Writing tools also support conversion
- rfc-2: note ngff-zarr implementation support

### DIFF
--- a/rfc/2/index.md
+++ b/rfc/2/index.md
@@ -287,6 +287,18 @@ See [Background](#background) for a list of v3-capable Zarr libraries.
 
 Support for the OME-Zarr 0.5 metadata is under development in [ome-zarr-py](https://github.com/ome/ome-zarr-py/pull/383/files) and other implementations.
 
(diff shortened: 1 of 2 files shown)


thewtex:
Example validation: https://ome.github.io/ome-ngff-validator/?source=https://bafybeihtjs5tjn26jfcbhe65wrewgwvhlbeyqe2qztjiqjq5yz7ycmcclq.ipfs.w3s.link/

joshmoore (approved):
> Example validation ...

Lovely! Makes my heart happy. (Now we just need an `@ngffbot` to pick up URLs and send them to the validator :smile:)

I'll hold off on merging for any comments from @normanrz as the author. Personally I think this adds weight to the imminent 0.5 and so makes sense. Long-term, I think the strategy of needing or wanting to update the RFCs page with tools may become burdensome. (If so, I'd almost say we automate it via a little JSON in the repository that lets us autogenerate the tools/ page and the RFC support sections.)

PR #285, opened 2024-12-12
RFC-6: Multiscale

normanrz:
This RFC proposes to change the `multiscales` array into a single `multiscale` object.

co-authored by @dstansby

### DIFF
--- a/rfc/6/index.md
+++ b/rfc/6/index.md
@@ -48,6 +48,12 @@ This proposal is very early. Status: D1
     - Australian National University
     - 2024-12-17
     -
+*   - Endorser
+    - Joel Lüthi
+    - jluethi
+    - BioVisionCenter, University of Zurich
+    - 2024-12-18
+    -
 ```
 
 ## Overview
--- a/rfc/6/index.md
+++ b/rfc/6/index.md
@@ -47,21 +47,6 @@ The current spec seems to acknowledge that this is to be expected to some degree
 
 This RFC aims to codify what already seems to be common practice by moving from a multiscales array to a single multisclaes object. This will reduce complexity for implementations.
 
-<!--
-The next section is the "Background" section. This section should be at least
-two paragraphs and can take up to a whole page in some cases. The \*\*guiding goal
-of the background section\*\* is: as a newcomer to this project (new employee, team
-transfer), can I read the background section and follow any links to get the
-full context of why this change is necessary? 
-
-If you can't show a random engineer the background section and have them
-acquire nearly full context on the necessity for the RFC, then the background
-section is not full enough. To help achieve this, link to prior RFCs,
-discussions, and more here as necessary to provide context so you don't have to
-simply repeat yourself.
-
--->
-
 ## Proposal
 
 The OME-Zarr metadata in a `zarr.json` file of a multiscale MUST contain a single `multiscale` object. This replaces the current `multiscales` array.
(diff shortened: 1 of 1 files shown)


normanrz:
If you would like to endorse this RFC, please leave a :+1: reaction or comment.

ziw-liu:
One (hypothetical) use case for an array of multiscales is to store measurements of the same sample-space grid (same `scale` as in coordinate transforms) that are in different data types. For example, for widefield fluorescence microscopy, the raw camera frames are in unsigned integers, and the deconvolved images are in floating point numbers. In practice many people (including me) just store them in separate Zarr stores, which has it's own problems. What would be the recommended solution for this?

dstansby:
> In practice many people (including me) just store them in separate Zarr stores, which has it's own problems. What would be the recommended solution for this?

I think our recommendation would be to store them in separate OME-zarr datasets. Could you share some more details about what problems you have with storing them like this?

ziw-liu:
> I think our recommendation would be to store them in separate OME-zarr datasets. Could you share some more details about what problems you have with storing them like this?

Mostly the hassle of having multiple URLs for the same dataset. For example to visualize them in a viewer, the user needs to open each store individually; to run some simple processing (e.g. make a crop), the user needs to run the same pipeline multiple times. Each user would also need to come up with their own practices to keep track of all the associated stores, as opposed to being able to follow a standardized storage hierarchy.

normanrz:
I agree that passing around multiple URLs is a hassle. However, I don't think multiple multiscales were intended for that and almost no visualization tools support them properly.

Your use case is very similar to grouping segmentations or predictions together with the image. It also is similar to grouping images from other modalities through coordinate transformations. Together with a few other people, I am writing another RFC that aims to introduce collections for OME-Zarr that would enable these use cases. The draft is still very early but you can have a look here: https://hackmd.io/rQPfIMw0Q9SqP9WWspCiLQ# I would love to hear your feedback there.

joshmoore:
> I think our recommendation would be to store them in separate OME-zarr datasets.

Do you mean in a separate multiscale, @dstansby? Or does datasets here == array? Or a separate Zarr fileset?

dstansby:
(after looking at the spec for the right terms) I *think* I meant storing them as separate OME-Zarr images

ziw-liu:
> The draft is still very early but you can have a look here

I get 403 forbidden for the link. I'll definitely read once it's an RFC though!

normanrz:
That link currently requires a hackmd account to access.

normanrz:
> (after looking at the spec for the right terms) I _think_ I meant storing them as separate OME-Zarr images



> > I think our recommendation would be to store them in separate OME-zarr datasets.
> 
> Do you mean in a separate multiscale, @dstansby? Or does datasets here == array? Or a separate Zarr fileset?

I think you mean the same thing: a separate Zarr group with multiscale metadata.

perlman:
👍 (endorse)
Happy to see this written down, including the concise json example.

rfc/6/index.md
```diff
@@ -0,0 +1,190 @@
...
+>     if named["name"] == "3D":
+>         datasets = [x["path"] for x in named["datasets"]]
+>         break
+> if not datasets:
+>     # Use the first by default. Or perhaps choose based on chunk size.
+>     datasets = [x["path"] for x in multiscales[0]["datasets"]]
+> ```
+
+
+This RFC aims to codify what already seems to be common practice by moving from a multiscales array to a single multisclaes object. This will reduce complexity for implementations.
```
joshmoore, line 98:
```suggestion
This RFC aims to codify what already seems to be common practice by moving from a multiscales array to a single multiscale object. This will reduce complexity for implementations.
```

rfc/6/index.md
```diff
@@ -0,0 +1,190 @@
+# RFC-6: Multiscale
+
+Turn the `multiscales` array into a single `multiscale` object.
+
+
+## Status
+
+This proposal is very early. Status: D1
```
joshmoore, line 8:
```suggestion
This RFC is currently open for reviews (R1).
```

joshmoore (commented):
Few minor fixes that I'll apply immediately and then merge.

dstansby:
Is this meant to be showing up on https://ngff.openmicroscopy.org/rfc/index.html?

joshmoore:
Thanks @dstansby -- https://github.com/ome/ngff/pull/313

PR #288, opened 2025-01-13
OME-NGFF -> OME-Zarr on tools page

dstansby:
Cleans up the tools page a little bit by using OME-Zarr instead of OME-NGFF consistently.


joshmoore (approved):
Thanks, @dstansby.

Issue #289, opened 2025-01-20
Explicit define the relationship between axes and `dimension_names`


xgui3783:
This issue originated from https://imagesc.zulipchat.com/#narrow/channel/328251-NGFF/topic/.E2.9C.94.20does.20ng-ff.20v0.2E5.20axes.20metadata.20mandate.20.60dimension_names.60.2E.2E.2E

First of all, thanks for all the work you have done. The ngff (and zarr) specification is very detailed and easily understood. 

The question involves the `### 2.1. "axes" metadata` in ngff , where we have:

> The "dimension_names" attribute in the zarr.json of the Zarr array of a multiscale level MUST match the names in the "axes" metadata.

While the `dimension_names` is listed under optional members. 

Do I understand correctly that this would mean (emphasis my own):

> The "dimension_names" attribute in the zarr.json of the Zarr array of a multiscale level *MUST be defined* and MUST match the names in the "axes" metadata.

@ziw-liu commented the following:

> 1. Array metadata MAY contain dimension_names. (Zarr specification)
> 2. IF dimension_names is defined, it MUST be the same as axes names. (NGFF specification, no ambiguity)
> 3. IF dimension_names is undefined, implementations can skip the 'MUST MATCH' comparison, and use the axes names as-is. (My personal preference, could be wrong)

I would find it useful if this could be clarified in the specification. 

This issue is raised, so that any clarification PR can reference this issue.

joshmoore:
Thanks for opening this, @xgui3783.

I think the interpretation that is spelled out matches my understanding from previous specs. Cc'ing folks involved in spec work around this -- @bogovicj @normanrz @constantinpape -- in case that was either not the intent or that that intent is changing.

An additional aspect of this decision that we might want to consider is that in Zarr v2, the definition of the dimension names field was coming from the xarray community -- https://docs.xarray.dev/en/latest/internals/zarr-encoding-spec.html . With Zarr v3, this is now a part of Zarr and we might should reconsider if an additional name field is even needed.

----

*For reference, the Zarr spec page: https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#dimension-names*

ziw-liu:
> we might should reconsider if an additional name field is even needed

In NGFF the axis metadata contains more information than just the name. Without the name field, how would the name from the array metadata be associated with the extra metadata? Will it just be assumptions about list/array ordering?

will-moore:
[...]

PR #291, opened 2025-01-22
Clarifies that `dimension_names` remains an optional attribute

xgui3783:
fixes https://github.com/ome/ngff/issues/289


xgui3783:
superceded by https://github.com/ome/ngff/pull/292

Issue #303, opened 2025-03-26
RFC4: limit orientation metadata to spatial axes


dstansby:
As far as I can tell, RFC 4 doesn't limit orientation metadata to spatial axes. But it probably should - I don't think it makes any sense for there to be an anatomical orientation of a temporal axes!

dstansby:
I believe this has been taken into account now; thanks @thewtex!

PR #306, opened 2025-03-31
Add a comment ot RFC 4

dstansby:
Originally in https://github.com/ome/ngff/issues/303

### DIFF
--- a/rfc/4/comments/1.md
+++ b/rfc/4/comments/1.md
@@ -16,7 +16,7 @@ n/a
 
 RFC-4 defines anatomical orientation metadata.
 As currently written, it does not restrict this metadata to be present only on spatial axes.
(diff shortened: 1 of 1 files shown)


rfc/4/comments/1.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Summary
+
+n/a
+
+## Significant comments and questions
+
+RFC-4 defines anatomical orientation metadata.
+As currently written, it does not restrict this metadata to be present only on spatial axes.
+Because it does not make sense to define an anatomical orientation for non-spatial axes (e.g., channel or time axes), RFC4 should be updated to specific that anatomical orientation data can only be present on spatial axes.
```
joshmoore, line 19:
```suggestion
Because it does not make sense to define an anatomical orientation for non-spatial axes (e.g., channel or time axes), RFC4 should be updated to specify that anatomical orientation data can only be present on spatial axes.
```

PR #308, opened 2025-04-01
Migrate spec subdirectories to submodules (fix #271)

### LINKED ISSUE #271: spec diffs can be hard to read

joshmoore:
This is a fresh implementation of #282, creating new branches as of the state of the repo as of today (https://github.com/ome/ngff/commit/73083996ce716ffcaa9d5c08f594528c06d41c37). Once the build here is green I will:
 * [x] push the v0.x branches to ome/ngff
 * [x] update the URL of the submodules
 * [ ] merge in order to unblock 0.6.dev1

As mentioned in https://github.com/ome/ngff/pull/282#issuecomment-2769439432, this is a first step removing the duplicate information from the mainline and from here we have a few (partially independent) options:

 * move the branches to a new repository
 * use RTD subprojects rather than submodules
 * use subtree rather than submodules


will-moore:
@joshmoore I find that validator is no-longer able to find the schemas to validate with. E.g. https://raw.githubusercontent.com/ome/ngff/main/0.4/schemas/image.schema is 404.

Will there be an easy way to construct a similar link again?

joshmoore:
*Ack*. Sorry, @will-moore. I didn't realize raw URLs were being used. The new URL is: https://raw.githubusercontent.com/ome/ngff/v0.4/schemas/image.schema

i.e. replace `main/0.4` with `v0.4`.

PR #310, opened 2025-04-22
Enable json-schema-for-humans for all schemas

joshmoore:
As Will pointed out, the schema files were not being hosted properly by readthedocs. My assumption is that this is because the subdirectories do not contain HTML files. Incorporating @dstansby work in https://github.com/dstansby/ome-zarr-schema-docs
this runs the json-schema-for-humans generation for all schema directories.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,3 +18,6 @@ help:
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+clean:
+	rm -rf $(BUILDDIR) _bikeshed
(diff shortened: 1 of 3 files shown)


joshmoore:
Before I forget, another project we might want to integrate is https://sphinx-jsonschema.readthedocs.io/en/latest/directive.html

will-moore:
The human readable schemas look great, and I love that you can get a URL to each point in the schema e.g. https://ngff--310.org.readthedocs.build/0.5/schemas/image.html#tab-pane_ome_multiscales_items_datasets_items_coordinateTransformations_items_oneOf_i0

However, I still can't see that the schemas themselves are hosted / deployed under the `ngff` repo (as per my original question)?

joshmoore:
Finally found the issue with the validation jobs in the RFC-4 job. Merging.

Discussed with @jo-mueller embedding in the main text as a next step.

PR #311, opened 2025-04-30
Use s prefix for multiscale example

mkitti:
There is a tendency to replicate the examples exactly. Having multiscale paths only being a number can be easily confused with chunk paths. Overall, the examples are being interpreted as recommended conventions so we may want to consider the implications of the names used in the examples.

This is a minor change since it just modifies the example without modification to the actual schemas. Having a path that is just a number remains valid. This would also match the existing n5 convention:

https://github.com/saalfeldlab/n5-viewer?tab=readme-ov-file#n5-viewer-metadata

### DIFF
--- a/examples/multiscales_strict/multiscales_example.json
+++ b/examples/multiscales_strict/multiscales_example.json
@@ -16,7 +16,7 @@
           ],
           "datasets": [
             {
-              "path": "0",
+              "path": "s0",
               "coordinateTransformations": [
                 {
                   // the voxel size for the first scale level (0.5 micrometer)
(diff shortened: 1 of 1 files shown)


d-v-b:
related: https://github.com/ome/ngff/pull/277

will-moore:
Ah, it's actually index.bs: e.g. https://github.com/ome/ngff/blob/8cbba216e37407bd2d4bd5c7128ab13bd0a6404e/index.bs#L82 "...is often a sequence starting at 0"
I got there by clicking on the "0.5" submodules link from https://github.com/ome/ngff

jo-mueller:
Ugh sorry @will-moore, I had this ported already a while ago and forgot to raise the hand here clearer 😬 My bad!

PR #314, opened 2025-05-09
update name of the viewer

jburel:
Update name and link so it matches on-going on table see https://github.com/ome/ome-ngff-tools/pull/26


joshmoore:
Thanks, @jburel.

PR #318, opened 2025-07-08
rfc 4 updates

thewtex:
- RFC-4: Bump linkml to 1.9.2
- RFC 4: Models updates based on feedback

### DIFF
--- a/rfc/4/index.md
+++ b/rfc/4/index.md
@@ -7,7 +7,6 @@ reviews/index
 comments/index
 responses/index
 versions/index
-markdown/index
 ```
 
 Summary: An optional, explicit field for specification of imaging axis orientation, with primary focus on anatomical orientation for bipeds or quadrupeds.


rfc/4/orientation.yml
```diff
@@ -1,39 +1,330 @@
...
+      inferior-to-superior: Describes the directional orientation from below (inferior) to above (superior) in an anatomical structure or body.
+      superior-to-inferior: Describes the directional orientation from above (superior) to below (inferior) in an anatomical structure or body.
+      dorsal-to-ventral: Describes the directional orientation from the top/upper (dorsal) to the belly/lower (ventral) in an anatomical structure or body.
+      ventral-to-dorsal: Describes the directional orientation from the belly/lower (ventral) to the top/upper (dorsal) in an anatomical structure or body.
+      dorsal-to-palmar: Describes the directional orientation from the top/upper (dorsal) to the palm of the hand (palmar) in a body.
+      palmar-to-dorsal: Describes the directional orientation from the palm of the hand (palmar) to the top/upper (dorsal) in a body.
+      dorsal-to-plantar: Describes the directional orientation from the top/upper (dorsal) to the sole of the foot (plantar) in a body.
+      plantar-to-dorsal: Describes the directional orientation from the sole of the foot (plantar) to the top/upper (dorsal) in a body.
```
thewtex: Todo: add back dropped values!

rfc/4/orientation.ts
```diff
@@ -27,19 +121,52 @@ export enum Orientation {
...
+export interface Axes {
+    /** A list of axes. Although serialized as list, it MUST be dealt with as being a set as in the name of each axis MUST be unique. Furthermore, if the attribute orientation is defined for one axis of type space, it  MUST be defined for all the axes of type space. In this case, the type of each orientation MUST be the same and the value MUST be unique.
+ */
+    axes?: string[],
+}
+
+
+
+export interface Axis {
+    name: string,
```
thewtex, line 149: Reference limited typing here

rfc/4/orientation.yml
```diff
@@ -1,39 +1,330 @@
...
+
+  AnatomicalOrientation:
+    attributes:
+      type:
+        range: string
+        required: true
+        equals_string: "anatomical"
+      value:
+        range: AnatomicalOrientationValues
+        required: true
```
thewtex: Note, at least in text (and LinkML if possible), that new future `orientation`'s should have `type` and `value`.

rfc/4/orientation.yml
```diff
@@ -1,39 +1,330 @@
...
+  linkml: https://w3id.org/linkml/
+  orcid: https://orcid.org/
+imports:
+  - linkml:types
+default_range: string
+default_prefix: ngff
+
+created_on: "2025-05-29"
+created_by: orcid:0000-0003-1666-5421
+contributors: orcid:0000-0001-9475-3756
```
thewtex: Add David F's ORCID here.

thewtex: `contributors` should be an array

thewtex: And Wouter.

rfc/4/orientation.yml
melonora:
```suggestion
      - unit

  Orientation:
    abstract: true
    attributes:
      type:
        range: string
        required: true
      value:
        required: true

  AnatomicalOrientation:
    is_a: Orientation
    attributes:
      type:
        range: string
        required: true
        equals_string: "anatomical"
        ifabsent: string(anatomical)
      value:
        range: AnatomicalOrientationValues
        required: true
```

Small side note, linkml does not strictly respect inheritance here as it makes value optional if you don't specify `required: true`.

rfc/4/orientation.yml
```diff
@@ -1,39 +1,330 @@
...
+
+slots:
+  name:
+    description: The name identifier for an axis
+    range: AxesNames
+    required: true
+  type:  
+    description: The type category of an axis
+    range: AxisType
+    required: true
```
melonora:
```suggestion
slots:
  name:
  type:  
```
This would make the slots generally reusable throughout the specification.

thewtex:
@joshmoore what should be done about the `Validation` checks?

neurolabusc:
@thewtex your text
```
Note: In bipeds, anterior/posterior corresponds to ventral/dorsal due to upright posture. In quadrupeds, dorsal refers to the back (spine) side and ventral to the belly side. The terms superior/inferior are rarely used in veterinary anatomy except for specific structures like the eye.
```
is inaccurate. In bipeds anterior/posterior direction is the same as ventral/dorsal for the **spine**, but it is rostral/caudal for the **brain**. With respect to quadrupeds, the bipedal head is rotated 90-degrees relative to the rest of the body. This explains why the DICOM standard specifies the unambiguous [anterior/posterior and head/foot directions for bipeds](https://dicom.innolitics.com/ciods/ct-image/general-series/00102210)

My sense is you could rewrite this:
```
Note: In bipeds, the terms rostral/caudal and dorsal/ventral are ambiguous because their orientation depends on the anatomical structure. For example, these directions are orthogonal when describing the human brain versus the spinal cord. Therefore, the unambiguous terms dorsal/ventral and inferior/superior (sometimes referred to as head/foot) are generally preferred for bipeds.
```

thewtex:
@neurolabusc :+1: I made an update accordingly.

joshmoore:
```
WARNING: 'myst' cross-reference target not found: 'types/String.md' [myst.xref_missing]
```

Probably more for @melonora, any ideas what we should do about the links to the String.md file?

joshmoore:
Ah, ok. So hence I can add a command to ignore those markdowns.

melonora:
can confirm it is markdown

rfc/4/index.md
```diff
@@ -168,48 +198,61 @@ Who has a stake in whether this RFC is accepted?
...
   - Discussed at the Get Your Brain Together HKC03
+  - Discussed at the OME-Zarr 2024 Zurich hackathon
 
 ## Implementation
 
 The implementation will involve adding the `orientation` field to the `axes` metadata
 in the NGFF schema. The field will use the controlled vocabulary specified above.
 
+A complete working implementation of this RFC is available in the `ngff-zarr` package, including validation, serialization, and examples. See the [ngff-zarr RFC-4 documentation](https://ngff-zarr.readthedocs.io/en/latest/rfc4.html) for detailed usage examples and API reference.
+
```
joshmoore, line 209: ❤️

rfc/4/index.md
```diff
@@ -43,58 +44,87 @@ Existing standards provide some prior art for this proposal:
...
 ## Proposal
 
-We propose adding an OPTIONAL `anatomicalOrientation` field to the `axes` metadata in the OME-Zarr specification. This field will use a controlled vocabulary to explicitly define the anatomical orientation of the image.
+We propose adding an OPTIONAL `orientation` field to the `axes` metadata in the OME-Zarr specification. This field will use a controlled vocabulary to explicitly define the orientation of the image.
 
-This metadata data MUST only be used in cases where there is a single subject in the acquired volume or extracted image region-of-interest and the subject is roughly aligned to the imaging axes.
+The `orientation` field MUST only be used on spatial axes (axes with `type` field set to `"space"`). It MUST NOT be used on non-spatial axes such as channel, time, or other axis types.
 
-### Controlled Vocabulary
+This metadata MUST only be used in cases where there is a single subject in the acquired volume or extracted image region-of-interest and the subject is roughly aligned to the imaging axes.
```
dstansby, line 51: Could there be cases where there are multiple subjects, but they are all aligned such that orientation still makes sense for all of them, for example imaging two heads in one image that are oriented in the same direction?

rfc/4/index.md
```diff
@@ -43,58 +44,87 @@ Existing standards provide some prior art for this proposal:
...
 
-The controlled vocabulary for the `anatomicalOrientation` field will include:
+The `orientation` field is structured as an object with a `type` field that specifies the orientation domain (e.g., "anatomical") and a `value` field that specifies the specific orientation within that domain.
```
dstansby, line 53: Is type limited to a string, or can it be an arbitrary JSON object?

dstansby, line 53: To be consistent with how the rest of the spec defines required fields, should this instead be something like

> The `orientation` field is structured as an object and MUST have a `type` field that specifies the orientation domain (e.g., "anatomical") and MUST have a `value` field that specifies the specific orientation within that domain.

rfc/4/index.md
```diff
@@ -43,58 +44,87 @@ Existing standards provide some prior art for this proposal:
...
-- `rostral-to-caudal`
-- `caudal-to-rostral`
+- engineering/microfluidics: upstream/downstream (flow direction)
+- geographical: north/south, east/west
+- oceanographic: increasing-depth
+- histological: basal/apical
 
-- `cranial-to-caudal`
-- `caudal-to-cranial`
+For anatomical orientation, the controlled vocabulary for the `orientation` field's `value` will include:
```
dstansby, line 93: This doesn't sspecify what the "type" field MUST be set to for using this controleld vocabulary; presumably it MUST be set to "type": "anatomical"?

rfc/4/index.md
```diff
@@ -43,58 +44,87 @@ Existing standards provide some prior art for this proposal:
...
+- `rostral-to-caudal` (nose/beak-to-tail, especially for nervous system)
+- `caudal-to-rostral` (tail-to-nose/beak, especially for nervous system)
+- `cranial-to-caudal` (head-to-tail)
+- `caudal-to-cranial` (tail-to-head)
+- `dorsal-to-ventral` (back/top-to-belly/bottom)
+- `ventral-to-dorsal` (belly/bottom-to-back/top)
+
+Note: In bipeds, the terms rostral/caudal and dorsal/ventral are ambiguous because their orientation depends on the anatomical structure. For example, these directions are orthogonal when describing the human brain versus the spinal cord. Therefore, the unambiguous terms anterior/posterior and inferior/superior (sometimes referred to as head/foot) are generally preferred for bipeds.
+
 A set of NGFF `axes` MUST only have one of the set `{ "left-to-right", "right-to-left" }` or `{ "anterior-to-posterior", "posterior-to-anterior" }` or the remaining values.
```
dstansby, line 121: I think these sets need updating to the new `{type:..., value:...}` structure of the field.

rfc/4/index.md
thewtex: No, this is for a single subject, as stated.

rfc/4/index.md
thewtex: The type is a string.

rfc/4/index.md
thewtex: Yes, "type": "anatomical" is currently the only type defined.

rfc/4/index.md
thewtex: Is is not demonstrating the structure of the field but using mathematical set notation.

rfc/4/index.md
```diff
@@ -260,12 +304,36 @@ class Orientation(str, Enum):
...
+    Orientation object that can represent different types of orientation information.
+    """
+    type: str
+    value: str
+
+class AnatomicalOrientation(Orientation):
+    """
+    Anatomical orientation specific implementation.
+    """
+    type: str = "anatomical"
```
dstansby, line 319: I don't *think* it was mentioned above that the "type" field should take a default value if not present?

rfc/4/index.md
dstansby: It works for a single subject, but I'm saying couldn't there be images where there are multiple subjects and defining an orientation makes sense? In which case it might be nice to remove this restriction of having a single subject.

rfc/4/index.md
dstansby: But I don't think it's specified that the controlled vocabularly defined here is specifically for the case of "type": "anatomical". To suggest a concrete fix,

> For anatomical orientation,

could be changed to

> For orientation where the "type" field is "anatomical",

rfc/4/index.md
thewtex: If the type is `AnatomicalOrientation`, then "type" has a value "anatomical".

rfc/4/index.md
dstansby: > Yes, "type": "anatomical" is currently the only type defined.

Do you mean that the "type" field cannot take an arbitrary value, but instead can only take the value "anatomical"?

rfc/4/index.md
dstansby: Does that value have to be present in the JSON metadata though, even if it can only take one valid value, or can it be absent?

rfc/4/index.md
thewtex: @dstansby I added the MUST's here: https://github.com/ome/ngff/pull/329

rfc/4/index.md
thewtex: Yes.

rfc/4/index.md
thewtex: It has to be present.

rfc/4/index.md
thewtex: For simplicity we only support a single subject.

rfc/4/index.md
dstansby: Is that stated anywhere in the current RFC? Maybe I missed a section, but I don't see a bit that says that the only value it can take is "anatomical".

rfc/4/index.md
dstansby: In that case I think the correct way to model that in Python+pydantic is
```python
type: Literal["anatomical"]
```
which requires the value to be present (it has no default value), but limits it to a single value using the typing.

rfc/4/index.md
thewtex: An explicit statement is here: https://github.com/ome/ngff/pull/329

rfc/4/index.md
thewtex: Added here: https://github.com/ome/ngff/pull/330

PR #319, opened 2025-08-01
Add Microscopy Nodes to the list of Image Viewers

Tom-TBT:
I added Microscopy Nodes to the tool index as previously discussed


joshmoore:
Validation errors are due to Ubuntu 20.04 in the job. Just fixed these for RFC-4.

Issue #323, opened 2025-08-20
Provide contribution guide on how to proposed changes to JSON schemas


dstansby:
I have identified a fix that needs to be made to the JSON schemas, for at least versions 0.4 and 0.5 of the specification. It is completely unclear to me how I can submit a pull request to make this fix however, because there is a complex setup of recursive submodules for each version of the specification.

This currently presents a huge barrier to community contributions to the JSON schemas, so it would be good if someone who understands the workflow of updating different versions of the specification can put together a contribution guide explaining how to do this.

Alternatively, I would strongly advocate for the JSON schemas to be moved to a separate repository, so they could have a simpler (flat) repsository layout, and making fixes to them would be much simpler than having to deal with the complex submodule setup in this repository.

Issue #322, opened 2025-08-14
Plate JSON examples in the specification are invalid


dstansby:
The specification says:

> The plate dictionary MUST contain a version key whose value MUST be a string specifying the version of the plate specification.

But this version key is not in the example JSON metadata given later in [section 2.7](https://ngff.openmicroscopy.org/0.5/index.html#plate-md).

dstansby:
Surely the specification is, by definition, correct? The top of the specification says

> This document contains next-generation file format (NGFF) specifications for storing bioimaging data in the cloud

And the specification makes no reference to the JSON files.

jo-mueller:
Also, related: https://github.com/ome/ngff/issues/309

Issue #309, opened 2025-04-15
Are previous version keys allowed in v0.5?


ziw-liu:
Context: I'm implementing v0.5 and need a clear definition of various `version` fields in the v0.4/v0.5 spec (MUST, MUST NOT, SHOULD, MAY etc).

From RFC-2:

https://github.com/ome/ngff/blob/106d7d9d5726182226ca49297b058570707c354b/rfc/2/index.md?plain=1#L261

And the latest JSON schema do not contain `.ome.plate.version`, `.ome.multiscales[0].version` etc.

However the spec also states that:

https://github.com/ome/ngff/blob/8cbba216e37407bd2d4bd5c7128ab13bd0a6404e/index.bs#L501-L502

https://github.com/ome/ngff/blob/8cbba216e37407bd2d4bd5c7128ab13bd0a6404e/index.bs#L545-L546

So we have the following situation, using `.ome.plate.version` as an example:

- RFC-2, JSON examples in the v0.5 spec, and the schema indicate that having this field should cause a validation error.
- The v0.5 spec text says that NOT having this field should cause a validation error.

Different implementations can interpret these requirements in ways that are not compatible with each other.

normanrz:
That needs correcting in the spec doc. Neither plates nor wells should have a `version` key.

dstansby:
Since the JSON schema is just one implementation of the specification, I think the thing to do here is:

- Fix the JSON schema to match the specification
- Change these fields in v0.6 of the specification so they are removed as intended originally for v0.5

PR #324, opened 2025-08-21
sync spec text with json examples

jo-mueller:
Dear all,

This PR is intended as a fix for #322 and #309. Some confusion was raised in these issues as to whether different `version` keys would be allowed in an ome-zarrs top-level `zarr.json` and in a following `plate`/`well`/`labels` groups therein. 

In the current examples on the 0.5 branch (e.g., [plate example](https://github.com/ome/ngff/blob/8cbba216e37407bd2d4bd5c7128ab13bd0a6404e/examples/plate_strict/plate_2wells.json)), the `version key` is only given at the top level and not in the group-level. This is in contradiction to the current text of the [plate spec](https://ngff.openmicroscopy.org/0.5/index.html#plate-md), which states:

> The plate dictionary MUST contain a version key whose value MUST be a string specifying the version of the plate specification.

Same for [wells ](https://github.com/ome/ngff/blob/8cbba216e37407bd2d4bd5c7128ab13bd0a6404e/examples/well_strict/well_2fields.json) and [labels ](https://github.com/ome/ngff/blob/8cbba216e37407bd2d4bd5c7128ab13bd0a6404e/examples/label_strict/colors_properties.json) groups examples.

## Context and consequences

For context, the `version` key was originally moved up into the top-level `ome` group under the pretext of discouraging different metadata versions inside the same file according to RFC2:

> While technically possible, OME-Zarr 0.5 (with Zarr v3) and OME-Zarr 0.4 (with Zarr v2) metadata could exist side-by-side in a Zarr hierarchy, it is NOT RECOMMENDED. This may be useful for short periods of time (i.e. during migrations from 0.4 to 0.5), but should not be used longer term. Multiple metadata versions can lead to conflicts, which may be hard to resolve by implementations. If implementations encounter 0.4 and 0.5 metadata side-by-side, 0.5 SHOULD be treated preferentially.

Removing the statements about the `version` keys in the plate/well/labels groups essentially drops the idea differing versions inside a single ome.zarr file, which I think is a reasonable step from an implementation perspective. Also, it brings the spec back into sync with the examples, which I think is important for the integrity of this project.

I modified the statement at the top-level `ome` group into the following to make this clear:

> The OME-Zarr Metadata is stored in the various `zarr.json` files throughout the above array
hierarchy. The OME-Zarr Metadata version MUST be consistent within a hierarchy. The top-level `ome`
group MUST contain a `version` key whose value MUST be a string specifying the version of the
metadata specification.

Let me know what you think :)

cc @joshmoore @normanrz @dstansby @ziw-liu 

PS: Hope I got this PR configured right - had a bit of trouble getting the knack about contributing with submodules (#323)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -47,20 +47,6 @@ implementations that are later submitted as a formal specification. (See [[#bf2r
 Some of the JSON examples in this document include comments. However, these are only for
 clarity purposes and comments MUST NOT be included in JSON objects.
 
-Extra keys in JSON objects
---------------------------
-
-Unless otherwise stated, a JSON object defined in this document MAY contain additional keys
-beyond those enumerated in its definition. The presence of an additional key in such a JSON
-object MUST NOT be treated as an error by implementations.
-
-Whether an implementation ignores the extra keys or propagates them is up to that particular
-implementation. It is recommended that implementations clearly declare how they handle extra
-keys, with a particular emphasis on informing users in the context of irreversible operations.
-For example, an implementation that resaves OME-Zarr data and ignores extra keys could implement
-a routine that warns users when they attempt to resave OME-Zarr data with extra keys.
-
-
 Storage format {#storage-format}
 ================================
 


index.bs
```diff
@@ -158,11 +158,12 @@ The "OME-Zarr Metadata" contains metadata keys as specified below
 for discovering certain types of data, especially images.
 
 The OME-Zarr Metadata is stored in the various `zarr.json` files throughout the above array
-hierarchy. In this file, the metadata is stored under the namespaced key 
-`ome` in `attributes`.
-The version of the OME-Zarr Metadata is denoted as a string in the `version` attribute within the `ome` namespace.
+hierarchy. The OME-Zarr Metadata version MUST be consistent within a hierarchy. The top-level `ome`
+group MUST contain a `version` key whose value MUST be a string specifying the version of the
```
d-v-b:
```suggestion
hierarchy. The OME-Zarr Metadata version MUST be consistent within a hierarchy. The attributes of the top-level `ome`
group MUST contain a `version` key whose value MUST be a string specifying the version of the
```

jo-mueller: @d-v-b thanks for the feedback . Tbh, to me that would sound like the hierarchy being like this?

```json
"ome": {
  "attributes": {
    "version": "0.5"
  }
}
```

but in the examples it's the other way around?
```json
"attributes": {
  "ome": {
    "version": "0.5"
  }
}
```

Unless "the attributes of the top-level `ome`" are simply its members? If that is the case, I would rather only use the term "attributes" to the actual metadata field "attributes 😬

d-v-b: quite right, I mistakenly thought "the ome group" denoted a zarr group, instead of a key in a JSON object

d-v-b: maybe avoid the term "group" here, since it's overloaded in a Zarr context? "`"ome"` object" or "`"ome"` key" might be more clearly scoped to JSON

jo-mueller: That's a really good point. Maybe something like that could be added to the contribution guidelines in the future to avoid confusion with the terminology. The correct phrasing would then be: 

> The top-level `ome` key MUST contain a `version` key whose value MUST be a string specifying the version of the ...

d-v-b: I would probably say 

> The group attributes MUST contain a key `"ome"`. The value of the `"ome"` key MUST be a JSON object that MUST contain a `"version"` key, the value of which MUST be a string... 

But I also think english prose is kind of underpowered for describing JSON, and for me a tabular form works a lot better:

[Group.attributes](###group.attributes)

| key | value | required |  notes
| ---- | -------- | -------- | ---- | 
| `"ome"` | [`OmeZarrMetadata`](#OmeZarrMetadata) | yes |

[OmeZarrMetadata](####OmeZarrMetadata)

| key | value | required |  notes
| ---- | -------- | -------- | ---- | 
| `"version"` | string | yes | must specify the version ....

But your suggestion also totally works here

jo-mueller: Fixed in 4be384eca123e213219057ed18fe4b39609aedc1

Also, completely agree on the shortcomings of written text for concepts like nested JSON objects. Maybe it's as much a shortcoming of english language as it is of bikeshed as a document format. Sphinx provides a few cool [directives](https://sphinx-jsonschema.readthedocs.io/en/latest/directive.html) that just spawn such tables from json files but they are not available for bikeshed :/

normanrz: I like the suggestion of describing the keys in tabular form. I think that would help to avoid confusion.

jo-mueller: Will raise an issue to discuss human-readability further!

will-moore:
@jo-mueller I'm not up-to-date with the strategy for PRs, or even know if this should be opened against v0.5 (which would create a v0.5.n+1 version of the spec) or if we just go for v0.6?
The discussion at https://github.com/ome/ngff/issues/276 seems to have some pointers... I'll see what I can understand from that...

ziw-liu:
For existing v0.5 datasets that have version keys inside plate and well objects, will this change make them invalid OME-Zarr files?

ziw-liu:
> Generally I don't think there's language in the spec that prevents extra keys in the various JSON objects

I agree with this observation. However in practice various implementations do prohibit extra keys. Maybe a separate clarification is needed, e.g. `ome objects MAY contain keys not described in this specification`.

index.bs
```diff
@@ -47,6 +47,21 @@ implementations that are later submitted as a formal specification. (See [[#bf2r
 Some of the JSON examples in this document include comments. However, these are only for
 clarity purposes and comments MUST NOT be included in JSON objects.
 
+Extra keys in JSON objects
+--------------------------
+
+Unless otherwise stated, a JSON object defined in this document MAY contain additional keys
+beyond those enumerated in its definition. OME-Zarr Implementations MUST allow extra keys in
+such JSON objects. This means that, when reading OME-Zarr data, the presence of an extra key
+in such a JSON object MUST NOT be treated as an error.
```
dstansby:
```suggestion
Unless otherwise stated, a JSON object defined in this document MAY contain additional keys
beyond those enumerated in its definition.
```
I think the second two sentences are redundant - other existing sections of the specification do not explicitly say how/when implementations should error, but instead just enumerate conditions for valid OME-Zarr Zarr groups (including their metadata)

jo-mueller: Hi @dstansby , thanks for the input. Maybe we can shorten it a bit (see following) to keep in concise?

> Unless otherwise stated, a JSON object defined in this document MAY contain additional keys
beyond those enumerated in its definition. This means that, when reading OME-Zarr data, the
presence of an extra key in such a JSON object MUST NOT be treated as an error.

Given the comments of @ziw-liu and @LDeakin I feel like it has merit to spell this out once, clearly. Beyond that I agree, the logic following the conditional phrasing in the spec should lead to the correct conclusion throughout the document and further clarification is not needed.

dstansby: 👍 sounds good; how about a slightly shorter version (as you can tell I think there's high value in making the spec as short and to the point as possible!):

> Unless otherwise stated, a JSON object defined in this document MAY contain additional keys
beyond those enumerated in its definition. The presence of an additional key in such a JSON object MUST NOT be treated as an error.

I also switched "extra" > "additional" in the second sentence, to be consistent with the first sentence.

d-v-b: +1 to simplifying the text, but fwiw "the spec allows extra keys" is a weaker statement than "implementations must allow extra keys". We want the latter.

jo-mueller: All sound good to me, added in 251fca92f95e4b621a6e01e4d50d9ecd7b48d0fc🫡

d-v-b: > (as you can tell I think there's high value in making the spec as short and to the point as possible!):

In this spirit, it might be helpful to separate normative sections of the spec from "implementation notes" that would go in their own section like this

> [!NOTE]
> dont randomly delete user data

Altering the implementation notes would not be considered changing the spec, and so this would simplify questions of spec versioning

dstansby: > +1 to simplifying the text, but fwiw "the spec allows extra keys" is a weaker statement than "implementations must allow extra keys". We want the latter.

(out of interest/my own confusion) What's an example metadata document that would satisfy "the spec allows extra keys", but an implementation could still error on and be compliant with the spec?

d-v-b: > (out of interest/my own confusion) What's an example metadata document that would satisfy "the spec allows extra keys", but an implementation could still error on and be compliant with the spec?

Sorry, I meant "stronger" in a rhetorical sense -- "your implementation must not error when encountering extra keys" is IMO a lot more direct than "extra keys are allowed in `<json object>`". You're totally right that these are two different ways of saying the same thing.

d-v-b: and if everyone thinks the global "extra keys are allowed" language is clear enough, we don't need the stronger statement

jo-mueller: @d-v-b in the interest of separating concerns of PRs - would you agree to this being marked as resolved? I totally agree that there is much left to be desired with the organization of the document, but maybe that's for a future PR? Would love to make this document more visual/clear/better for readers.

d-v-b: absolutely!

dstansby:
It's worth noting that if arbitrary extra keys are allowed, then this removes any future possibility of backwards compatibility between versions of the specification in future.

To give a concrete example, say someone writes data that contains a `color: "blue"` field somewhere in the "ome" metadata in a hypothetical version 1.2 of OME-Zarr, and then version 1.3 of OME-Zarr introduces a specifciation for the color field that says the metdata MAY have a "color" field and it MUST be a color hex string. The previous `color: "blue"` metadata would now be invalid. This means OME-Zarr 1.2 data could not generally also be considered valid OME-Zarr 1.3 data.

If instead arbitrary extra keys are prohibited in the "ome" metadadata, it allows the possibility of all valid metadata written with one version of the specification to also be valid with the next version.

Given OME-Zarr has reserved the top-level "ome" key in the JSON attributes, I actually now think that extra keys should be prohibited to give the option of backwards compatibility. Instead, extra keys should live outside the "ome" field.

This would also limit proliferation of "unofficial" metadata under the "ome" field, keeping the metadata there tightly controlled to what is allowed within the specifciation.

d-v-b:
This is a good point! In terms of what is right for the future of the spec, I don't have strong feelings about whether extra keys should be allowed or not. I can definitely see the appeal of keeping the `"ome"` namespace tight. Users have the rest of the `attributes` namespace for their own stuff. On the other hand, ome-zarr might want to define some extensibiltiy semantics, in which case maybe extra keys could be allowed, as long as they can be differentiated from "official" keys (e.g., via a `"omezarr."` prefix or something. but that could get tedious.

But in terms of interpreting what the spec says today, I think disallowing extra keys would be potentially breaking for anyone who saved data with an extra key because they thought it was allowed.

d-v-b:
>    Users MUST NOT add arbitrary fields inside the ome object or sub-objects, to allow backwards compatibility if fields of the same name are added in future versions of the OME-Zarr spec.
>
>    Implementations SHOULD prevent users from adding such arbitrary fields where possible.

This requires language permitting lower-level objects to relax this rule as needed, since the spec currently recommends(!) [saving an unstructured dict in the `multiscales` metadata](https://github.com/ome/ngff/blob/8cbba216e37407bd2d4bd5c7128ab13bd0a6404e/index.bs#L327):

> It SHOULD contain the field "metadata", which contains a dictionary with additional information about the downscaling method.


> Nobody can stop a JS or python user from doing something stupid if they really try, but this at least sets expectations and guides implementations with better type systems.

IMO JS (via typescript) and python both offer everything we need to completely model this stuff. Whether people _use it_ is a separate question.

jo-mueller:
Thanks for linking the issue! I'll write up a summary of this thread over there.

d-v-b (approved):
This looks good to me, _provided_ that everyone considers it obvious that extra keys are allowed for reading. Because this change removes required metadata keys, there will be datasets in the wild that were previously compliant and used these keys. They should continue to be compliant after these changes. That means their extra keys must be accepted by readers.

dstansby:
Presumably (because data written after this change is not compatible with the spec before this change) this

jo-mueller:
As for versioning, I think the idea now would be to create a new branch (`0.6dev`) or something of the sort and then collect all changes for a future `0.6` there. For release, the branch should be renamed to `0.6` the branch would be added as a submodule to the main repo.  The link under `latest` should be changed to point to the `0.6` submodule.

Maybe @joshmoore could help us there?

PR #342, opened 2025-09-27
Some improvements to the tools page

dstansby:
- Remove tools that do not mention OME-Zarr in their documentation. This includes some tools that work with Zarr but not specifically OME-Zarr. In order to keep the list focused and specific to OME-Zarr, I think this helps, but opinion may differ?
- Add a new validation section. Because ome-zarr-models now supports CLI validation, and there's the web validator, I think it's worthy of it's own section now.
- Use "Dataset" instead of "File" or "Image" in the titles, because OME-Zarr datsets are not single files and some of them are not just images but collections of images.
- Re-order the entries so they're alphabetical.


tools/index.md
```diff
@@ -4,54 +4,56 @@ A list of tools and libraries with OME-Zarr support. These are developed by vari
...
-| ITKIOOMEZarrNGFF | [![image](https://github.githubassets.com/favicons/favicon.svg)](https://github.com/InsightSoftwareConsortium/ITKIOOMEZarrNGFF) | An ITK external module for IO of images stored in OME-Zarr file format. |
-| iohub| [![image](https://github.githubassets.com/favicons/favicon.svg)](https://github.com/czbiohub-sf/iohub) | Pythonic and parallelizable I/O for N-dimensional imaging data with OME metadata |
 
 
+## Dataset Validation
+
+| Name    | Link | Description |
+| -------- | ------- | ------- |
+| ome-ngff-validator | [![image](https://github.githubassets.com/favicons/favicon.svg)](https://github.com/ome/ome-ngff-validator) | Web page for validating OME-Zarr files. |
+| ome-zarr-models | [![image](https://github.githubassets.com/favicons/favicon.svg)](https://github.com/ome-zarr-models/ome-zarr-models-py) | Python package and and command line interface that can validate OME-Zarr files |
```
will-moore: Duplicate "and"

dstansby: thanks, fixed 👍

joshmoore:
> I think this is good to go now (?)

Not quite. I've tried to implement 

> Please do add back Bio-Formats ...

meeting both of our requirements. Please let me know what you think of: https://ngff--342.org.readthedocs.build/tools/index.html#dataset-reading-writing

image

d-v-b:
do you want to direct people to OmeZarrReader? It seems like it's not [going to be maintained](https://github.com/ome/ZarrReader/issues/109#issuecomment-3013108390): 

> There probably won't be much further development happen on the ZarrReader. Don't expect that it'll will keep up with zarr versions and ngff spec versions evolving.

PR #343, opened 2025-09-29
RFC-8: Collections

normanrz:
This is the work-in-progress draft for RFC-8.

cc @jluethi @lorenzocerrone @tischi @perlman @matthewh-ebi


rfc/8/index.md
```diff
@@ -0,0 +1,1117 @@
...
+
+
+### Metadata
+
+This RFC defines two main objects for OME-Zarr: `Collection`, `CollectionNode`. 
+
+#### `Collection` keys
+
+* `"type"` (required). Value must be `"collection"`.
+* `"nodes"` (required). Value must be an array of `CollectionNode` or `Collection` objects.
```
d-v-b: since every node has a unique name, why is this an array and not an object?

rfc/8/index.md
```diff
@@ -0,0 +1,1117 @@
...
+Nodes either reference OME-Zarr images (multiscales) or collections.
+Nested collections can also be inlined in the metadata.
+Nodes reference images or collections that are stored locally relative to the collection or remotely (using URLs).
+Arbitrary user or implementation metadata may be added to collections or nodes, which is an opportunity to add metadata that is only valid for a node in the context of a collection (e.g. rendering settings).
+Images may be added as nodes to multiple collections.
+
+
+### Metadata
+
+This RFC defines two main objects for OME-Zarr: `Collection`, `CollectionNode`.
```
d-v-b: A `CollectionNode` can be a `Collection`, so it's a bit confusing to say that these are two objects unless you explain that "object" here means something like "interface" or "protocol"

rfc/8/index.md
normanrz: Yeah that could also work.

rfc/8/index.md
normanrz: What would be the best term here? Is it a class?

d-v-b: As I understand it, there are currently 3 entities that need to be defined:
- collection
- multiscales
- root 

collection and multiscales can be discriminated based on their `type` field, and `collection` has attributes that `multiscales` does not, so regular inheritance from a base class doesn't express their relationship very well.

Maybe defining these as protocols would work? e.g., there's a core `Node` protocol, which the fields `{type, name, attributes}`, and objects that implement `Node` can also implement `Collection` OR `Multiscales` (but not both, because of the requirement on the `type` key). Finally, there's a `Root` protocol which can only be implemented by a `Collection`

rfc/8/index.md
```diff
@@ -0,0 +1,1117 @@
...
+
+While attributes are effective for creating specialized collections and nodes, implementations are not required to parse them.
+Implementations are encouraged to provide graceful fallback strategies for specialized collections and nodes that are not understood by the implementation. 
+Strategies could include falling back to basic collection semantics, providing selector screens, or rendering nodes with default settings.
+
+
+### Examples
+
+#### Simple example
+```json
```
d-v-b: With this tagged as regular JSON the comments appear as a syntax error. I'm not sure if tagging this with `jsonc` would work, but if it doesn't, I think a `javascript` tag would make the comments appear commenty

rfc/8/index.md
will-moore: Presumably `bioformats2raw.layout` and `plate` collections will still be around (not removed with this proposal). So a `Node` could be `Collection` or `Multiscales` or `bioformats2raw` or `plate`?

d-v-b: actually I was wrong, regular inheritance isn't problematic for `Collection` and `Multiscales` -- there's a base `Node`, `Collection` and `Multiscales` (and anything else) inherit from `Node` (totally fine for them to add new attributes as children).

As for the requirement is that there be only 1 root node, I don't think that can be expressed in a type system easily as long as the root is structurally compatible with a `Collection`, but that can be added as a regular requirement

d-v-b: If the requirement is just that the root node _have_ `version` (weaker than requiring that _only_ the root node have `version`), then this is a bit simpler.

d-v-b: Suggested text:

This RFC defines a basic interface for a OME-Zarr metadata object, which we name `Node`. Objects that implement `Node` must have the following fields: 

| field | type | required? | notes | 
| - | - | -| -|
| type | string | yes | Identifies the type of the node | 
| name | string | yes | Value must be a non-empty string. It should be a string that matches `[a-zA-Z0-9-_.]+`. |
| attributes | object | no | See [link to attributes section][]

The `type` field of a `Node` defines the additional fields, if any, it has. This RFC proposed two `Node` types: `Multiscale` and `Collection` ...

d-v-b:
this is looking really cool!

rfc/8/index.md
normanrz: > Presumably `bioformats2raw.layout` and `plate` collections will still be around (not removed with this proposal). So a `Node` could be `Collection` or `Multiscales` or `bioformats2raw` or `plate`?

The idea is to remove `bioformats2raw.layout` and `plate` as separate entities with this proposal and express the functionality through attributes in the collection nodes. We need to work more on these.

rfc/8/index.md
normanrz: I wonder if representing an order may be desired, though. For example, https://ngff.openmicroscopy.org/latest/index.html#bf2raw states "Parsers like Bio-Formats define a strict, stable ordering of the images in a single container ...".
If it were an object the ordering would likely get lost in some JSON implementations. It could be represented through sortable node names, but that also seems less convenient.

d-v-b: order might also be useful for collections of layers in the context of an image visualization tool. Although you can always add an `"order"` field to the elements that's an integer (sort of the reverse of adding a `"name"` field that must be unique in the container).

rfc/8/index.md
```diff
@@ -0,0 +1,1224 @@
...
+                "example-viewer:settings": {
+                    "isDisabled": true
+                },
+                "other-viewer:voxelType": "intensities"
+                ... // arbitrary user-defined metadata
+            },
+        }, {
+            "name": "..",
+            "type": "collection",
+            "path": "./nested_collection.json"
```
will-moore: The collection should be a directory that contains a `zarr.json`, right?
e.g. `"path": "./nested_collection.zarr"`

will-moore: Ah, now I see that this standalone json file is proposed as part of this RFC. But that isn't covered until much later below under Examples `Where is this collection metadata stored?`. Maybe that should be moved up above this point?

If an implementation is using e.g. `zarr-python` or another zarr library to retrieve zarr metadata, then it may be kinda painful to also support fetching of vanilla `file.json` files using a different mechanism? Don't know about other libs.

rfc/8/index.md
```diff
@@ -0,0 +1,1224 @@
...
+
+#### `Multiscale` Node
+
+References a Node that represents an OME-Zarr multiscale image.
+
+| Field | Type | Required? | Notes |
+| - | - | - | - |
+| `"type"` | string | yes | Value must be `"multiscale"`. |
+| `"name"` | string | yes | Value must be a non-empty string. It should be a string that matches `[a-zA-Z0-9-_.]+`. Must be unique within one collections JSON file. |
+| `"path"` | string | yes | Value must be a string containing a path. [See paths section](#paths) |
```
lorenzocerrone: Hi @normanrz,

Is the `path` for a multiscale node really required?

In the example `custom-nodes.json`:

```json
{
  "type": "multiscale",
  "name": "color",
  "attributes": {
    "webknossos:category": "color",
    "webknossos:bounding_box": {
      "topleft": { "x": 128, "y": 128, "z": 128 },
      "size": { "x": 5445, "y": 8380, "z": 3285 }
    },
    "webknossos:data_type": "uint8"
  },
  "path": "/absolute/path/to/l4dense_motta_et_al_demo/color"
}
```

Here, the `attributes` do not contain any OME-NGFF metadata, so (if I understand correctly) the axes, datasets, etc. are expected to be found in
`/absolute/path/to/l4dense_motta_et_al_demo/color/zarr.json`.

In contrast, in the other example `inline-multiscale.json`, the OME-NGFF metadata is provided at the top level:

```json
{
  "type": "multiscale",
  "name": "segmentation",
  "attributes": {
    "webknossos:category": "segmentation",
    "webknossos:bounding_box": {
      "topleft": { "x": 0, "y": 0, "z": 0 },
      "size": { "x": 5632, "y": 8704, "z": 3584 }
    },
    "webknossos:data_type": "uint32",
    "webknossos:values": { "max": 100000 }
  },
  "multiscales": [
    {
      "axes": [
        { "name": "c", "type": "channel" },
        { "name": "x", "type": "space", "unit": "nanometer" },
        { "name": "y", "type": "space", "unit": "nanometer" },
        { "name": "z", "type": "space", "unit": "nanometer" }
      ],
      ...
    }
  ]
}
```

In this second case, there is no `path` for the node.

lorenzocerrone: (besides the paths in the datasets)

normanrz: There is still an open design question, whether we should allow inlining multiscales. If yes, the path will not be required anymore. My example was just an experiment, not normative.

rfc/8/index.md
jo-mueller: Could this work similar to how I proposed it for the coordinate transforms? In essence, the `path`s specified in the `plate` metadata could be allowed to contain a `Collection`, which would contan the reference to the path.

toloudis:
User story 1 :  combining independent segmentation zarrs with raw image zarrs. 

We produce multiscale zarrs of our raw microscope images, using filtered downsampling.
We later produce multiscale zarrs of segmentations of the above images, using unfiltered (nearest neighbor) downsampling.

In viewers we want to give users an easy way of combining the two.  In particular, our users are interested in seeing the data as if it were actually separate channels of the same volume.  This may or may not be a viewer implementation detail, but it could be interesting if the spec supported this, pointing to two separate zarrs and treating them as consecutive channels.  For our viewer, this only works if the spatial dimensions are the same, and can be transformed to the same origin (always  trivially true for the data I describe).

User story 2:  dataset releases

Is it practical to have one single very large collection?  as in 1000s of zarrs or more?   We would likely produce collections of matched raw+segmentation zarrs as described in my user story 2.

jo-mueller:
### TODO:

This post is a bit stream of consciousness-y - I hope I manage to express the bump I a stumbling over with the current state of transforms in here. In the version of this RFC, when `coordinateTransformations` were added, the `input` and `output` fields of transforms were supercharged to bei either of:
- the `name` of a coordinate system
- the `path` to a multiscales group or single zarr array
- and now, *additionally*, the `name` of a node instance.

In https://github.com/ome/ngff-spec/pull/117, this was made more explicit, so that these `input`/`output` fields explicitly declare these. In an RFC8 world, this would look like this:

```json
"input": {
  "path": "./scale0",
  "node": "node_name",
  "name": "coordinate_system_name"
}
```

And I think porting over this formalism is important, because instances of `nodes` (of whatever kind) may own multiple coordinate systems.

*This has implications*. In RFC8, the transforms for `Singlescales` nodes live in the singlescale metadata, which *may* not be the same metadata document as the multiscales metadata (unless the singlescale metadata is inlined). So currently, the transforms metadata for an inlined signlescale would look like this:

```json
{
    "ome": {
        "version": "0.x",
        "type": "collection",
        "name": "example",
        "attributes": {
            "coordinateSystems": [
                {
                  "id": "world",
                  "name": "world",
                  "axes": [...]
                }
            ]
        },
        "nodes": [{
            "name": "raw",
            "type": "multiscale",
            "nodes": [{
                "id": "raw_0",
                "type": "singlescale",
                "path": {
                  "type": "zarr",
                  "path": "./raw/0"
                },
                "attributes": {
                    {
                      "type": "scale",
                      "scale": [1, 1, 1],
                      "input": {
                        "path": "raw_0",
                      },
                      "output": {
                        "name": "world"
                        "node": "raw_0"
                      }
                    }
                  ]
                }
            }, ...]
        }, ... ]
    }
}
```

The question I'm stuck with now: If the Singlescale is **not inlined** - where does the `coordinateSystem` metadata for the coordinate system called `"world"` live? Should it live in the `attributes` of the `Multiscales` node? In that case, the output of the `coordinateTransformations` metadata that is located in the Singlescales nodes would have to point up to the Multiscales metadata, which is tricky for obvious reasons. The other way, would be to require an instance of `coordinateSystems` and `coordinateTransformations` metadata in the attributes of the Multiscales node. The transforms could either:
- Link to the coordinateSystems in the singlescale nodes via `identity` transforms
- duplicate the `coordinateTransformations` in the Singlescale nodes

I don't have a good idea about which to prefer, though.

lubianat (approved):

lubianat (commented):
(sorry, the approval was a misclick on GH mobile when hastily ok'ing https://github.com/ome/ngff-spec/pull/128)

will-moore (commented):
Just adding comments, but seems I have to create a review...

will-moore:
Seems that adding comments to the changes page isn't working for me at the moment. So I'll add some here

> Non-root `Node` objects SHOULD NOT have a `version` field and MUST NOT have a different `version` value than the root `Node`.

Many of the collections I would like to represent with this spec contain images of different OME-Zarr versions. E.g. the figure at https://ome.github.io/omero-figure/?file=https://gist.githubusercontent.com/will-moore/75a7f0de5be0f7b4202d5f0229cadcc9/raw/ngff_images_figure.json or the list of samples at https://idr.github.io/ome-ngff-samples/ so this would be a blocker for many use-cases.

> Singlescale: This new interface replaces the dataset metadata defined in the previous versions of the OME-Zarr specification.

I'm not sure what the motivation is for `Singlescale` nodes, and what it means to "replaces the dataset metadata". Presumably Multiscale images/nodes still use "datasets"?
What is the difference between a Multiscale image with a single dataset and a Singlescale? Do we need both?

normanrz:
Thanks @will-moore!

> > Non-root `Node` objects SHOULD NOT have a `version` field and MUST NOT have a different `version` value than the root `Node`.
> 
> Many of the collections I would like to represent with this spec contain images of different OME-Zarr versions. E.g. the figure at https://ome.github.io/omero-figure/?file=https://gist.githubusercontent.com/will-moore/75a7f0de5be0f7b4202d5f0229cadcc9/raw/ngff_images_figure.json or the list of samples at https://idr.github.io/ome-ngff-samples/ so this would be a blocker for many use-cases.

Collections will likely be a feature of OME-Zarr 1.0. I don't think it is reasonable to referentially include all previous versions of the spec in the 1.0 release because of the burden that would put on implementations. 

> > Singlescale: This new interface replaces the dataset metadata defined in the previous versions of the OME-Zarr specification.
> 
> I'm not sure what the motivation is for `Singlescale` nodes

The motivation for `Singlescale` nodes is to a) have a name for resolution levels within a multiscale (other name suggestions welcome!), b) allow images to be `Singlescales` without a multiscale (i.e. the metadata is in the zarr.json of an array instead of a group with a single item), c) make multiscales behave like collections (i.e. they have `nodes`). 

> , and what it means to "replaces the dataset metadata". Presumably Multiscale images/nodes still use "datasets"?

Multiscales are now collections of Singlescales. The field `datasets` (which imo is very overloaded) is replaced by `nodes`.

> What is the difference between a Multiscale image with a single dataset and a Singlescale? Do we need both?

Multiscales with a single Singlescale are not disallowed, but not required anymore. Users can just create Singlescales as Zarr arrays without the need for enclosing Zarr groups.

d-v-b:
> The motivation for Singlescale nodes is to a) have a name for resolution levels within a multiscale (other name suggestions welcome!)

could you define the term "image" to mean "a Zarr array", and "multiscale image" to mean "a collection of images at different levels of detail". Starting with the more basic thing (a single array) and defining the collection in terms of that seems better than starting with the collection (multiscales) and defining the more basic thing in terms of it.

will-moore:
It feels like we have been working on RFC-5 for a long time and have finally reached a consensus on transforms and scenes etc. But even before v0.6 is released we are proposing to re-work all that again (and other core concepts like Multiscales.datasets that have been around since v0.1).
I don't know what the timeline for this looks like but it feels like too much churn.

Are we saying that OME.zarr data v0.6 and earlier are not expected to be supported by tools that read v1.0 because they are too different? That would discourage adoption of OME.zarr v0.6 because it's sunsetted even before it's released. 

My first impression of RFC-8 was that it's a way of grouping existing Multiscales images into Collections. But this proposal looks like starting from scratch and ditching previous work and support for existing data?

I'm not even sure I fully understand @jo-mueller's question above, except that it shows all the hard RFC-5 discussions are going to need to be revisited again?

jo-mueller:
@will-moore thanks for the feedback. About my [comment above](https://github.com/ome/ngff/pull/343#issuecomment-4334403555), I think discussing intents and structure last week in Düsseldorf helped to structure my ideas for RFC8. I opened https://github.com/normanrz/ngff/pull/4 with some suggestions that address some of my concerns.

normanrz:
> It feels like we have been working on RFC-5 for a long time and have finally reached a consensus on transforms and scenes etc. But even before v0.6 is released we are proposing to re-work all that again (and other core concepts like Multiscales.datasets that have been around since v0.1). I don't know what the timeline for this looks like but it feels like too much churn.

I appreciate the design work that has gone into RFC-5 and I think RFC-8 is building on top of that. I'll review with @jo-mueller next week whether to bring back the scene metadata.

> My first impression of RFC-8 was that it's a way of grouping existing Multiscales images into Collections. But this proposal looks like starting from scratch and ditching previous work and support for existing data?

I think it is important to look at RFC-8 as part of the long-term vision of the 1.0 release. This probably warrants its own RFC, but in my view 1.0 is supposed to be a long-term release that carries us through the next decade without breaking changes. Up until now every release of OME-Zarr has been breaking and I think that needs to stop to foster serious adoption. That also means this is the last opportunity in a while to break things in order to make the OME-Zarr spec more consistent and extensible. Basically, take all the learnings from the 0.x releases and make a great long-term 1.0 release.

> Are we saying that OME.zarr data v0.6 and earlier are not expected to be supported by tools that read v1.0 because they are too different? That would discourage adoption of OME.zarr v0.6 because it's sunsetted even before it's released.

I definitely think that tools should be considered compliant with the 1.0-spec if they only support v1.0 and no previous versions. This is already the case with 0.x versions. Only very few tools understand 0.1-0.3 and some tools only understand 0.5 and not 0.4 anymore. I think that is totally fine, because they are 0.x releases.

That being said, I think the extension mechanism could be used to include 0.x OME-Zarrs in 1.0 Collections. Just define an extension node type that references 0.x multiscales. Tools could voluntarily support that, if they find it useful.

I want to add that 0.5 -> 0.6 -> 1.0 are metadata-only changes. I don't think it is unreasonable for users to consider migrating the metadata. This will be less of a lift than the 2024 NGFF challenge, where we actually converted the data.

d-v-b:
seconding norman's POV. And a broader point about churn: churn during development is valuable if it buys a better released product. This churn affects devs for months, but users will interact with 1.0 for _years_. It would be unfortunate if they had to tolerate a deficient product because devs settled too early. Now is the time to fix stuff. It only gets harder later.

jo-mueller:
I think this is a super-useful discussion here. If anything, it will help RFC8 authors to get a feeling from which direction to expect feedback or sharpen RFC8 towards. I think there are two separate things to take from this discussion:

Minimally, I think the relationship between coordinate system and nodes needs to be clarified. To a degree, this already happened in 0.6.dev3 -> 0.6.dev4. The important thing to note here is that coordinate systems and transformations define their own graph like structure, that can be **independent of the collection/node layout**. Since a `node` in RFC8 can cleanly be resolved to a `path`, the only crucial thing to update is how coordinate systems are referenced, which is indeed a result of a quite lengthy discussion, but it's not hard to fix. Transform inputs and outputs can refer to coordinate systems and the node inside which they live and that'll be fine.

The *other* thing is the following:

> 0.5 -> 0.6 -> 1.0 are metadata-only changes.

*I'm not so sure about that*. In 0.x, the smallest interpretable, indivisible aggregation of data and metadata is the `multiscales` specification. Any implementation of ome-zarr minimally needs to understand the concept of a multiscales; Without that, plates, wells or scenes don't make much sense. At the same time, the multiscales brings the functionality that allows viewers and other consumers to give users the smooth access to large data that we often advertise. Without the multiscales concept, this is simply not possible. 

The introduction of the `Singlescale` acknowledges individual resolution levels (aka zarr arrays) as part of the ome-zarr metadata, which changes what elements of a multiscales are understood to be "load-bearing" parts. Moving the `singlescale` metadata into the `zarr.json` of the individual multiscale arrays seems to me like those arrays are intended to be interpretable by themselves, irrespective of the context of a multiscales into which they are embedded. That, in requires consumers to be able to handle singlescale arrays, which I think many (i.e., viewers) cannot do easily. 

Don't get me wrong, I'm not opposed to renaming `datasets` into `nodes` where every entry in the array is an instance of a `Singlescale`. As @normanrz points out, that's indeed just a change of metadata. I'm skeptical about distributing this metadata into the `zarr.json` of several arrays, all of which should be interpretable without the context of a multiscale node that's defined elsewhere in the hierarchy.

What I propose in https://github.com/normanrz/ngff/pull/4 is simply a stratification and clarification of where metadata sits and what collections are expected to collect:
- Collections declare collections of **images** and potentially other stuff
- An **image** is a special kind of collection, consisting of one or more resolution levels.
- An image (aka multiscale group) forms the most elementary, portable, self-describing unit of a collection, that can be copied, referenced and validated independently.

This is currently not necessarily the case with the `Singlescale` node. RFC8 already suggests a possible way out: Inlining the singlescale metadata into the multiscale node sorts this out. I would just require multiscales to *always* inline singlescales.

Imho, making this restriction doesn't take from the expressiveness and elegance of RFC8, but adds to the integrity and reliability of images - aka multiscales - as an essential concept in the spec.

clbarnes (commented):
Without going too deep on the content, which can be discussed once the RFC is published: this is obviously a very deep change with a lot of new concepts spanning different areas of the spec and OME-Zarr usage. I think it would help to guide productive comments/ reviews if those (somewhat) distinct features could be enumerated near the top. For example:

- Node interface: a consistent JSON structure for several different types of OME-Zarr metadata object, where fields specific to the node type are inside an `attributes` field, and the root only stores information used for identifying and referencing the object.
- Collections: arbitrary collections of nodes which can be specialised for different use cases.
- Reworking existing structures to a Node/Collection-based framework:
  - Single-scale image arrays
  - Multiscale image groups
  - HCS layout
- Reference interface: a consistent system for referring to local and remote OME-Zarr metadata objects
- Extension system: namespacing within Nodes' `type`s and fields within their `attributes` to allow extension of the OME-Zarr by specific vendors or for specific use cases

These elements obviously touch each other but I think are distinct enough that they could support discussion in parallel and it would be good to bring attention to that as part of the overview.

normanrz:
Thanks @clbarnes! Your suggestions are pretty good! I adapted them slightly and added them as tl;dr to the proposal section.

rfc/8/index.md
```diff
@@ -6,4 +6,1951 @@
...
+* Reworking existing structures to a Node/Collection-based framework:
+
+  * Single-scale image arrays
+  * Multiscale image groups, bioformats2raw.layout
+  * Label maps
+  * HCS layout
+
+* Reference interface: a consistent system for referring to local and remote OME-Zarr metadata objects
+* Extension system: namespacing within Nodes' types and fields within their attributes to allow extension of the OME-Zarr by specific vendors or for specific use cases
+* Integration of coordinate transformations (RFC-5) metadata
```
clbarnes:
```suggestion
* [Node interface](#node): a consistent JSON structure for several different types of OME-Zarr metadata object, where fields specific to the node type are inside an attributes field, and the root only stores information used for identifying and referencing the object.
* [Collections](#collection-node): arbitrary collections of nodes which can be specialised for different use cases.
* Reworking existing structures to a Node/Collection-based framework:

  * [Single-scale image](#singlescale-node) arrays
  * [Multiscale image](#multiscale-node) groups, bioformats2raw.layout
  * [Label maps](#label-maps-and-other-derived-images)
  * [HCS layout](#high-content-screening-hcs-metadata)

* [Reference interface](#path-interface): a consistent system for referring to local and remote OME-Zarr metadata objects
* [Extension system](#extensibility): namespacing within Nodes' types and fields within their attributes to allow extension of the OME-Zarr by specific vendors or for specific use cases
* Integration of [coordinate transformations](#coordinate-transformations) (RFC-5) metadata
```

Added what I hope are the correct links to make it a bit easier to jump around, but I don't feel strongly about it.

clbarnes (approved):

jo-mueller (approved):

PR #364, opened 2025-11-12
feat(rfc-9): Add zip comment structure for central directory

mkitti:
Adds a new section to RFC-9 to describe the structure of the JSON comment in the zip archive. This comment provides metadata about the central directory, including:

- `jsonFirst`: A boolean flag to indicate if the `zarr.json` files are ordered breadth-first.
- <strike>`sorted`: A boolean flag to indicate if the remaining entries in the central directory are sorted.</strike>
-<strike> `duplicateResolution`: A string to specify how to handle duplicate filenames.</strike>

The text was written to only include the `jsonFirst` boolean key.

This metadata allows for more efficient parsing and validation of the zip archive.

### DIFF
--- a/rfc/9/index.md
+++ b/rfc/9/index.md
@@ -143,14 +143,9 @@ The zip comment is intended to provide metadata pertinent to the zip file struct
 
 The `ome` attribute in the zip archive comment MAY contain a `zipFile` attribute, which in turn MAY contain a `centralDirectory` attribute. The `centralDirectory` attribute provides metadata about the central directory's structure and content.
 
-The `centralDirectory` attribute MAY contain the following keys:
+The `centralDirectory` attribute MAY contain the following key:
 
-- `jsonBreadthFirst`: If `true`, this indicates that the `zarr.json` files are ordered breadth-first in the central directory and precede other content, as recommended above. This allows the hierarchical structure of the contents to be discovered without parsing the entire central directory, which could contain many entries of Zarr chunks. If this key is omitted, it is assumed to be `false`.
-- `sorted`: If `true`, this indicates that the entries in the central directory are sorted lexicographically by name. The root `zarr.json` MUST always be the first entry. If `jsonBreadthFirst` is `true`, this sorting applies to the non-`zarr.json` file entries. If `jsonBreadthFirst` is `false`, then all files are sorted, including the `zarr.json` files, with the exception of the root `zarr.json`. This assists with searching for keys in the directory via particular searching algorithms, such as binary search. If this key is omitted, it is assumed to be `false`.
-- `duplicateResolution` (string): Specifies which entry to use when duplicate filenames are present.
-    - `"first"`: Use the first occurrence of the duplicate filename.
-    - `"last"`: Use the last occurrence of the duplicate filename.
-    If this key is omitted, the default behavior is to use the `"last"` entry.
+- `jsonFirst`: If `true`, this indicates that the `zarr.json` files are ordered breadth-first in the central directory and precede other content, as recommended above. This allows the hierarchical structure of the contents to be discovered without parsing the entire central directory, which could contain many entries of Zarr chunks. Implementations MAY assume that no further `zarr.json` files exist beyond the first non-`zarr.json` file if `jsonFirst` is `true`. If `jsonFirst` is omitted, the value defaults to `false`.
 
 For example,
 ```json
@@ -159,9 +154,7 @@ For example,
     "version": "XX.YY",
     "zipFile": {
       "centralDirectory": {
-        "jsonBreadthFirst": true,
-        "sorted": true,
-        "duplicateResolution": "last"
+        "jsonFirst": true,
       }
     }
   }


jbms:
Every zip implementation I've seen, including the ones in neuroglancer and tensorstore, read the entire central directory at once and therefore a particular ordering of entries in the central directory does not provide any advantage when reading, significantly increases complexity for writing, and prevents using an existing zip implementation for writing.

jbms:
If the zip central directory is huge then your zip file is going to be difficult to use regardless of this metadata ordering, because to read a single chunk you will still need to read the entire central directory. 

However, you can always avoid a large central directory just by using (possibly nested) sharding.

mkitti:
After step 4, the user could make a decision that they are not interested in the contents of that zip file based solely on the contents of the zarr.json files that they can access via partial parsing of the central directory. Perhaps the user is browsing a collection of multiple zip files, only one of which contains the content of interest.

1. User writes zipped OME zarr data using $TOOL_A resulting in a zip file with a very large number of entries, and finds that reading it with $TOOL_B is slow or impossible.
2. User ask someone or looks at the OME zarr spec and finds this zip ordering metadata.
3. User implements or asks someone else to implement support in $TOOL_A and $TOOL_B for this zip ordering metadata.
4. User does some initial tests and finds that they can now read the metadata using $TOOL_B
5. User determines from the metadata using $TOOL_B that this is not the file of interest, discontinues processing of the file, and does not waste time processing the rest of central directory.

In the future, some future OME or Zarr metadata specification could refer to the existence of an external or embedded shard-like chunk index that memoizes the offsets and sizes of chunks in the zip file after its initial creation. Zip files can contain file entries that are not indexed in the central directory allowing for "hidden" files that may contain fixed length indexes that then enable chunk-based random access.

1. User writes zipped OME zarr data using $TOOL_A resulting in a zip file with a very large number of entries, and finds that reading it with $TOOL_B is slow or impossible.
2. User ask someone or looks at the OME zarr spec and finds this zip ordering metadata.
3. User implements or asks someone else to implement support in $TOOL_A and $TOOL_B for this zip ordering metadata.
4. User does some initial tests and finds that they can now read the metadata using $TOOL_B
5. User recognizes from the metadata that the zipped OME zarr also implements OME-Zarr RFC-99 encoding a base64 encoded shard-like fixed length index embedded in the zarr.json file for the array of interest
6. User implements or asks someone to implement OME-Zarr RFC-99 in $TOOL_B
7. User uses $TOOL_B to randomly access chunks in a shard-like manner via the retrofitted index

Providing easy access to metadata enables informed decision making about whether to proceed with further processing of the zip and allows for future extensions that may help overcome the limitations of the zip file central directory.

mkitti:
I created some single array test cases to examine the effect of the number of central directory entries:

https://s3.janelia.org/keller-lab/kittisopikulm/zip_test/

Each of these contain arrays of shape `[X, X, X]` where X is one of 128, 256, 512, 1024, 2048, 4096, or 8192. The chunk shape is `[32, 32, 32]`. Each element of the array has a value of `1` and is compressed using Zstandard.

At `X=8192`, we have `16,777,216` chunks within the archive.

Neuroglancer succeeds up to [1024](https://neuroglancer-demo.appspot.com/#!%7B%22dimensions%22:%7B%22dim_0%22:%5B1%2C%22%22%5D%2C%22dim_1%22:%5B1%2C%22%22%5D%2C%22dim_2%22:%5B1%2C%22%22%5D%7D%2C%22position%22:%5B0.5%2C256.5%2C313.5%5D%2C%22crossSectionScale%22:1%2C%22projectionOrientation%22:%5B-0.00556982820853591%2C-0.4551905393600464%2C-0.012357192113995552%2C0.8902908563613892%5D%2C%22projectionScale%22:512%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22https://s3.janelia.org/keller-lab/kittisopikulm/zip_test/test_array_v3_1024.zarr.zip%7Czip:%7Czarr3:%22%2C%22tab%22:%22source%22%2C%22name%22:%22test_array_v3_512.zarr.zip%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22test_array_v3_512.zarr.zip%22%7D%2C%22layout%22:%224panel-alt%22%7D) but then fails starting at [2048](https://neuroglancer-demo.appspot.com/#!%7B%22position%22:%5B0.5%2C335.3097839355469%2C407.5%5D%2C%22crossSectionScale%22:1.4484242512392533%2C%22projectionOrientation%22:%5B-0.00556982820853591%2C-0.4551905393600464%2C-0.012357192113995552%2C0.8902908563613892%5D%2C%22projectionScale%22:512%2C%22layers%22:%5B%7B%22type%22:%22image%22%2C%22source%22:%22https://s3.janelia.org/keller-lab/kittisopikulm/zip_test/test_array_v3_2048.zarr.zip%7Czip:%22%2C%22tab%22:%22source%22%2C%22name%22:%22test_array_v3_512.zarr.zip%22%7D%5D%2C%22selectedLayer%22:%7B%22visible%22:true%2C%22layer%22:%22test_array_v3_512.zarr.zip%22%7D%2C%22layout%22:%224panel-alt%22%7D).

I suspect there may be issues with the zip64 implementation. We may need to test zip64 across implementations to ensure this will work.

jbms:
With the fix, Neuroglancer handles the 4096 file. For the 8192 file, the encoded central directory is 900MB.  I should probably add a size limit on the central directory.

mkitti:
> With the fix, Neuroglancer handles the 4096 file. For the 8192 file, the encoded central directory is 900MB. I should probably add a size limit on the central directory.

Thank you for the quick fixes.

Do you think the autodetection step could be short-circuited?
Screenshot from 2025-11-23 23-28-37

PR #372, opened 2025-11-14
Ignore zar

jburel:
Ignore `zar` spelling leading to error


jburel:
Closing problem is already fixed using `` <!-- codespell:ignore -->``

PR #387, opened 2025-11-25
RFC-5: add review 1b

joshmoore:
Sent by mail

cc: @toloudis @dyf @fcollman


joshmoore:
Taking no objections from the authors as implicit approval as well. :smile:

Issue #386, opened 2025-11-25
Add the "NDPI issue" (lossy compression codecs not supported by Zarr) to the "When would I not use OME-Zarr?" section


jwindhager:
Currently, using OME-Zarr for data that has been compressed using a lossy codec not natively supported by Zarr inflates storage requirements (cf. the "NDPI issue"). IMO, it would be good to add this to the [When would I not use OME-Zarr?](https://ngff.openmicroscopy.org/#when-would-i-not-use-ome-zarr) section.

Unfortunately, I'm not too familiar with the scope of the issue myself, so its probably best if someone more experienced would add that section.

cc @govekk @SeanDuHare (I think)

clbarnes:
JPEG-XL is 4D (although IIRC good compression only happens in 2D + up to 3 channels) and can contain original JPEG buffers so it would be a good codec to target for this case. But library support is pretty poor and not improving.

PR #388, opened 2025-11-25
added line about ndpi bloat to main page

govekk:
Addressing original comment in #386 by adding a warning note to the main page "When would I not use OME-Zarr?" section about lossy compression


joshmoore:
Ok. Moving forward here then and @lubianat can propose a refactoring.

PR #391, opened 2025-12-03
Add Comment on RFC-9 based on workshop review

folterj:
RFC-9 comment 2

### DIFF
--- a/rfc/9/comments/2/index.md
+++ b/rfc/9/comments/2/index.md
@@ -1,0 +1,25 @@
+# RFC-9: Comment 2
+
+(rfcs:rfc9:comment2)=
+
+## Comment authors
+
+This comment was written by: Joost de Folter, BioImaging-NL
+
+## Conflicts of interest (optional)
+
+None
+
+## Summary
+
+We have implemented a validator and a simple implementation using Zarr-python [validation](https://github.com/jwindhager/ngff-rfc9-zipped-ome-zarr/tree/main/playground/validation).
+
+## Minor comments and questions
+
+- Explicitly state that only Zarr v3 is supported
+- Can one use ZIP64 on files that don't require it? It appears that ZIP64 records only exist in zip files that pass the 64bit length boundary.
+- https://github.com/python/cpython/issues/47073 - conceptually, data and metadata needs to be written together (add to implementation section)
+
+## Recommendation
+
+We recommend accepting this RFC with the suggested clarifications above.


rfc/9/comments/2/index.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Summary
+
+We have implemented a validator and a simple implementation using Zarr-python [validation](https://github.com/jwindhager/ngff-rfc9-zipped-ome-zarr/tree/main/playground/validation).
+
+## Minor comments and questions
+
+- Explicitly state that only Zarr v3 is supported
+- Can one use ZIP64 on files that don't require it? It appears that ZIP64 records only exist in zip files that pass the 64bit length boundary.
+- https://github.com/python/cpython/issues/47073 - conceptually, data and metadata needs to be written together (add to implementation section)
```
mkitti, line 21: I am not sure if you are asking if there a technical ability or otherwise.

Note that there is a `force_zip64` flag in the zipfile Python standard library:
https://github.com/python/cpython/blob/aea5531583aaa8bfdf3ebca914e9c694617c3489/Lib/zipfile/__init__.py#L1684
https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.open

rfc/9/comments/2/index.md
```diff
@@ -0,0 +1,25 @@
...
+
+None
+
+## Summary
+
+We have implemented a validator and a simple implementation using Zarr-python [validation](https://github.com/jwindhager/ngff-rfc9-zipped-ome-zarr/tree/main/playground/validation).
+
+## Minor comments and questions
+
+- Explicitly state that only Zarr v3 is supported
```
mkitti, line 19: OME-NGFF v0.5 is only specified on Zarr v3. Presumbably, if the RFC is accepted would be incorporated into OME-NGFF v0.7 or greater.
https://ngff.openmicroscopy.org/0.5/index.html#storage-format

Do you think the language above needs to be clarified further?

rfc/9/comments/2/index.md
joshmoore: Confirmed that this is re: technical ability with Joost. I'll update the comment.

rfc/9/comments/2/index.md
joshmoore: Confirmed; now moot. I'll remove. Thanks, @mkitti and @folterj

PR #408, opened 2026-01-07
change layout and add navbar

### LINKED ISSUE #407: Change website theme
Currently, the page's used theme is `sphinx_book_theme`. My proposal is to change this to `pydata_sphinx_theme`, as it comes with some really neat built-in customization options that can simply be specified in the `conf.py`. For instance, just adding this option:

```python
html_theme_options = {
  "header_links_before_dropdown": 4
}
```

Creates a nice layout with a navbar at the top, which I think goes a long way in adding more structure to the page:

https://github.com/user-attachments/assets/412ef535-87f3-4439-8afa-0ee651adf387

We can at some point also think about adding a fancier, more html-heavy frontpage with a boxed layout à la [scikit-learn](https://scikit-learn.org/stable/), [ro-crate](https://www.researchobject.org/ro-crate/), [napari](https://napari.org/stable/), etc. This could be a cool place to point to some highlights of things you can do with ome-zarr.

jo-mueller:
Fixes #407

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -53,7 +53,7 @@ redirects = {
 html_theme = "pydata_sphinx_theme"
 
 html_theme_options = {
-#    "header_links_before_dropdown": 4,
+    "header_links_before_dropdown": 6,
     "icon_links": [
         {
             "name": "GitHub",


lubianat (approved):

clbarnes:
Do you have control over what ends up in the More section of the navbar? When full-screened on my 15in, there's quite a lot of horizontal space to spare but Resources and Help Desk end up behind an extra click.

jo-mueller:
Ok. I had to set 

```python
"header_links_before_dropdown": 4,
```

but this makes all the navbar items appear without the `More` dropdown.

cc @erickmartins and @SeanDuHare if you want to have a look at it.

PR #410, opened 2026-01-08
merge contribution guide contents

### LINKED ISSUE #405: Move `Contributing` section and contribution guide into the same place
Currently, there are two places where contributions are explained:

- `/contributing/index.md`
- `/CONTRIBUTING.md`

I think these should be in the same place. For me this is relevant regarding #399, but I think it also makes the place tidies aside from that.

jo-mueller:
Fixes #405 

This moves all the information about contributing into a single place. I can see more information on how to contribute become necessary when #404 becomes a reality, but for now I think this is just a good clean-up step and help keep the repo tidy.

### DIFF
--- a/contributing/index.md
+++ b/contributing/index.md
@@ -22,9 +22,20 @@ please leave a comment in a new page under the "comments/" directory for the
 relevant RFC. A template is also available for formatting your comment:
 [template](../rfc/1/templates/review_template).
 
+### How to change the specification
+
+To make a change such as the end-product of an RFC or a minor correction to the currently developed specification version,
+please do so via a pull request (PR) against the `main` branch of the [ome/ngff-spec](https://github.com/ome/ngff-spec) repository.
+This repository contains the specification documents in Markdown format,
+json examples and the schema files for validation.
+
 ## Changes to this website
 
-The repository for this website is [ome/ngff](https://github.com/ome/ngff). Contributions to these pages are welcome as issues or PRs.
+The repository for this website is [ome/ngff](https://github.com/ome/ngff).
+Contributions to these pages are welcome as issues or PRs.
+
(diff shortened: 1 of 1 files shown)


contributing/index.md
```diff
@@ -24,4 +24,34 @@ relevant RFC. A template is also available for formatting your comment:
...
+### Previews
+Each PR receives a unique preview URL of the format `https://ngff--<PR#>.org.readthedocs.build/` where `<PR#>` is the PR number. This link is also posted to each PR by the Github actions bot in an "Automated Review URLs" comment as the "Readthedocs" link.
+
+Please check that your changes render correctly at this URL. New commits will automatically be live at the PR url after a few minutes.
+
+## Contributing to OME
+(contributing:ome)=
+
+Detailed guidance for contributing developers is available at
+https://docs.openmicroscopy.org/contributing/
```
lubianat: This URL is not hyperlinked in the website, maybe it needs markdown formatting

contributing/index.md
```diff
@@ -24,4 +24,34 @@ relevant RFC. A template is also available for formatting your comment:
...
+
+### The Quick Version
+
+* Fork the repository on GitHub
+* Create a branch for your work based on the latest `dev_x` e.g. dev_5_0
```
lubianat: is this still accurate? maybe we could use this PR to (at least) document that the exact workflow is being refined

contributing/index.md
jo-mueller: Fixed in a923a25e26c7e90090b7463b0e4a1310450b3730

contributing/index.md
jo-mueller: Yeah...probably hasn't been correct since the introduction of the submodules.

PR #420, opened 2026-01-14
RFC-3: comment 2

clbarnes:
tl;dr accept

### DIFF
--- a/rfc/3/comments/2/index.md
+++ b/rfc/3/comments/2/index.md
@@ -1,5 +1,7 @@
 # RFC-3: Comment 2
 
+(rfcs:rfc3:comment2)=
+
 ## Comment authors
 
 Chris Barnes <chris.barnes@gerbi-gmb.de>


rfc/3/reviews/1/index.md
```diff
@@ -0,0 +1,38 @@
+# RFC-4: Review 1
```
will-moore: RFC-3?

clbarnes: Whoops. Had already started thinking about #421 ...

rfc/3/comments/1/index.md
```diff
@@ -0,0 +1,38 @@
+# RFC-3: Review 1
+
```
jo-mueller, line 2: Minor suggestion (unrelated to content): In rfc5, I added reference anchors in the reviews like so:

```markdown
(rfcs:rfcX:reviewY)=
```

which makes it easy to cross-reference across the ngff page in a controlled manner. But since this comment is likely going to be the end of a discussion rather than the start of it, you might as well ignore this :)

PR #422, opened 2026-01-15
RFC1: Change COMMENTS to REVIEWS

lubianat:
Fixes unclear wording on the diagram. Also bumps figure version to 3.1.

The concept of "Comments" is reserved to a different thing in RFC 1. 

Related to https://github.com/German-BioImaging/ome-zarr-ideas/issues/55


joshmoore:
Thanks, @lubianat!

PR #428, opened 2026-01-20
rfc4: clarify coordinate direction convention

thewtex:
anterior-to-posterior is from negative to positive along the axis.
negative to positive along the axis is defined by its spatial metadata.

Requested: https://forum.image.sc/t/rfc-4-updates-enhanced-anatomical-orientation-support-for-ome-ngff/115625/6


thewtex:
@melonora @dyf @btbest please take a look :pray:

btbest:
Good addition, clearly phrased imo, looks good! 👍

PR #432, opened 2026-01-30
RFC-9 review: Curtis Rueden

ctrueden:
Here are my thoughts on RFC-9. Hope it helps, and happy to discuss further as needed!


joshmoore:
Thank you, @ctrueden!

PR #435, opened 2026-02-05
RFC 4: review from David Horsfall/ Hannifa lab

clbarnes:
This review was submitted externally; I am not the author but will manage its passage through github.

### DIFF
--- a/rfc/4/reviews/3/index.md
+++ b/rfc/4/reviews/3/index.md
@@ -1,0 +1,46 @@
+# **RFC-4: Review 3**
+
+(rfcs:rfc4:review3)=
+
+* [https://github.com/ome/ngff/pull/253](https://github.com/ome/ngff/pull/253)
+* [https://ngff.openmicroscopy.org/rfc/4/](https://ngff.openmicroscopy.org/rfc/4/)
+* [https://ngff.openmicroscopy.org/rfc/1/templates/review\_template.html](https://ngff.openmicroscopy.org/rfc/1/templates/review_template.html)
+
+**Review Authors**: Dave Horsfall
+**Conflicts of Interest**: None declared
+
+This review was primarily generated through discussions during a Hannifa Lab meeting, which involved diverse roles, including wet lab scientists, clinicians, bioinformaticians, data scientists, and engineers. I have tried to capture the key points of discussion in this review.
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


rfc/4/reviews/3/index.md
```diff
@@ -0,0 +1,46 @@
...
+
+The current proposal seems primarily geared toward imaging where the subject is a whole patient. In spatial transcriptomics and pathology, our subject is frequently a tissue biopsy, or histology slide.
+
+* **The Gap:** The RFC states this metadata "MUST only be used... where the subject is roughly aligned to the imaging axes." In our use cases, we may not know how the biopsy was aligned to the patient, but we do know the internal orientation of the tissue itself.
+* **Recommendation:** Clarify that "subject" can refer to local tissue structures. For a skin biopsy, the z-axis is "Superficial-to-Deep" regardless of the sample's original global position on the donor. Supporting local orientation makes this RFC universally applicable to the tissue-profiling community without requiring full "Patient-to-Atlas" registration.
+
+### **Controlled Vocabulary Expansion**
+
+The current vocabulary has a focus on biped/quadruped canonical directions. To support clinical and other research contexts (e.g., dermatology, cardiology, and oncology), we recommend adding terms that describe layered and polarized tissues to controlled vocabulary:
```
clbarnes, line 29: The current RFC gives the orientation a "type"; for now, the RFC only defines "type":"anatomical". The examples given below are good candidates for additional types with a separate namespace.

In my opinion, the namespacing means that this feedback should not block the acceptance of the RFC as-is; if reasonably complete vocabularies can be provided for other orientation types then they could be added to RFC-4, or if it takes longer/ a more in-depth survey to gather those vocabularies, they can be added in another RFC or via some (yet-to-be-proposed) extension mechanism.

thewtex (approved):
Thank you! This is an excellent review.

I'll improve the text and enumerated values in the future.

thewtex:
@melonora I think they should all live in the same enum. Given the diversity of creatures out there, there is not going to be clear and universal buckets.

thewtex:
Yes, and I think we did a good job of allowing other orientation types.

While there may be shared orientation values across types, e.g. "left-to-right" for an anatomical type and a material type, it is fine (and clearer and easier to implement from my perspective) if they are explicitly listed in each type. And left-to-right is likely to have a different meaning when used outside the anatomical context. It means relative to the subject/specimen with anatomical. In other contexts left-to-right is likely relative to something else.

PR #508, opened 2026-05-01
fix #507

### LINKED ISSUE #507: Vizarr embedding does not work correctly on some Chrome builds
Chrome: Version 147.0.7727.138 (Official Build) (arm64)

It seems to be a window size-related issue, and it fixes itself by changing the window size.

Fullscreen:
Image
Resized:
Image

lorenzocerrone:


lubianat:
Thank you, @lorenzocerrone! I'll see if this works in multiple devices. If yes, we merge. Otherwise, we take out the fish  as a hotfix and work on a way..  

A extra motivation is that the "science stories" (https://github.com/German-BioImaging/ome-zarr-ideas/issues/94) project (i.e. writing motivations at the bio-scientist level with browsable images) needs reliable embedding

PR #509, opened 2026-05-01
Use biongff vizarr

lubianat:
* Changes the URL for the Vizarr embedding
* Changes the ~500GB 2D fish for a ~200GB 4D brain  (~185, summing up all scales via validator) 

Supersedes #508


lubianat:
It does not work on my phone, but it was not working before either.  It works on other phones. Likely a different bug.

lorenzocerrone:
> It does not work on my phone, but it was not working before either. It works on other phones. Likely a different bug.

I can confirm, both this PR version and the main version do not work on my phone (iPhone).

joshmoore:
A few things:
 * Note the change of the submodules
 * Generally supportive of using the new URL wherever we *do* embed :+1:
 * Might either `@`-mention the team here if there's not an open issue
 * Moving forward with #511 for the moment though

PR #511, opened 2026-05-12
Remove the Vizarr display

lubianat:
Reverting #483 while a robust fix for #507 is not found. It is really not great to show a broken viewer in the first part of the website (sorry for that!). 

The fix on #509  is only partial; it also fails for some mobile devices (as noted in #507). 

cc @lorenzocerrone

### DIFF
--- a/index.md
+++ b/index.md
@@ -2,11 +2,7 @@
 
 Welcome to the Next-Generation File Formats (NGFF) main page! This site is dedicated to providing resources for the NGFF community and those that are interested in getting started with OME-Zarr.
 
-In summary, OME-Zarr allows you to store, share and stream large images like these 500 GB:
-
-<iframe style="width: 100%; height: 500px" name="vizarr" src="https://hms-dbmi.github.io/vizarr/?source=https://livingobjects.ebi.ac.uk/idr/zarr/v0.1/4495402.zarr"></iframe>
-
-You can find more samples of such data from the Image Data Resource at <https://idr.github.io/ome-ngff-samples/>.
+In summary, OME-Zarr allows you to efficiently store, share and stream large bioimages.
 
 ## NGFF vs OME-Zarr, what is the difference?
 
--- a/index.md
+++ b/index.md
@@ -2,7 +2,7 @@
 
 Welcome to the Next-Generation File Formats (NGFF) main page! This site is dedicated to providing resources for the NGFF community and those that are interested in getting started with OME-Zarr.
 
-In summary, OME-Zarr allows you to efficiently store, share and stream large bioimages.
+In summary, OME-Zarr allows you to store, share and stream large images. You can browse samples of such data from the Image Data Resource at <https://idr.github.io/ome-ngff-samples/>.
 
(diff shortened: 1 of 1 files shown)


lubianat:
@will-moore would you mind sharing your thoughts on this quickly?

index.md
```diff
@@ -2,11 +2,7 @@
 
 Welcome to the Next-Generation File Formats (NGFF) main page! This site is dedicated to providing resources for the NGFF community and those that are interested in getting started with OME-Zarr.
 
-In summary, OME-Zarr allows you to store, share and stream large images like these 500 GB:
-
-<iframe style="width: 100%; height: 500px" name="vizarr" src="https://hms-dbmi.github.io/vizarr/?source=https://livingobjects.ebi.ac.uk/idr/zarr/v0.1/4495402.zarr"></iframe>
-
-You can find more samples of such data from the Image Data Resource at <https://idr.github.io/ome-ngff-samples/>.
```
clbarnes, line 9: Why remove the IDR link? It would be nice to give a link to some (viewable) data.

will-moore:
I have tried to reproduce the bug with all the options on #507 but have so far failed (Chrome Version 147.0.7727.102 on Mac). So it's tricky for me to understand where this bug is coming from.
If I change my page zoom then I sometimes see the image disappear (moves out of the viewport) but I think this is because the current viewport position (maintained in the iframe vizarr url) no-longer is suitable for the resized viewer. This is always fixed for me with a page refresh.
It's not clear to me how the bug demos on #507 would be affected by a page refresh after zooming/resizing the page? I guess the original bug report was not due to a page resize/zoom.

jo-mueller:
My feeling would be that while it's of course problematic that it doesn't work on mobile devices, it's kind of not *that* bad as long as it displays on other devices?

index.md
lubianat: It was added alongside the fish, I just removed it by default. Happy to keep it there.

lubianat:
@will-moore there are two bugs detected: 

* One desktop issue with Chrome and particular Zoom levels, fixed on https://github.com/BioNGFF/vizarr. It led not only to the image disappearing, but to the canvas "twitching". I could reproduce on multiple pages embedding the hms Vizarr page.  

* One issue on a few mobile devices, which affects at least my device. 

As the issue occurs also when going to Vizarr and, say, the new BIA website (ping @matthewh-ebi), I would remove the image from the main page as a hotfix while things are not solved on the Vizarr side.

joshmoore:
Supportive of @lubianat's comment this morning that he won't be able to drive this. Removing for the moment, but very supportive if/when someone else wants to take on the investigation and/or assert that it's been fixed upstream. Thanks!

cc: @davehorsfall

jo-mueller:
Interesting...I still the viewer on the landing page?

lubianat:
@jo-mueller @joshmoore 

Apparently there was some unknown issue with the RTD build, though the preview (https://ngff--511.org.readthedocs.build/) worked fine.

grafik

I wonder if the build can just be re-triggered on RTD

PR #517, opened 2026-05-22
docs: add version history to spec page build

### LINKED ISSUE #515: Make RFC-5 page up to date
https://ngff.openmicroscopy.org/rfc/5/index.html has (at least!) two improvements that could be made:

- The status is out of date (I think the RFC is accepted??)
- The changelog is empty. It should be ideally updated, but also pointing to https://ngff-spec.readthedocs.io/en/latest/version-history/ as where the up-to-date changelog is would be very helpful.

jo-mueller:
Fixes https://github.com/ome/ngff/issues/515

This adds the version history file from the ngff-spec page to the ngff page and updates the version history on the rfc5 page. I hope it also clarifies that the last version associated with rfc5 is 0.6.dev3, after which came some smaller changes. Those are not represented by an RFC, but still listed in the version history in more detail.

### DIFF
--- a/rfc/5/index.md
+++ b/rfc/5/index.md
@@ -14,7 +14,7 @@ Add named coordinate systems and expand and clarify coordinate transformations. 
 
 ## Status
 
-This RFC is currently in RFC state `R4` (authors prepare responses).
+This RFC is currently in RFC state `S4` (Update implementations).
 
 | **Role** | Name | GitHub Handle | Institution | Date | Status |
 |----------|------|---------------|-------------|------|--------|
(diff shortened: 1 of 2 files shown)


sbesson (dismissed):
Thanks @jo-mueller, I was looking for this page under https://ngff.openmicroscopy.org/specifications/ as part of https://github.com/ome/ngff-spec/pull/131. 

If this is the new standard location for everything related to the specification, having the version history as a top-level section makes sense

lubianat:
@jo-mueller would you mind adding here the current status of the RFC? 

I guess we are in  "S3 - Update implementations"?

jo-mueller:
@lubianat Welp, I guess the state RFC5 was associated with was indeed to update the RFC5, but it's a bit confusing since there is no rfc5 version that is associated with 0.6.dev4. I hope the changelog at the bottom of the rfc5 document clears this up!

PR #545, opened 2026-06-23
Minor: update status of RFC-4

DragaDoncila:
This PR updates the status of RFC-4 to R4 (authors prepare response) and lists the two reviewers and their reviews in the table.

cc @joshmoore


joshmoore:
ty! 🙌🏽

PR #552, opened 2026-07-01
Restructure RFC documents to use a YAML front matter for Status and author/reviewer/commenter metadata

lubianat:
First step towards a system with more automation/standardization for rfc status management. 

https://github.com/German-BioImaging/ome-zarr-ideas/issues/54

- **feat: build authors from yaml header**
- **test setup for rfc 9 comment 5**

grafik

Next steps:

- [x] Change the metadata for other reviews / comments
- [x] Build the summary on the RFC main page from the index 
- [x] Build the overall index from the metadata on each page using a decision tree (has comment -> y, etc etc)
- [x] Validate the front matter in target documents with a LinkML schema

EDIT: Ready to review! 

To review: 

* Compare the generated RFC pages with the older one. 
* Provide feedback on the new look and feel and if changes are needed
* Compare the page listing the different RFCs with the older one
* If checking all 57 documents is too much, pick a sample and see if it is okay

* If you are a schema person, look at `rfc/schema/front_matter.yaml` for the LinkML schema
* If you are a sphinx person, look at `_ext` and `document_authors.py` and `rfc_status.py` and see if the directives make sense

Note that a lot of the code and changes have been aided by a Claude Opus 5 agent, though I tried to steer for minimal changes and maximize readability.


jo-mueller:
Might benefit from a template, i.e. for RFCs and comments?

jo-mueller:
Also; it may be worth to drop the affiliations from the rendered text, similar to how it's done at ngff-spec. The reasoning there was, that a person's orcid is likely informative enough about past and present affiliations.

rfc/9/comments/5/index.md
```diff
@@ -1,14 +1,28 @@
+---
+authors:
+  - name: Anna Kreshuk
+    orcid: https://orcid.org/0000-0003-1334-6388
+    affiliation: ilastik
```
gouttegd, line 5: Two different suggestions here:

(A) Use a project URL (which in this case would be `https://www.ilastik.org/`) instead of a single name. This would make absolutely clear which project is referenced, and would also help to avoid using a slightly different project name across the various comments (for example, `ilastik` in one comment, `Ilastik` in another).

(B) Alternatively, use a dedicated structure for “affiliation”, allowing to explicitly specify a project name, a URL, and possibly other informations. Something like:

```yaml
authors:
  - name: Anne Kreshuk
    orcid: https://orcid.org/0000-0003-1334-6388
    affiliation:
      - name: Ilastik
        url: https://www.ilastik.org/
```

rfc/9/comments/5/index.md
```diff
@@ -1,14 +1,28 @@
...
+  - name: Dominik Kutra
+    github: k-dominik
+    orcid: https://orcid.org/0000-0003-4202-3908
```
gouttegd: Nitpick (which means, “feel free to ignore” ;) ): It annoys me that the GitHub identifier is a single name, while the ORCID identifier is a URI.

I’d rather have, either all URIs:

```
github: https://github.com/k-dominik
orcid: https://orcid.org/0000-0003-4202-3908
```

or all non-URIs:

```
github: k-dominik
orcid: 0000-0003-4202-3908
```

lubianat: good suggestion. I'll follow https://github.com/ome/ngff-spec/blob/main/authors.yml and keep only the id

rfc/9/comments/5/index.md
lubianat: I like this in theory for removing ambiguity, but some trade offs are

(1) it would create a curation burden for me and (2) add some more complexity to the model 
 
I want to keep this very minimal and only re-shape information that is already present on the documents

gouttegd: Fair enough! :)

lubianat:
> a person's orcid is likely informative enough about past and present affiliations.

in theory, yes, it should.  in practice, I need to keep the information provided by the authors on under which affiliation that comment was made. 

an example is the rfc 9 / comment 5

Anna Kreshuk - ilastik was the intended affiliation, though that is a project and not an employee, and not listed on https://orcid.org/0000-0003-1334-6388

joshmoore:
Did you mean to bump `specifications/dev`?

rfc/0/index.md
```diff
@@ -1,3 +1,18 @@
...
+    github: joshmoore
+    affiliation: German BioImaging, e.V.
+    affiliation_url: https://ror.org/05tpnw772
+    role: Author
+    date: "2024-08-30"
+manual_status: N/A
+status_note: historical RFC, outdated by RFC-1
+description: Original consensus model for decision making
+ome_zarr_version: "N/A"
+date: 2024-08-30
```
gouttegd, line 13: LinkML-defined, date-typed slots should always be written as strings in YAML serialisation. That is, they should be quoted (`date: "2024-08-30"`).

Not quoting them will make the YAML parser parse the value into a `date` object, which may cause all sorts of unexpected problem down the line (see https://github.com/linkml/linkml/issues/1532 for some details).

_ext/document_authors.py
```diff
@@ -0,0 +1,92 @@
...
+    def run(self):
+        env = self.state.document.settings.env
+        src = env.doc2path(env.docname)
+        with open(src, encoding="utf-8") as f:
+            text = f.read()
+
+        parts = text.split("---", 2)
+        if len(parts) < 3:
+            raise self.error("rfc-authors: no YAML front matter found")
+        meta = yaml.safe_load(parts[1]) or {}
```
gouttegd, line 32: Just so you know, since you have a LinkML schema, you could also use it to generate a `RFCFrontMatter` Python object to deal more easily with the contents of the front matter.

That is, you could produce a `frontmatter.py` Python module:

```console
$ linkml generate pydantic front_matter.yaml > front matter.py
```

and then in this script:

```python
from frontmatter import RFCFrontMatter
from linkml_runtime.loaders import yaml_loader

[...]
meta = yaml_loader.load(parts[1], RFCFrontMatter)
```

The `meta` variable now contains a `RFCFrontMatter` object whose members you can use directly, instead of having to query the dictionary to get each key:

```python
affils, order = {}, []
for author in meta.authors:
    if author.affiliation and author.affiliation not in affils:
        order.append(author.affiliation)
        affils[author.affiliation] = len(order)
```

And so on.

Not requesting any change because your way of course also works, but it does feel strange (to me at least!) to be using LinkML only halfway through like that. :)

_ext/rfc_status.py
```diff
@@ -0,0 +1,395 @@
...
+import yaml
+from docutils import nodes
+from docutils.parsers.rst import Directive
+from sphinx import addnodes
+
+# Display-only labels for the state codes of the (non-normative) table under
+# resources/rfc-status-codes. The code in an RFC's front matter stays the source
+# of truth; this only spells it out for readers.
+STATE_LABELS = {
+    "D1": "Initial idea",
```
gouttegd, line 12: This is another thing that a LinkML-generated Python module could do for you. :)

rfc/schema/front_matter.yaml
```diff
@@ -0,0 +1,196 @@
...
+        description: Contact e-mail address.
+      affiliation:
+        description: Institution the person is credited under, as plain text.
+      affiliation_url:
+        description: Link for the institution, e.g. a ROR record or its homepage.
+        range: uri
+      role:
+        description: >-
+          The person's role on this RFC, shown in the "Status" column, e.g.
+          "Corresponding Author", "Co-author", "Editor" or "Implemented".
```
gouttegd, line 53: Is that supposed to a free-text value? Or is there a clearly defined list of roles that people should pick from? If the latter, this could be an `enum`.

rfc/schema/front_matter.yaml
```diff
@@ -0,0 +1,196 @@
...
+          PR's merge. Its year is the listing's "Date" column.
+        range: date
+      reference_pr:
+        description: The pull request that introduced the RFC.
+        range: uri
+      authors:
+        description: The RFC's authors, in the order they should be credited.
+        range: Person
+        multivalued: true
+        inlined_as_list: true
```
gouttegd, line 100: Just FYI, this is not necessary: `Person` does not have either a “key” slot or an “identifier” slot, so it can _only_ be inlined as a list, regardless of whether you say that explicitly or not.

(Though I do myself like to always make that explicit, as it makes the intention clearer. ;) )

rfc/schema/validate.py
```diff
@@ -0,0 +1,98 @@
...
+        return _as_strings(yaml.safe_load(parts[1]) or {})
+    except yaml.YAMLError as error:
+        return f"front matter is not valid YAML: {error}"
+
+
+def _as_strings(value):
+    """Turn the dates PyYAML parses back into ISO strings.
+
+    An unquoted `date: 2025-07-02` becomes a datetime.date, while a quoted one
+    stays text; the pages treat both the same, so validation should too.
```
gouttegd, line 55: I would much rather consistently use quoted dates rather than having such a workaround…

rfc/0/index.md
gouttegd: OK, I see you have an explicit workaround in `validate.py` for that. I don’t like it, but well, it works. :)

gouttegd (commented):
Just commenting for now. I have mostly looked at the LinkML stuff, _not_ the Python code in details.

rfc/schema/front_matter.yaml
lubianat: I think free text for now, there is no strong "role" definition in the RFC process and maybe an `enum` is overspecifying

_ext/document_authors.py
lubianat: Oh, interesting! The LinkML schema came after `document_authors.py`, that is why the piecer are not connected, so a refactoring may make sense indeed.

lubianat:
> Just commenting for now. I have mostly looked at the LinkML stuff, _not_ the Python code in details.

That is excellent, thank you. I'll actually refactor some of it as you suggested. 

As I mentioned in the PR description, there is a lot that was generated with Claude assist. Reviewed, but still, it is a big change, so I am not sure how to make the review process less painful.

jo-mueller:
Something that bumps me mildly is that the author information is rendered differently for the RFC document itself (the old table) and the reviews/comments/etc, where the frontmatter yaml is used.

I think jupyter-book or myst-md are also able to do this based on frontmatter yaml that looks very similar to what we have here. If you check the auto-built metadata examples and schema markdown files that are generated by the `pre-build.py` over at the ngff-spec repo, it will generate a frontmatter like this:

```
---
---
```

to make sure that we don't render the author information on every page of the examples and schemas section ^^"

lubianat:
> Something that bumps me mildly is that the author information is rendered differently for the RFC document itself (the old table) and the reviews/comments/etc, where the frontmatter yaml is used.

@jo-mueller I think I don't get exactly what you mean here. Would you mind rephrasing your suggestion?

Issue #426, opened 2026-01-19
Sort RFCs from highest number to lowest on website


lubianat:
As new RFCs appear in the future, it may make sense to display the most recent (i.e. high number) on top, so navigation to active RFCs is easier

<img width="977" height="811" alt="Image" src="https://github.com/user-attachments/assets/8e6c6ec7-865d-48fa-b5a0-70b327c57ac0" />

basically just inverting the order on the table and on the Section Navigation bar

joshmoore:
:+1:

jo-mueller:
Maybe even mark the finished ones accordingly in the navigation section? I.e. with a checkmark emoji?

lubianat:
That is interesting. But another maintenance point 🫠

So maybe one for #552 to address programatically

PR #559, opened 2026-07-03
RFC-10: propose NGFF editorial board

joshmoore:
As agreed during the OME2026 community meeting. For review by the OME Management Group.

see: https://bit.ly/ome2026-gov-proposal

cc: @vuhlmann

### DIFF
--- a/rfc/10/index.md
+++ b/rfc/10/index.md
@@ -151,6 +151,7 @@ The EB as a whole:
   * Having team members listed as co-authors on RFC reviews etc.
   * One editor as a clear lead on each RFC
   * Include commit right concepts that would be in other charters
+* Coordinate and communicate with the wider Zarr community and the Zarr Steering Council (ZSC).
 
 EB Members are expected to:
 
--- a/rfc/10/index.md
+++ b/rfc/10/index.md
@@ -125,12 +125,35 @@ requirements have been satisfied. Such intervention should be exceptional and
 should not substitute the OMG's judgement for that of the Board on technical
 matters.
 
-The OMG does not:
+The OMG shall not:
 
 * Vote on routine RFC decisions.
 * Direct the technical content or editorial decisions of the Board.
 * Routinely review or approve RFC decisions made by the Board.
 * Override Board decisions except in formally escalated cases.
+
+### Lazy consensus and OMG oversight
+
+Matters referred to the OMG for review shall normally follow a lazy-consensus
+process: the Board may proceed unless a substantive objection is raised. The
+OMG should make reasonable efforts to review such matters in a timely manner.
+
+The target review period is one full working week, with an outer limit of two
+full working weeks. If no substantive objection has been raised within the
+outer limit, the Board may proceed. Periods of individual OMG member
+unavailability do not automatically extend these limits.
+
+A substantive objection shall identify the governance, procedural, or other
+basis for intervention. The OMG's oversight shall not substitute its judgement
+for the technical authority delegated to the Board.
+
+The following require affirmative OMG action and are not subject to lazy consensus:
+
+* Appointment or removal of Editorial Board members.
+* Adoption or amendment of this RFC or other documents establishing the
+  authority or structure of the Board.
+
+Any other matter explicitly designated as requiring affirmative OMG action.
 
 ## Managing Editor
 
--- a/rfc/10/index.md
+++ b/rfc/10/index.md
@@ -103,6 +103,9 @@ The OMG shall:
   the Editorial Board.
 * Ratify changes to the membership of the Editorial Board through review of the
   Editorial Board Roster.
+* Remove or replace Editorial Board members who are unable to fulfill the
+  responsibilities of membership, subject to any procedures defined in the
+  Editorial Board Policies.
 * Serve as an escalation body in case of:
   - persistent Board deadlock;
   - procedural disputes; or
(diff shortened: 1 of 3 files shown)


rfc/10/index.md
```diff
@@ -0,0 +1,308 @@
...
+shall apply to all stakeholders in the NGFF governance bodies.
+
+## OME Management Group (OMG)
+
+The [OME Management Group (OMG)](https://www.openmicroscopy.org/governance/management-group/)
+provides strategic and procedural supervision of the NGFF Editorial Board. It
+does not participate in routine RFC deliberations unless escalation is
+required. 
+
+The OMG shall:
```
normanrz, line 100: Not sure where you wanted to specify this, but I still think the formal acts of the OMG should be a default-accept of recommendation of the EB within a specified time frame. See https://docs.google.com/document/d/1picvG4yq0K_gIiF83C5rAy3AguqJeWkdAxL-vds5BgI/edit?disco=AAAB3IaXQ3Q

joshmoore: Ah, thank you. I see what happened. A copy (fork 🙈) went live for the OME2026 meeting and the two drifted. I'll re-sync post vacation.

lubianat:
also for after :palm_tree: ordering is alphabetical, probably not what we want: 

grafik

related: 
- #426

rfc/10/index.md
```diff
@@ -0,0 +1,308 @@
...
+The EB continues to be tasked with supporting timely decision-making, ensuring
+continuity and accountability, and providing a clear governance structure
+within the NGFF community which will become increasingly important during the
+finalization phase of the 1.0 release.
+
+# Background
+
+The path to NGFF 1.0 involves a series of interdependent technical choices—some
+of which require decisive, binary resolutions (e.g., those raised in [RFC-3][] and
+[RFC-6][]. To avoid prolonged uncertainty and to maintain momentum toward a stable
```
lubianat:
```suggestion
[RFC-6][]). To avoid prolonged uncertainty and to maintain momentum toward a stable
```

rfc/10/index.md
joshmoore: I think the main two missing sections were:
 * ZSC: re-added
 * OMG relationship: rewritten
 
 Let me know what you think.

rfc/10/index.md
```diff
@@ -0,0 +1,328 @@
...
+timeline or stability concerns may be closed. The Board acts only after
+reasonable opportunity for public review and community input.
+
+A formal vote may be called when:
+
+* Consensus has not emerged after reasonable discussion; or  
+* Timeline constraints require resolution.
+
+The Managing Editor determines when voting is appropriate. Quorum is defined as
+two-thirds of active Editorial Board members, rounded up to the nearest whole
```
normanrz, line 237: How large do you envision the EB to become in the future (let's say 5 years)? With just a handful of people, two-thirds essentially is "everybody - 1". Under failure modes this becomes blocking. I think there should be a mechansim to lower the quorum requirements in case of prolonged blocking.

normanrz (commented):
Thanks for the edits! 
I think a few failure modes need more thoughts:
- If EB members fail on their duties (e.g. prolonged inactivity), how can they be involuntarily removed? Is that a responsibility of the OMG?
- What options does the EB have in case of failure of the OMG to fulfill their duties (e.g. approvals, appointments)?

rfc/10/index.md
joshmoore: I'm not sure we know how large is going to be ideal, and I agree `everyone - 1` is also not ideal, however, I think the mechanism is precisely escalation to the OMG. I'm hesitant to try to build anything more complicated at this stage, but definitely support reviewing this, especially if we start to have these issues.

To your other points (in order to keep the single thread:

1. If EB members fail on their duties (e.g. prolonged inactivity), how can they be involuntarily removed? Is that a responsibility of the OMG?
  - Pushed two minor updates. Hopefully much of this can be specified in the policies as the EB & OMG learn how best to track & handle such situations. 
  
2. What options does the EB have in case of failure of the OMG to fulfill their duties (e.g. approvals, appointments)?
  - Pushed a section on using lazy consensus. I would not try to deal with the failure of the OMG in this document but that should be handled at the top-level itself.

normanrz: Thanks! I think this sounds all very reasonable.

rfc/10/index.md
```diff
@@ -125,13 +125,36 @@ requirements have been satisfied. Such intervention should be exceptional and
...
+### Lazy consensus and OMG oversight
+
+Matters referred to the OMG for review shall normally follow a lazy-consensus
+process: the Board may proceed unless a substantive objection is raised. The
+OMG should make reasonable efforts to review such matters in a timely manner.
+
+The target review period is one full working week, with an outer limit of two
+full working weeks. If no substantive objection has been raised within the
+outer limit, the Board may proceed. Periods of individual OMG member
+unavailability do not automatically extend these limits.
```
joshmoore: @vuhlmann addition:

> The review period may be extended by mutual agreement between the Board and the OMG where the Board considers additional OMG feedback valuable.

cc: @normanrz for a last :+1:

lubianat:
the rfc/index.md file has not been changed, so I cannot make an inline suggestion. 

changing the toctree content in `rfc/index.md` to 
```
:maxdepth: 1
:hidden:
0/index
1/index
2/index
3/index
4/index
5/index
6/index
7/index
8/index
9/index
10/index
```

Fixes the sidebar issue 

image

joshmoore:
Pushed, @lubianat. Anything else?

lubianat:
not on my end, LGTM

lubianat (approved):

PR #560, opened 2026-07-04
Improvements to the text of RFC-3

jni:
This PR updates RFC-3 to account for comments, questions, and data types raised
since it was created:

- There was [discussion][exotic] that RFC-3 is primarily for exotic use cases.
  I'll admit that I have some of those in mind :joy: but it is actually mostly
  going to serve standard microscopy images, as demonstrated by [Zeiss's
  interest][sebi], as well as [comment 3] from Cornelia Wetzker detailing their
  use of FLIM. In private communications, Cornelia mentioned that most of their
  data is in LIF (Leica) format, which supports 6D. I've tried to amend the
  text to front load these standard microscopy use cases.
- I've added FLIM to the data types that RFC-3 targets, to address Comment-3.
- I've added some user stories around some of the data types.
- I've added specifics to the proposal about how the spec's language would
  change.
- Finally, I've made two sample RFC-3 datasets using real-world data in
  clbarnes/ome-zarr-rfc3-data#1 that can be used to test compliance with RFC-3.

Hopefully, this is enough to get this over the finish line soon...!

[exotic]: https://imagesc.zulipchat.com/#narrow/channel/328251-NGFF/topic/RFC-3.3A.20remove.20dimension.20restrictions/near/568823806
[sebi]: https://imagesc.zulipchat.com/#narrow/channel/328251-NGFF/topic/RFC-3.3A.20remove.20dimension.20restrictions/near/568244152
[comment 3]: https://ngff.openmicroscopy.org/rfc/3/comments/3/index.html

### DIFF
--- a/rfc/3/index.md
+++ b/rfc/3/index.md
@@ -149,30 +149,37 @@ RFC-5, maintains these limitations:
 > - `axes` MAY contain one additional entry of `type:channel` or a null /
 >   custom type.
 >
-> - axes entries MUST be ordered by type where the time axis must come first
->   (if present), followed by the channel or custom axis (if present) and the
->   axes of type space.
+> - `axes` entries MUST be ordered by `type` where the `time` axis must come
+>   first (if present), followed by the `channel` or custom axis (if present)
+>   and the axes of type `space`.
 >
 > - If there are three spatial axes where two correspond to the image plane
->   (yx) and images are stacked along the other (anisotropic) axis (z), the
->   spatial axes SHOULD be ordered as zyx.
+>   (`yx`) and images are stacked along the other (anisotropic) axis (`z`), the
+>   spatial axes SHOULD be ordered as `zyx`.
 
 
 These restrictions prevent users from converting existing
 datasets to OME-Zarr. For example, Zeiss .czi datasets [may contain][czi format
 dimensions] dimensions such as H, I, and V to store different phases,
-illumination directions, or views respectively. To say nothing of synthetic data
-that may contain "artificial" dimensions such as principal components or axes of
-other dimensionality reduction-techniques from many images. They may also hamper
-the adoption of OME-Zarr as an acquisition-time format due to performance concerns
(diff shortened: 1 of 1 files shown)


joshmoore:
Thanks @jni 🙏 Would you mind also updating the version history at the bottom before merge?

jni:
@joshmoore yup, did the thing!

jo-mueller:
Looking cool! Just one more thing that came up during the discussion around https://github.com/ome/ngff-spec/pull/130 (and this is only relevant because RFC3 happened to go in after RFC5):

I think there could (should?) be a statement about the `discrete` flag in the axes metadata, by which a dimension can declare whether interpolation is well-defined along itself. By extension, this can imform transformations how to treat all non-spatial axes.

jni:
@jo-mueller what kind of statement? RFC-5 states axes/dimensions can be discrete or not, and RFC-3 has nothing more to say on the matter — we agree! :joy:

jni:
Having looked at it in more detail, the rc0 version of the spec states:

> [axes metadata] MAY contain the field discrete. The value MUST be a boolean, and is true if the axis represents a discrete dimension (see below for details).

and

> Arrays are inherently discrete (see Array coordinate systems, below) but are often used to store discrete samples of a continuous variable. The continuous values “in between” discrete samples can be retrieved using an interpolation method. If an axis is continuous ("discrete" : false), it indicates that interpolation is well-defined. Axes representing space and time are usually continuous. Similarly, joint interpolation across axes is well-defined only for axes of the same type. In contrast, discrete axes ("discrete" : true) may be indexed only by integers. Axes representing a channel, coordinate, or displacement are usually discrete.

I can reiterate that in the multiscale metadata section, but it would be redundant?

jni:
@joshmoore as far as I'm concerned this is ready to be merged and sent out for (hopefully final) review! :pray:

lubianat (approved):
looks good!

Issue #535, opened 2026-06-11
Use html-proofer or similar in the CI to check for broken links and alt text


lubianat:
via #533 and @elichad, who shared: 

Thanks! FWIW for the RO-Crate website we use [html-proofer](https://github.com/gjtorikian/html-proofer) in our CI to check for broken links (and also alt text on images) - it's saved us a lot more times than I expected. [example CI job](https://github.com/ResearchObject/ro-crate/blob/5683d43e5a7085ad8dfb4b7141667421effca93c/.github/workflows/test.yml#L36-L44)

PR #561, opened 2026-07-17
Improve "Publications" page and add minor clean-ups in the front page

lubianat:
Adds a "last articles about OME-Zarr + NGFF" feed to the /publications page 

Also changes the first page to add clearer links + improving wording on when to use/not as the ecosystem evolves

### DIFF
--- a/.github/workflows/html-proofer.yml
+++ b/.github/workflows/html-proofer.yml
@@ -1,10 +1,29 @@
 name: "Build and Lint Site"
 
 on:
-  push:
-    branches: ["main"]
+  workflow_dispatch:
   pull_request:
     branches: ["main"]
+    paths:
+      - "specifications/**"
+      - "rfc/**"
+      - "resources/**"
+      - "community/**"
+      - "contributing/**"
+      - "help-desk/**"
+      - "_html_extra/**"
+      - "_static/**"
+      - "_ext/**"
+      - "images/**"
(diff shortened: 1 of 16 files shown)


jo-mueller:
Looks cool! I think I'd personally prefer a scroll box for navigation because the page layout makes the nav section at the bottom a bit confusing:

image

Also, the link to the publications page on the frontpage does not work.

- Link address: https://ngff--561.org.readthedocs.build/index.html#./publications/index
- Publications page: https://ngff--561.org.readthedocs.build/resources/publications/index.html

An alternative approach could be to have a small python function that queries `"https://www.ebi.ac.uk/europepmc/webservices/rest/search` as done in this PR and pulls a bibtex entry for every discovered publication. One could pretty easily just insert something like 

````{markdown}
```{bibliography}
citations.bib
```
````

at `publications/index.md`

lubianat:
I added a wrong internal link in the last commit, similar to #562 

grafik


it indeed triggered a blocked check

lubianat:
and thanks @jo-mueller ! the python function would work as well, but then we would need to set up GH actions to run it regularly, and keep sure all pieces were well connected 

the html block is somewhat ugly, but it may be more contained/maintainable

lubianat:
This now closes #562 and closes #535

lubianat:
Split in #567 and #566

PR #564, opened 2026-07-30
feat: serve dev schemas under their actual name

### LINKED ISSUE #69: Links to version schema broken
e.g., at https://ngff-spec.readthedocs.io/en/latest/schemas/well/:

Image

### LINKED ISSUE #563: 0.6rc0 schemas not served from homepage
I just noticed over at the validator of 0.6 data, that the schemas apparently aren't being served correctly from the ngff domain, which makes the validator fail.

This works: [https://ngff.openmicroscopy.org/0.5/schemas/image.schema](https://ngff.openmicroscopy.org/0.5/schemas/image.schema)

this doesn't work: [https://ngff.openmicroscopy.org/0.6rc0/schemas/image.schema](https://ngff.openmicroscopy.org/0.6rc0/schemas/image.schema)

... and this does: [https://ngff.openmicroscopy.org/dev/schemas/image.schema](https://ngff.openmicroscopy.org/dev/schemas/image.schema)

As it currently stands, there are generally two ways to get the schemas from the web:
- Either by pulling them from content that is served from this page, i.e., by following the above links. When 0.6 is released, there should be another submodule on this page which then provides the schemas under `https://ngff.openmicroscopy.org/0.6/schemas/something.schema` besides the `dev` branch for the latest schemas.
- Directly from the release of the `ngff-spec` repo, i.e., like this: [https://raw.githubusercontent.com/ome/ngff-spec/0.6rc0/schemas/image.schema](https://raw.githubusercontent.com/ome/ngff-spec/0.6rc0/schemas/image.schema). This doesn't currently work for the older versions, because they are not tagged and released formally from the ngff-spec page.

We could simply create the tags for the past versions, then the second option would also work for all versions.

cc @jni since we talked about this recently.

jo-mueller:
Fixes #563 
Fixes https://github.com/ome/ngff-spec/issues/69

Small change to how the schemas are provided from the ngff page. In the current build, schemas are always served under a folder name that is derived from the name of the git submodule. For the dev versions, I see value in serving the (latest) schemas under both

- `https://ngff.openmicroscopy.org/dev/schemas/...`
- `https://ngff.openmicroscopy.org/0.9.dev17/schemas/...`

which would be possible through this PR.

*Edit*: I think this will also greatly reduce the number of warnings for the RTD build, since all the schema URLs will now correctly resolve to the ngff page, which is something @dstansby reported already a while ago (see https://github.com/ome/ngff-spec/issues/69). I thought it was a json-schema-forhumans-issue, but the problem is maybe that schema URLs could not be resolved correctly since they weren't served under their actual name.

cc @lubianat

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -108,6 +108,15 @@ def build_served_html():
         if os.path.isdir(os.path.join("specifications", d))
     ]
 
+    # create a special folder for extra html for the latest schemas
+    # that refers to the development version of the specification
+    sys.path.insert(0, str(Path(__file__).parent))
+    from specifications.dev._version import __version__ as dev_version
+
+    os.makedirs(f"_html_extra/{dev_version}/schemas", exist_ok=True)
+    for schema in glob.glob(f"specifications/dev/**/*.schema", recursive=True):
+        shutil.copy2(schema, f"_html_extra/{dev_version}/schemas/")
+
     for version in versions:
 
         # copy schemas to _html_extra


jo-mueller:
Downloading from [https://ngff--564.org.readthedocs.build/0.6rc0/schemas/image.schema](https://ngff--564.org.readthedocs.build/0.6rc0/schemas/image.schema) works 👍

lubianat:
Nice! what about the previous 0.6 dev versions? e.g. https://ngff--564.org.readthedocs.build/0.6dev4/schemas/image.schema?  should we figure out a way to serve them from here, or not really? 

My worry is that https://ngff--564.org.readthedocs.build/0.6rc0/schemas/image.schema  will work now, but as soon as we update the "dev" branch, it will not anymore, right?

jo-mueller:
Decision: This PR would make pre-release URLs (i.e. `https://ngff.openmicroscopy.org/0.9.dev17/schemas/...`) essentially transcient, which is problematic. The better solution is to

- create the 0.1...0.5 tags over at ngff-spec
- access all schemas through `https://raw.githubusercontent.com/ome/ngff-spec/<version>/schemas/image.schema`
- add redirects here to `conf.py` so we have better control about which tags/schemas exactly are served from here.

PR #565, opened 2026-07-31
docs: document OME-Zarr for particular scientific modalities

lubianat:
Open to new ideas. 

This goes in line with the idea of displaying "where the science is" in connection with NGFF. 

Pages to curate resources releated to OME-NGFF plus a particular scientific community (WSI, Volume-EM, dynamics, omics, etc). 

The goal is to be able to provide a quick link to share that showcases the value of OME-NGFF for people heavily invested in topic X.


lubianat:
@uermel added a Cryo-ET page (as mentioned on https://github.com/uermel/chimerax-ome-zarr/issues/18) 

If you have anything else to add, feel free to do so. 

It is all mostly a draft and pages are supposed to be somewhat simple

uermel:
Looks good to me, I don't think I have anything to add. CryoET portal seems duplicated in cryoET section and in the volume EM section, not sure if that is intentional (it's not wrong).

lubianat:
I may add a section on 'expansion microscopy' to the hub. I saw an interesting discussion at 
https://forum.image.sc/t/does-anyone-use-multiscale-coordinatetransformations/122243 and 
https://github.com/AI-HHMI/miao/issues/25

(ping @jo-mueller , interesting RFC-5 use case!)

PR #567, opened 2026-08-03
Adds HTML-Proofer and fixes broken links

lubianat:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/.github/workflows/html-proofer.yml
+++ b/.github/workflows/html-proofer.yml
@@ -1,10 +1,29 @@
 name: "Build and Lint Site"
 
 on:
-  push:
-    branches: ["main"]
+  workflow_dispatch:
   pull_request:
     branches: ["main"]
+    paths:
+      - "specifications/**"
+      - "rfc/**"
+      - "resources/**"
+      - "community/**"
+      - "contributing/**"
+      - "help-desk/**"
+      - "_html_extra/**"
+      - "_static/**"
+      - "_ext/**"
+      - "images/**"
+      - "**/*.md"
+      - "conf.py"
+      - "requirements.txt"
+      - "readthedocs.yml"
+      - ".github/workflows/build-lint.yml"
+
+concurrency:
+  group: build-lint-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build-site:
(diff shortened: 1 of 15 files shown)


contributing/specification/index.md
```diff
@@ -6,14 +6,14 @@ request for comment (RFC) documents.
 
 In the future we will flesh out this page with a guide to RFCs, but in the
 meantime the RFC process is outlined in the
-[Implementation section of RFC 1](../rfc/1/index.md#implementation).
+[Implementation section of RFC 1](rfc1-implementation).
```
joshmoore, line 9: :100: for improving anchor usage. We likely want to document suggestions on the anchor naming strategy.

rfc/3/versions/2026-07-17/index.md
```diff
@@ -304,36 +304,31 @@ document, taking as base the current development version:
...
 >    SHOULD be named 'z', 'y', and 'x', except where rule 4 applies.
-> 3. *If* a dataset contains exactly 1 time dimension, it should be named `t`.
+> 3. _If_ a dataset contains exactly 1 time dimension, it should be named `t`.
 > 4. When image data axes map straightforwardly to axes with common names in
 >    the relevant field of practice, those axes SHOULD be named according to
 >    such conventions. For example, spatial frequency axes resulting from a
 >    Fourier transformation of `z', 'y', and 'x' SHOULD be named 'w', 'v', and
->    `u`, respectively. Similarly, a temporal frequency axis resulting from
->    a Fourier transformation of the `t` axis SHOULD be named `w` or `ω`.
+`u`, respectively. Similarly, a temporal frequency axis resulting from
```
clbarnes: I don't agree with the space removals here

.github/workflows/html-proofer.yml
```diff
@@ -0,0 +1,62 @@
...
+      - "requirements.txt"
+      - "readthedocs.yml"
+      - ".github/workflows/build-lint.yml"
+
+concurrency:
+  group: build-lint-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-site:
```
clbarnes, line 29: Marginal preference for one job to build the site and save it as an artifact, and then a dependent job which reads the artifact and runs the proofer. Then it's easier to tell at a glance exactly what's failing.

rfc/3/versions/2026-07-17/index.md
lubianat: thanks, yeah, this was automatic; I'll rollback this

.github/workflows/html-proofer.yml
lubianat: @clbarnes that is doable, but the trade-off is more time on the CI run, perhaps adding 1-2 min for building the artifact and getting it again.

 It is currently faster than the RTD build. I wouldn't like for this to be the bottleneck, as it runs every single time.

lubianat: @clbarnes ok for you?

lubianat: (oh sorry for the nudged, I had already tagged in the previous comment)

clbarnes: That makes sense, but I'd be wary about adding any further validation which need to run against the built pages and stacking them all into one job.

contributing/specification/index.md
jo-mueller: In RFC5 I used `(rfcs:rfc5:some_anchor)=` which can be referenced from everywhere as [text](rfcs:rfc5:some_anchor), even if the file with the anchor/the link move. I think this can easily be extended for the rest of the cross-referencing?

jo-mueller: See https://github.com/ome/ngff/issues/576

jo-mueller (approved):

jo-mueller:
Looking good! I opened https://github.com/ome/ngff/issues/576 as a follow up so we don't forget to add visible documentation about anchor usage

.github/workflows/html-proofer.yml
lubianat: agreed! if we want more checks later, we refactor.  @clbarnes, thanks! would you be confident to give an explicit approval here? or better find someone else? (totally ok, of course)

clbarnes (approved):

lubianat:
Thanks @clbarnes and @jo-mueller! I'll merge it now.

Issue #462, opened 2026-03-03
Rename latest schema documents (#237) — moved from PR


lubianat:
The PR #237 was opened by @d-v-b and stated: 

> changes the suffix of the schema documents in the latest version of the spec from .schema to .schema.json, to enable JSON syntax highlighting in text editors.
> Because this will be a breaking change for anyone linking to the published schemas under the old names, this PR should not be merged until we reach a solution to this problem (e.g., a redirect).
> addresses #236

The Pull Request currently needs to be ported to #ome/ngff-spec

Issue #236 stated: 

> The [schema documents](https://github.com/ome/ngff/tree/main/latest/schemas) are named with a .schema file extension. Because the schema documents are JSON schema, (which is JSON) .json is probably the right file extension to use instead. This change would make these documents much easier to read / edit, e.g. on github or in IDEs that provide json syntax highlighting. If it's important to designate these documents as JSON schema, we could do <name>.schema.json.

>Is there a compelling reason for .schema as an extension, or can we move to schema.json?

This issue supersedes #236, with an update. 

On today's PR review (https://github.com/German-BioImaging/incubator/issues/54) we discussed this as one of the first things _after_ the 0.6 release (i.e. for 1.0.dev), if that is okay. 

@jo-mueller may make the PR when the time comes, though others are welcome to do so too

PR #581, opened 2026-08-14
feat: serve schemas based on which tags exist on gh

jo-mueller:
PR following #564 that solves the problem a bit more elegantly (I hope):

This PR changes the build process of the page so that the following happens on build:

- All tags on ome/ngff-spec are automatically retrieved with `git ls-remote --tags https://github.com/ome/ngff-spec`
- This gives a list of tagged versions on that repo (currently the ones [listed below](https://github.com/ome/ngff/pull/581#issuecomment-5450270176))
- Copy all the schemas over here and serve them under the respective tag as _.schema_ as well as _.schema.json_
- The schema publishing for the older versions (0.1...0.5 + dev) remains active as we still have the submodules in place here, from which we copy the schemas and serve them as well. Currently, schemas from the submodules **supersede the schemas from the tags**, but that's not a problem since none of the submodule'd schemas are from stagged version (i.e., 0.1...0.5 have not been tagged at ngff-spec)

This way, we are able to access the schemas of all past (we'd need to tag them) and present through the ngff page, _including the dev versions_. This also centralizes the tagging over at ngff-spec as a core mechanic to make schemas available through the ngff page.

To figure out next: If the 0.5 schema has changed, should it be published under 0.5 or under 0.5.X? which is a question we don't yet have answered.

cc @jni

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -105,43 +105,38 @@ def build_served_html():
     
     # Create .htaccess to serve schemas inline (not download)
     os.makedirs("_html_extra", exist_ok=True)
-    htaccess_content = """<FilesMatch "\\.schema$">
-    Header set Content-Disposition "inline"
-</FilesMatch>
-"""
-    with open("_html_extra/.htaccess", "w") as f:
-        f.write(htaccess_content)
-    print(f"✅ Created .htaccess to serve schemas inline")
-    
+
     # Fetch GitHub tags and download schemas
-    try:
-        result = subprocess.check_output([
-            "git", "ls-remote", "--tags", "https://github.com/ome/ngff-spec"
-        ], text=True, timeout=10)
-        tags = [line.split()[1].replace("refs/tags/", "").rstrip("^{}") for line in result.strip().split("\n") if line]
-        for tag in sorted(set(tags)):
-            schema_dir = f"_html_extra/{tag}/schemas"
-            os.makedirs(schema_dir, exist_ok=True)
-            # Download schemas from GitHub raw for this tag
-            gh_url = f"https://github.com/ome/ngff-spec/archive/refs/tags/{tag}.tar.gz"
-            try:
-                import tempfile, tarfile
-                with tempfile.NamedTemporaryFile(delete=False) as tmp:
-                    subprocess.check_call(["curl", "-sL", gh_url, "-o", tmp.name])
-                    with tarfile.open(tmp.name) as tar:
-                        for member in tar.getmembers():
-                            if "/schemas/" in member.name and member.name.endswith(".schema"):
-                                # Extract just the filename, flatten into schema_dir
-                                target = os.path.join(schema_dir, os.path.basename(member.name))
-                                tar.extract(member, path=tempfile.gettempdir())
-                                src = os.path.join(tempfile.gettempdir(), member.name)
-                                shutil.copy2(src, target)
-                    os.unlink(tmp.name)
-                print(f"✅ Downloaded schemas for {tag}")
-            except Exception as e:
-                print(f"⚠️  Could not download schemas for {tag}: {e}")
-    except Exception:
-        pass
+    result = subprocess.check_output([
+        "git", "ls-remote", "--tags", "https://github.com/ome/ngff-spec"
+    ], text=True, timeout=10)
+    tags = [
+        line.split()[1].replace("refs/tags/", "").rstrip("^{}")
+        for line in result.strip().split("\n") if line
+        ]
+
+    # Clone repo once, checkout each tag (faster than per-tag tarball download)
+    repo_path = "_temp_ngff_spec"
+    if not os.path.exists(repo_path):
+        subprocess.check_call(["git", "clone", "https://github.com/ome/ngff-spec", repo_path])
+    
+    for tag in sorted(set(tags)):
+        schema_dir = f"_html_extra/{tag}/schemas"
+        os.makedirs(schema_dir, exist_ok=True)
+        try:
+            subprocess.check_call(["git", "-C", repo_path, "checkout", tag], 
+                                stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+            src_schemas = glob.glob(os.path.join(repo_path, "**", "*.schema"), recursive=True)
+
+            if len(src_schemas) == 0:
+                print(f"⚠️  No schemas found for {tag}")
+                continue
+
+            for schema_file in src_schemas:
+                shutil.copy2(schema_file, schema_dir)
+            print(f"✅ Checked out schemas for {tag}")
+        except Exception as e:
+            print(f"⚠️  Could not checkout {tag}: {e}")
     
     # Build specifications from local submodules
     displayed_spec_versions = [
@@ -175,6 +170,13 @@ def build_served_html():
 
         subprocess.check_call([sys.executable, script])
         print("✅ Built rendered examples/schemas for version", version)
+
+        # copy schemas to _html_extra for served html
+        schema_files = glob.glob(f"specifications/{version}/**/*.schema", recursive=True)
+        for schema_file in schema_files:
+            dest_dir = os.path.join("_html_extra", version, "schemas")
+            os.makedirs(dest_dir, exist_ok=True)
+            shutil.copy2(schema_file, dest_dir)
 
         # build jupyter-book docs in specification submodules
         myst_file = glob.glob(f"specifications/{version}/**/myst.yml", recursive=True)[
--- a/conf.py
+++ b/conf.py
@@ -133,7 +133,9 @@ def build_served_html(clean_up=True):
                 continue
 
             for schema_file in src_schemas:
-                shutil.copy2(schema_file, schema_dir)
+                dest_file = os.path.join(schema_dir, os.path.basename(schema_file))
+                shutil.copy2(schema_file, dest_file)
+                shutil.copy2(schema_file, dest_file + '.json')  # dual format
             print(f"✅ Checked out schemas for {tag}")
         except Exception as e:
             print(f"⚠️  Could not checkout {tag}: {e}")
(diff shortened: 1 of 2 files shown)


jo-mueller:
@lubianat do you have ideas how to make this nicer? I tried a few things with redirects, but maybe the [RTD redirects](https://docs.readthedocs.com/platform/stable/guides/redirects.html#saving-the-redirect) supersede any redirects to outside pages (i.e. via an htaccess file) I am defining here, which is annoying.

To have the browser render the schemas as text instead of downloading the files instantly, we would need to rename the files into .json, I think.

lubianat:
@jo-mueller should we go ahead and add the .json for the next releases (per https://github.com/ome/ngff/issues/462)? 

I don't have a suggestion on how to make it nicer, though

jo-mueller:
We could serve them as both *.schema and *.json? For the time being, this could also happen here until we make the change at ngff-spec?

lubianat:
> We could serve them as both *.schema and *.json? For the time being, this could also happen here until we make the change at ngff-spec?

I don't have a good perspective on the implications. I guess we can do it at least transitionally, while tools move towards `.json` or `.schema.json`

jni:
> including the dev versions

I expect that we would want those to be hidden under a toggle/details/whatever.

jni:
How do I review this? I don't know what you mean by "nicer", and looking at [the preview build](https://ngff--581.org.readthedocs.build/specifications/index.html) I only see 0.6rc0, no dev versions...

jo-mueller:
@jni I think there's two aspects to this:

1. displaying the spec texts of all tagged versions
2. delivering the schemas for all tagged versions

This PR only addresses the second one. I hadn't thought about doing the same with all the tagged versions, but I think that'll be a bit more complex (but not much). I guess the approach to automatically getting all the tagged spec versions here, would be something like:

- remove the submodules from this repo entirely
- on build, get all the tagged versions (as this PR does it)...
- ....and clone all of them (which this PR doesn't yet do).
- The toctree that organizes the content of the left sidebar on [https://ngff.openmicroscopy.org/specifications/index.html](https://ngff.openmicroscopy.org/specifications/index.html) would then need to be built automatically as part of that CI.

On a plus side, that would get rid of all submodule shenanigans here, which I think would be a big plus ^^" On the downside, that would potentially be quite a bit of cloning/updating stuff as part of a normal build command?

jo-mueller:
decision point: We serve the schemas as both .schema and .schema.json, in the future (on the 0.9dev branch) we rename the schemas to .json or .schema.json

jni:
@jo-mueller So, on the schemas side: where are the schemas actually served from this PR? Can I get a RTD build URL? Also, can you re-trigger build maybe with an empty commit so we can see the 0.9.dev1 schemas get served?

I finally grokked the .schema vs .schema.json issue. I suggest serving both for now. 0.9.dev1 links to the historical .schema files, so let's punt on fully renaming to at least .dev2.

jni:
Ok, found it:

https://ngff--581.org.readthedocs.build/0.6.dev1/schemas/label.schema

jni (approved):
Since currently the 0.9.dev1 schemas are not being published, I consider this a strict improvement and would advocate for it getting merged ASAP.

jo-mueller:
@jni small thing that was missing here: I modified the code a bit so that schemas are copied to the `_html_extra` directory for
- all tagged versions of the ngff-spec repo (0.6.dev1...0.9.dev1)
- all submoduled versions (0.1...0.5)

Without this change, we would stop serving schemas for 0.1...0.5 :)

Also to keep in mind that once https://github.com/ome/ngff-spec/issues/187 is resolved, we can do a bit less special-casing here and split the build process in two functions that

- function 1: clone/fetch all the schemas according to the tag
- function 2: Build and render the desired example documents for a subset of versions we want to display (0.1...0.6 + dev)

Edit: Also, the failing lint check points to some broken cross-links in the 0.9.dev1

> For the Links > Internal check, the following failures were found:
> 
> * At ./_build/html/_temp_ngff_spec/index.html:626:
> 
>   internally linking to #spec:example:coordinate_transformation; the file exists, but the hash 'spec:example:coordinate_transformation' does not
> 
> * At ./_build/html/_temp_ngff_spec/index.html:655:
> 
>   internally linking to #orientation-values; the file exists, but the hash 'orientation-values' does not

jo-mueller:
@jni ok, all the schemas are available on the RTD build:

- https://ngff--581.org.readthedocs.build/0.1/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.2/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.3/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.4/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.5/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.6.dev1/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.6.dev2/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.6.dev3/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.6.dev4/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.6rc0/schemas/image.schema
- https://ngff--581.org.readthedocs.build/0.9.dev1/schemas/image.schema

*Edit* And as a little extra: All the schemas are now also available with a .schema.json extension (e.g., https://ngff--581.org.readthedocs.build/0.9.dev1/schemas/image.schema.json) so the browser shows you the schema instead of downloading it.

conf.py
```diff
@@ -141,6 +175,15 @@ def build_served_html():
...
         print("✅ Built rendered examples/schemas for version", version)
 
+        # copy schemas to _html_extra for served html
+        schema_files = glob.glob(f"specifications/{version}/**/*.schema", recursive=True)
+        for schema_file in schema_files:
+            dest_dir = os.path.join("_html_extra", version, "schemas")
+            os.makedirs(dest_dir, exist_ok=True)
+            dest_file = os.path.join(dest_dir, os.path.basename(schema_file))
+            shutil.copy2(schema_file, dest_file)
+            shutil.copy2(schema_file, dest_file + '.json')  # ponytail: dual format
```
lubianat: ponytail comment left here

conf.py
```diff
@@ -102,20 +102,54 @@ def build_served_html():
...
 
-    for version in versions:
-
-        # copy schemas to _html_extra
-        os.makedirs(f"_html_extra/{version}/schemas", exist_ok=True)
-        schemas = glob.glob(f"specifications/{version}/**/*.schema", recursive=True)
-        for schema in schemas:
-            shutil.copy2(schema, f"_html_extra/{version}/schemas/")
-        print(f"✅ Copied schemas for version {version}")
+    for version in displayed_spec_versions:
```
lubianat, line 152: if a tag and a displayed_spec_version conflict, the displayed_spec_version overrides that schema? 

Just making sure that is the intended behaviour.

lubianat (dismissed):
LGTM; I think it closes #462 too (maybe that in the commit message)

conf.py
jo-mueller: yikes. fixed in f2a5b2c3ba7a6cc4324a3eb00662790255886314

conf.py
jo-mueller: @lubianat Yes, right now that's what happens and that will be the case until the old versions are [also tagged on ngff-spec](https://github.com/ome/ngff-spec/issues/187).

Even when the older versions are tagged over on ngff-spec, the tagged versions of these branches (i.e., the 0.5 branch) contains the same content as the *last version* of the same branch. Only when we make changes to some of these branches (0.1...0.5) that include changes to the schemas that are newer than the tag on these branches are we going to run into problems.

*Then* the question will be: *If the 0.5 schema has changed, should it be published under 0.5 or under 0.5.X?* which is a question we don't yet have answered. I'll make a note about that above in the PR description 👍

lubianat: I think 0.5 --> latest in the 0.5 series

And 0.5.0, 0.5.1, etc etc, whatever minor changes are added

jo-mueller: So if a 0.5.1 tag came into existence over at ngff-spec, it would also be served under that URL and 0.5 would continue to be served as it is the currently checked out submodule.

will-moore:
@jo-mueller I'm trying to update the validator to dynamically parse each schema to determine what `$refs` it contains and then to load those schemas, instead of the previous hard-coded behaviour, e.g. for v0.6:

```
    const names = ["coordinate_transformations", "coordinate_systems", "axes", "_version"];
    for(const name of names) {
      const schema = await getSchema(getSchemaUrl(name, version));
      schema["$id"] = `https://ngff.openmicroscopy.org/0.6rc0/schemas/${name}.schema`;
      refSchemas.push(schema);
    }
```

The new code is a work in progress (PR soon) but it's working for most `$refs` e.g. https://ngff--581.org.readthedocs.build/0.6rc0/schemas/image.schema.json contains `"$ref": "https://ngff.openmicroscopy.org/0.6rc0/schemas/coordinate_systems.schema#/$defs/coordinateSystem"` so then I load https://ngff--581.org.readthedocs.build/0.6rc0/schemas/coordinate_systems.schema.json

However, that `coordinate_systems.schema` contains `"$ref": "axes.schema"` which doesn't look like a valid `$ref`.
This needs some tricky hard-coding to work around that.
I see the same `"$ref": "axes.schema"` in older schemas too, e.g.
https://raw.githubusercontent.com/ome/ngff-spec/refs/heads/main/schemas/coordinate_systems.schema

Any chance that could be fixed to be consistent with the other refs?
Thanks

will-moore:
validator PR to test this PR is at https://github.com/ome/ome-ngff-validator/pull/86

jo-mueller:
@will-moore 

> Any chance that could be fixed to be consistent with the other refs?

Sounds like a real schema bug so we definitely should fix this!

PR #584, opened 2026-08-20
RFC-4: re-sync the axes schema description with the RFC text

vboussot:
The `axes` slot description in `rfc/4/orientation.yml` states two rules that the
RFC text does not:

> Furthermore, if the attribute orientation is defined for one axis of type
> space, it MUST be defined for all the axes of type space. In this case, the
> type of each orientation MUST be the same and the value MUST be unique.

Against the current `rfc/4/index.md`:

* `orientation` is OPTIONAL and no completeness requirement is stated, so an
  image may orient only some of its spatial axes. A limb scan, for example, can
  give a proximal/distal direction for one axis and leave the medio-lateral
  axis undefined.
* A single `type` is defined, `"anatomical"`. Requiring the declared types to
  agree with each other accepts metadata that uses a type the RFC does not
  define.
* The constraint on values is the mutual exclusion rule: "A set of NGFF `axes`
  MUST only have one of the set { left-to-right, right-to-left } or
  { anterior-to-posterior, posterior-to-anterior } or the remaining values."

The model structure already matches the RFC, since `orientation` is optional and
nullable. Only the description was out of date, and it propagates into
`orientation.schema.json`, `orientation.py`, `orientation.ts` and `markdown/`,
where implementers read it as normative.

The `orientation` slot description now also carries the equivalence added in
254c1b2: an absent orientation and a `null` orientation mean the same thing, and
writers should omit the field rather than serialize `null`.

Generated with `pixi run gen-all` in `rfc/4`, then the `orphan: true` front
matter was restored on the markdown pages, which `gen-markdown` does not emit.
Running the generators on an unmodified tree produces no other diff. The JSON
Schema changes only in `description` strings, so no document changes validity.

Found while aligning RFC 4 validation in ngff-zarr
(fideus-labs/ngff-zarr#613), where the removed sentences had been implemented
as constraints.

### DIFF
--- a/rfc/4/markdown/Axes.md
+++ b/rfc/4/markdown/Axes.md
@@ -17,6 +17,6 @@ URI: [ngff:Axes](https://w3id.org/ome/ngff/Axes)
 ### Own
 
  * [➞axes](axes__axes.md)  <sub>0..\*</sub>
-     * Description: A list of axes. Although serialized as list, it MUST be dealt with as being a set as in the name of each axis MUST be unique. Furthermore, if the attribute orientation is defined for one axis of type space, it MUST be defined for all the axes of type space. In this case, the type of each orientation MUST be the same and the value MUST be unique.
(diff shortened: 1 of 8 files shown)


thewtex (approved):

lubianat (approved):
if @thewtex is happy with it, I am happy with it

PR #595, opened 2026-08-27
RFC-4: enforce the anatomical orientation vocabulary again

### LINKED ISSUE #585: RFC-4: the anatomical orientation vocabulary is no longer enforced
`orientation.yml` defines `AnatomicalOrientationValues` with its 24 values and
their definitions, but no slot uses it as a range, so none of the generated
artifacts check the vocabulary. The published `orientation.py` takes this
happily, and the JSON Schema and the TypeScript interface do too:

```python
AnatomicalOrientation(type="planetary", value="north-to-south")
```

RFC 4 defines a single `type`, `"anatomical"`, and takes `value` from the
vocabulary in the same document.

## It looks like this went out with 535e661

At its parent 183ab43, `gen-json-schema` produced
`"type": {"const": "anatomical"}` and
`"value": {"$ref": "#/$defs/AnatomicalOrientationValues"}`. 535e661, "Remove
constraints on orientation name, type slots. Make them generally reusable
throughout the specification", removed the `attributes` block:

```diff
   AnatomicalOrientation:
     is_a: Orientation
-    attributes:
-      type:
-        range: string
-        required: true
-        equals_string: "anatomical"
-        ifabsent: string(anatomical)
-      value:
-        range: AnatomicalOrientationValues
-        required: true
```

The aim was the shared `name` and `type` slots, and the range on `value` looks
like it went along for the ride.

## Restoring it without touching the shared slots

`slot_usage` narrows an inherited slot inside one class, which is what the file
already does for `ChannelAxis`, `SpaceAxis` and `TimeAxis`:

```yaml
  AnatomicalOrientation:
    is_a: Orientation
    slot_usage:
      type:
        equals_string: anatomical
      value:
        range: AnatomicalOrientationValues
```

That brings back the `const` and the `$ref` in `orientation.schema.json`, and
gives `type: Literal["anatomical"]` and `value: AnatomicalOrientationValues` in
`orientation.py`. `orientation.ts` stays as it is, since `gen-typescript` emits
`AnatomicalOrientation` as an empty interface extending `Orientation` and does
not apply the subclass `slot_usage`.

It also helps downstream: when 9472330 added `superficial-to-deep` and the five
other layered and polarized tissue values, they could not be picked up by
regenerating, since the generated types carry the descriptions of the
vocabulary but not the vocabulary itself.

For context, the hand-written `schemas/axis_orientation.schema` added for
0.9.dev1 in ome/ngff-spec#190 does constrain both fields, so the normative side
is covered there. This issue is about the model here and the artifacts generated
from it, which implementations bundle and reuse.

Happy to open a PR if that would be useful.

vboussot:
Closes #585.

`AnatomicalOrientationValues` is defined in the model, but nothing uses it as a range, so the generated artifacts let anything through:

```python
AnatomicalOrientation(type="planetary", value="north-to-south")
```

This adds a `slot_usage` block to `AnatomicalOrientation` that narrows the two inherited slots, the same way `ChannelAxis`, `SpaceAxis` and `TimeAxis` already do. The shared `name` and `type` slots are left alone.

After `pixi run gen-all`, the JSON Schema has `"const": "anatomical"` and a `$ref` to the vocabulary again, and `orientation.py` has `type: Literal["anatomical"]` and `value: AnatomicalOrientationValues`. Both reject the example above and accept `left-to-right`. `orientation.ts` does not change, because `gen-typescript` writes `AnatomicalOrientation` as an empty interface extending `Orientation`.

The markdown keeps the `orphan: true` front matter from 0d68c8b, which `gen-markdown` does not write itself.


thewtex (approved):
@vboussot thank you 🙏 🚀

lubianat (approved):
not sure on the details, but no red flags to me and @thewtex ok'd, so LGTM

lubianat:
Thank you, @vboussot! Merging it here...

PR #597, opened 2026-08-28
RFC-3: Add Review 2 from Fideus Labs

thewtex:
Co-authored-by: Valentin Boussot <valentin@fideus.io>

### DIFF
--- a/rfc/3/reviews/2/index.md
+++ b/rfc/3/reviews/2/index.md
@@ -1,0 +1,100 @@
+# RFC-3: Review 2
+
+(rfcs:rfc3:review2)=
+
+## Review authors
+
+- Matthew McCormick, Fideus Labs
+- Valentin Boussot, Fideus Labs
+
+## Conflicts of interest (optional)
+
+The review authors develop and maintain [ngff-zarr](https://github.com/fideus-labs/ngff-zarr) and related open-source OME-Zarr tooling.
+
+The review is supported by a contract from Image Coop, under Biohub (prime).
+
+## Summary
+
+We thank the author and endorsers for this well-motivated proposal. We agree with the underlying need that RFC-3 addresses, and we find the use cases compelling and important: vendor dimensions such as the H, I, and V axes in Zeiss CZI files; FLIM decay axes; EBSD diffraction patterns and orientation maps; diffusion tensor imaging; Fourier-domain data; multi-coarseness segmentations and overlapping masks; and computational outputs such as parameter sweeps. The current restrictions on the number, names, ordering, and type of dimensions have genuinely limited the conversion of existing datasets, adoption by vendors, and use of OME-Zarr as an acquisition-time format. We support expanding the specification to meet the need of these communities.
+
+In order to preserve shared understanding of meaning, interoperability, and consistent application, we propose encouragement of standardized meaning, names, and identifiers for FAIR-ness. Our remaining concerns are about the consistency of the proposed normative text with itself and with the 0.6rc0 specification it is inserted into, about the semantics of two of the new rules, about how conformance is described, and about the guidance the new freedom calls for. We believe all of them are addressable with minor revisions, enumerated in the acceptance criteria checklist under our recommendation below.
+
+## Significant comments and questions
+
+### Axis ordering: the performance discussion and a SHOULD recommendation
+
+We disagree with part of the analysis in the Performance section. Axis order determines memory layout, and memory access patterns and cache locality make it important for standard use across the tooling written for `...ZYX` data. We agree that a fixed order does not need to be a MUST; real-time acquisition systems, for example, should be able to write in the acquired memory layout without transformation. However, the argument that Zarr chunk shapes make axis ordering unimportant addresses reads from storage, not the in-memory layout that libraries and viewers work with. We request that the specification include a statement that, if present, the `t`, `c`, `z`, `y`, `x` axes SHOULD occur in that order.
+
+### Axis types: encourage use and provide naming guidance
+
+The 0.6rc0 axes metadata already says that an axis SHOULD contain the field `type`, with a list of well-known values and a MAY for custom strings. RFC-3 keeps that sentence but does not restate the `type` recommendation in its own rules 1 to 3 (the Proposal paragraph does), and it gives no guidance on how custom types should be named, so `lifetime`, `decay`, and `flim` can all appear for the same axis. We ask that the rules carry the `type` recommendation alongside the names, and that the RFC provide guidance, with examples, on how and what to name custom types.
+
+### Frequency axis naming collides with the uniqueness rule
+
+Rule 4 recommends `w`, `v`, `u` for the spatial frequencies of `z`, `y`, `x`, and `w` or `ω` for the temporal frequency of `t`. Rule 5 then requires that axis names MUST NOT be repeated. A `t, z, y, x` time lapse taken into the Fourier domain and named per rule 4 becomes `w, w, v, u`, which rule 5 forbids. The recommended temporal frequency name should be distinct from the spatial ones; which name is up to the authors.
+
+### Normative `SHOULD` in rule 3
+
+Rule 3 reads "it should be named `t`", lowercase, whereas rules 1 and 2 use `SHOULD` and the Proposal paragraph says the time axis "SHOULD have name "t" and type "time"". Since the specification reads these key words per RFC 2119, the lowercase form is non-normative once inserted. We ask for `SHOULD`.
+
+### Scope of axis name uniqueness
+
+Rule 5 says names MUST NOT be repeated "within a dataset". The 0.6rc0 axes metadata says they "MUST be unique within the same coordinate system". A multiscales object can define several coordinate systems that naturally share axis names (the specification's own unitless example has `intrinsic` and `array` coordinate systems both using `y` and `x`), and "dataset" is also the specification's term for a resolution level. Unless a stronger requirement is intended, we ask that rule 5 say "within the same coordinate system"; if it is intended, the scope should be stated and reconciled with the existing sentence.
+
+### Rule 6 is not testable for custom axes
+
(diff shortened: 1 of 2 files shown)


thewtex:
CC @joshmoore @jni @vboussot

rfc/3/reviews/2/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Significant comments and questions
+
+### Axis ordering: the performance discussion and a SHOULD recommendation
+
+We disagree with part of the analysis in the Performance section. Axis order determines memory layout, and memory access patterns and cache locality make it important for standard use across the tooling written for `...ZYX` data. We agree that a fixed order does not need to be a MUST; real-time acquisition systems, for example, should be able to write in the acquired memory layout without transformation. However, the argument that Zarr chunk shapes make axis ordering unimportant addresses reads from storage, not the in-memory layout that libraries and viewers work with. We request that the specification include a statement that, if present, the `t`, `c`, `z`, `y`, `x` axes SHOULD occur in that order.
```
d-v-b, line 26: in the context of Zarr, logical axis order does not determine memory layout of decoded chunks. The Zarr V3 spec very deliberately treats the memory layout of decoded chunks as an implementation concern. That means two different implementations can decode the same Zarr chunks into different memory layouts. 

Also, codecs like [`transpose`](https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/codecs/transpose/index.rst) may change the logical order of array axes without affecting the memory layout.

And even if the Zarr spec did guarantee an in-memory layout for decoded chunks, time-based analysis benefits from different data locality from space-based analysis. So there is simply no generally performant layout that works for all analysis patterns.

Thus I would not recommend any particular order with a SHOULD statement, but the RFC could devote some language to explaining the points above, and we should explain how data consumers can signal to users which layouts are efficient and which are not. But I believe this will always be tool-specific.

rfc/3/reviews/2/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Forward Compatibility is out of date
+
+The section still describes coordinate transformations as "a draft proposal" (linking [ome/ngff#138](https://github.com/ome/ngff/pull/138)) that "would need to be updated" after RFC-3, while the Background of the same RFC states that 0.6 incorporates RFC-5. We ask that the section describe the 0.6rc0 state and name what remains to reconcile, namely the vector-axis placement wording and the rotation matrix enumeration, without RFC-3 having to resolve them.
+
+### Dataset conformance versus partial implementations
+
+The Testing section says that implementations "may check their compliance with this RFC using these datasets" and that "software is considered compliant if it provides an informative error message". These are two different things: a 6D dataset with two time axes is conformant, while a viewer that rejects it with a clear message is a partial implementation, not an implementation of the RFC. We ask for distinct terms (conformant datasets, supported features), with the informative error stated as expected behaviour for unsupported features rather than as the definition of compliance.
```
d-v-b, line 58: I suspect every numerical library with support for N-dimensional arrays will set an upper limit on the number of supported array dimensions. NumPy's current limit is [64](https://numpy.org/devdocs/reference/c-api/array.html#c.NPY_MAXDIMS). So technically this makes any ome-zarr package relying on numpy a partial implementation. 

If we don't want to label nearly all implementations as partial, maybe the RFC should define a recommended upper bound on the number of dimensions, based on the limits of known Zarr implementations.

d-v-b:
this is great! i had 2 inline comments about logical axis order and dimension limits

rfc/3/reviews/2/index.md
thewtex: This belongs in a different Comment or Review.

thewtex:
> i had 2 inline comments about logical axis order and dimension limits

@d-v-b please put these in your Comment or Review.

rfc/3/reviews/2/index.md
thewtex: We are referring to the default transpose and memory layout here. We are also referencing the most common analysis pattern, not all analysis patterns. We very strongly think a SHOULD statement is appropriate.

lubianat (approved):

rfc/3/reviews/2/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Frequency axis naming collides with the uniqueness rule
+
+Rule 4 recommends `w`, `v`, `u` for the spatial frequencies of `z`, `y`, `x`, and `w` or `ω` for the temporal frequency of `t`. Rule 5 then requires that axis names MUST NOT be repeated. A `t, z, y, x` time lapse taken into the Fourier domain and named per rule 4 becomes `w, w, v, u`, which rule 5 forbids. The recommended temporal frequency name should be distinct from the spatial ones; which name is up to the authors.
+
+### Normative `SHOULD` in rule 3
+
+Rule 3 reads "it should be named `t`", lowercase, whereas rules 1 and 2 use `SHOULD` and the Proposal paragraph says the time axis "SHOULD have name "t" and type "time"". Since the specification reads these key words per RFC 2119, the lowercase form is non-normative once inserted. We ask for `SHOULD`.
```
jni, line 38: Sorry can you clarify this? I don't understand. Are you saying that the recommendation should specify a *lowercase* "t" explicitly?

rfc/3/reviews/2/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Rule 6 is not testable for custom axes
+
+Rule 6 requires that "the order of the axes MUST match their ordering within the data if applicable". For `coordinate` and `displacement` axes this matches what RFC-5 already defines ("the `i`th value of the array along the `coordinate` or `displacement` axis refers to the `i`th output axis"). For the other axes the RFC motivates, such as a length-6 diffusion tensor axis, Euler angles, or quaternions, nothing in the axes metadata defines what index 0 is, so there is no ordering to match and the MUST cannot be checked. We are not asking RFC-3 to standardize tensor layouts; we ask that rule 6 be scoped to axes whose component semantics the specification defines, or reworded as guidance.
```
jni, line 46: I feel like "if applicable" is already doing the work requested here...

I disagree about diffusion tensors: if the image spatial axes have order zyx, then my diffusion tensor MUST look like:

```
[[Dzz, Dzy, Dzx],
 [Dyz, Dyy, Dyx],
 [Dxz, Dxy, Dxx]]
```

and MUST NOT look like:

```
[[Dxx, Dxy, Dxz],
 [Dyx, Dyy, Dyz],
 [Dzx, Dzy, Dzz]]
```

and vice-versa. There's multiple ways to encode the symmetric tensor as a 1D vector but my expectation is that the first and last element should be constrained by the language.

rfc/3/reviews/2/index.md
```diff
@@ -0,0 +1,100 @@
...
+
+### The list of specification changes is not exhaustive
+
+The Proposal states that "No further changes to the specification document are proposed by this RFC." The 0.6rc0 hierarchy diagram still says "All image arrays must be up to 5-dimensional with the axis of type time before type channel, before spatial axes"; the `coordinates` and `displacements` constraints still place the vector axis "after a time axis (if present) and before the spatial axes", which relies on the ordering RFC-3 removes; and the schemas keep the same limits (`axes.schema` `maxItems: 5` and two or three `space` axes, `mapAxis` and `projectAxis` indices bounded to 0..4, `rotation` matrices 2x2 to 5x5 only). The 0.9.dev1 draft ([ome/ngff-spec#190](https://github.com/ome/ngff-spec/pull/190)) had to change all of these and flags the `rotation` enumeration as "incompatible with RFC-3 but cannot be easily extended". We ask that the list be made exhaustive, text and schemas, or that the sentence be replaced by one that says consequential edits are expected and names the known ones.
```
jni, line 50: Noted. This oversight is an artifact of the RFC being originally authored around the time 0.5 was released.

rfc/3/reviews/2/index.md
vboussot: Ah sorry, badly worded on my side! Nothing about `t`, which is perfect. I meant the word "should": rules 1 and 2 use the RFC 2119 `SHOULD`, rule 3 spells it lowercase, so in the spec it would read as plain prose instead of a recommendation. Capitalizing it is the whole ask.

rfc/3/reviews/2/index.md
vboussot: Fair point on "if applicable", it does cover things like Euler angles or quaternions. What we're really asking is for the rule to say which axis types it applies to; today that's `coordinate` and `displacement`, where RFC-5 defines the mapping.

For the tensor as a matrix, your example is clear, agreed. It's the flattened single-axis form that worries me: first and last element don't pin it down. `Dzz, Dzy, Dzx, Dyy, Dyx, Dxx` and `Dzz, Dzy, Dyy, Dzx, Dyx, Dxx` both start with `Dzz`, end with `Dxx`, and follow `zyx`, yet they're different (C-style row-major vs LAPACK packed storage). So two writers can both satisfy the MUST and still disagree. Just naming the applicable types would fix it, no need for RFC-3 to standardize tensor packing.

d-v-b: memory order vs dimension order has [come up already](https://github.com/ome/ngff/pull/597#discussion_r3886011001), it would be helpful to get an explanation of what "ordering within the data" is intended to mean, as in this sentence: 

> "the order of the axes MUST match their ordering within the data if applicable"

There's the expectation (and need) that the list of axes declared in ome zarr metadata match _something_. Zarr V3 does not guarantee memory-order stability across implementations. So you can't require that ND arrays be decoded to C or F ordered memory, or even that they be decoded to contiguous memory at all, and so you can't require that any logical array element appear in any particular order in the in-memory representation of a decoded chunk. Implementations just have to agree on the _logical_ order of dimensions, and that's given by the `dimension_names` field. This is a subtle point, and I get the feeling that some folks in the conversation are working with the assumption that there's a "default memory order". Maybe ome-zarr can impose one? but it would be an uphill battle against a core feature of the zarr v3 spec.

joshmoore:
Thanks for the clarifying comments. Merging in expectation of @jni's response.

PR #599, opened 2026-08-31
RFC-4: give the orientation slot its range directly

vboussot:
`orientation` on `SpaceAxis` carries a one-element `any_of` and no `range`, so the generator emits the default string range beside the reference:

```json
"anyOf": [{"$ref": "#/$defs/AnatomicalOrientation"}, {"type": "null"}],
"type": "string"
```

Both constraints apply, so this branch of the schema rejects every oriented axis and `null` alike; a document only validates through the generic `Axis` branch of the axes `anyOf`, which checks less than `SpaceAxis` means to.

Stating the range directly removes the stray `type`, keeps the `null` branch the description promises, and lets `orientation.ts` type the attribute `AnatomicalOrientation` rather than `string`. Regenerated with `pixi run gen-all`; the markdown keeps the `orphan: true` front matter from 0d68c8b.

Independent of #595, which touches the same generated files; whichever merges second regenerates cleanly.


thewtex (approved):
👍

jo-mueller (approved):

PR #605, opened 2026-09-07
Update RFC-5 status

kevinyamauchi:
This PR updates the status of RFC-5 in the status table to "Implementation". An alternative would be "S3 update implementations), which is closer to the language in the RFC-1 flow chart, but I think "Implementation" more closely matches the language in the rest of the table.

Also, I changed the status in the RFC page to "S3" as I think "S4" is for when the first two implementations are complete and the RFC is adopted (see the [RFC-1 flowchart](https://ngff.openmicroscopy.org/rfc/1/index.html#implementation)).

cc @jo-mueller

### DIFF
--- a/rfc/5/index.md
+++ b/rfc/5/index.md
@@ -14,7 +14,7 @@ Add named coordinate systems and expand and clarify coordinate transformations. 
 
 ## Status
 
-This RFC is currently in RFC state `S4` (Update implementations).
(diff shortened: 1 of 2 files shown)


jo-mueller (approved):

lubianat (approved):

lubianat:
I'll merge it, it looks good. Thanks @kevinyamauchi

PR #607, opened 2026-09-09
Fix example/schema links on versioned spec pages pointing at dev

normanrz:
## Problem

On https://ngff.openmicroscopy.org/specifications/0.5/examples/index.html, every example link points at the **dev** spec instead of 0.5. Same for the schemas index:

| page | link |
|---|---|
| `0.5/examples/index.html` | `../../dev/examples/multiscales_strict/multiscales_strict.html#examples-multiscales-strict-multiscales-example` |
| `0.5/schemas/index.html` | `../../dev/schemas/image.html#schemas-image` |

This affects 0.1–0.5; only `dev` was correct, by accident.

## Cause

Each version submodule's `pre_build.py` generates its examples/schemas pages with *identical* MyST label names — `(examples:multiscales_strict:multiscales-example)=`, `(schemas:image)=`, and so on. Sphinx reads all six versions into a single project, so those labels collide. Sphinx keeps one instance, drops the rest (1500+ `duplicate label` warnings in the build log), and every `[…](#examples:…)` link in every version then resolves to the surviving one — `dev`.

The hand-written spec prose doesn't have this problem because its labels are already namespaced: `version0.5:intro`, `version0.5:multiscale-md`. The generated ones weren't.

## Fix

A `namespace_generated_labels()` pass in `conf.py` that runs right after `pre_build.py` and rewrites the generated markdown to prefix those targets and links with `version{version}:`, matching the existing convention.

It lives in the superproject for the same reason the Bikeshed boilerplate injection just above it does — otherwise this would mean editing, committing, and bumping all six `ngff-spec` version branches. The substitution is idempotent (an already-prefixed label no longer matches), so re-running the build is safe.

## Verification

Full local Sphinx build of all six versions:

| | before | after |
|---|---|---|
| 0.5 examples | `../../dev/examples/multiscales_strict/…#examples-…` | `multiscales_strict/multiscales_strict.html#version0-5-examples-…` |
| 0.5 schemas | `../../dev/schemas/image.html#schemas-image` | `image.html#version0-5-schemas-image` |

- The anchors exist on the target pages (`id="version0-5-examples-multiscales-strict-multiscales-example"`, `id="versiondev-schemas-image"`).
- `dev`'s own links still resolve to `dev`.
- All `duplicate label` warnings for `examples:`/`schemas:` are gone; one unrelated one remains (`rfcs:rfc9:comment2` in `rfc/9/comments/`).
- Build exits 0.

## Notes for the reviewer

- Anchor slugs change (`#examples-…` → `#version0-5-examples-…`). Any external deep link to a generated example/schema anchor will break. Given those anchors currently all land on the wrong version anyway, that seemed like the right trade, but worth a second opinion.
- An alternative would be to namespace the labels in `pre_build.py` inside `ngff-spec` itself, which is arguably the more correct home — at the cost of a change on all six version branches plus six submodule bumps. Happy to go that route instead if you'd prefer.
- Two pre-existing issues I did **not** touch:
  - `href="#schemas/image.html"` in every schemas index is separately broken — `pre_build.py` writes the HTML link as `schemas/image.html`, but the index already lives in `schemas/`, so it should be `image.html`. That fix belongs in the `ngff-spec` submodule.
  - The duplicate `rfcs:rfc9:comment2` label.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lubianat:
Thanks @normanrz ! It is a good change, I think, but the claude note is somewhat long, would you mind providing a tl;dr at least for triage?

normanrz:
tl;dr
If you go to https://ngff.openmicroscopy.org/specifications/0.5/examples/index.html, all the links point to the `dev` version examples instead of the `0.5` examples. This PR fixes that.

lubianat (approved):

jo-mueller:
Looking at the comment written by claude:

> This is done here in the superproject so we do not have to edit, commit, and bump every ngff-spec version submodule.

And:

> An alternative would be to namespace the labels in pre_build.py inside ngff-spec itself, which is arguably the more correct home — at the cost of a change on all six version branches plus six submodule bumps. Happy to go that route instead if you'd prefer.

I think this would actually be the cleaner option, because that's where the error originates. Each version branch already has a `_version.py` with the correct version declared, so we'd just need to import that in each instance of `pre_build.py` and prepend autogenerated anchors with it.


This needs to be done for each submodule, so:

- [ ] 0.1 https://github.com/ome/ngff-spec/pull/210
- [ ] 0.2 https://github.com/ome/ngff-spec/pull/209
- [ ] 0.3 https://github.com/ome/ngff-spec/pull/208
- [ ] 0.4 https://github.com/ome/ngff-spec/pull/207
- [ ] 0.5 https://github.com/ome/ngff-spec/pull/206
- [ ] dev (ideally before tagging 0.6): https://github.com/ome/ngff-spec/pull/205

Edit: Ok, that should resolve it more cleanly. I'd keep this PR open until all of the above-listed PRs are merged and the submodules bumped so we don't forget about it, but then this PR can be closed.

PR #614, opened 2026-09-16
RFC-3 response to review 2

jni:
This adds a response to review 2 of RFC-3 and makes most of the recommended changes to the RFC. Justification is provided where a change was not implemented.


rfc/3/responses/1/index.md
```diff
@@ -0,0 +1,102 @@
...
+### Minor changes
+
+- [ ] Add a statement that, if present, `t`, `c`, `z`, `y`, `x` axes SHOULD
+      occur in that order.
+      - We disagree with this recommendation, as one of the key motivations of
+        RFC is freedom to reorder these axes when acquiring data, or when
+        performing different analyses that may require different access
+        patterns. While the recommended ordering is optimal in many scenarios,
+        we continue to stress that documentation is the right place for this
+        recommendation, not the NGFF specification.
```
d-v-b, line 23: maybe also worth mentioning that in zarr v3, there is no fixed relationship between axis order and memory layout. so even recommending a particular axis order would not achieve what the authors want (a consistent memory layout)

jo-mueller: I see that request more informed by coordinate transformations. In many viewer tools, passing an array along with an affine transformation matrix to a viewer will almost certainly require stripping out the channel dimension from the affine matrix. This affine-wrestling (especially when transforming from coordinate systems with channel dimension into a coordinate system without channel dimension) is tricky. Having the channel always be the lead dimension (for instance) would simplify things a bit.

That's just to say the affine wrangling is annoying, but entirely possible and maybe the solution would rather be to not have a coordinate vector concern itself with transforms at all (which would be the way we are discussing over at #580).

d-v-b: then we have very different reads on what the authors were concerned about 😆 I thought they wanted to ensure that the default in-memory layout of decoded chunks was amenable to spatial algorithms like smoothing by packing spatial dimensions close together in decoded memory. But this is not something you can declare in zarr metadata.

jni: > maybe also worth mentioning that in zarr v3, there is no fixed relationship between axis order and memory layout. so even recommending a particular axis order would not achieve what the authors want (a consistent memory layout)

Are you referring to the RFC authors or the review authors? :sweat_smile: 

Also, I trust your statement here but I don't fully grok it. Can you construct me a pathological example with code in which the axis ordering is tczyx but the data is contiguous in t? A concrete example would help the conversation, I think.

Either way, this seems like it is a broader discussion that should not (imho) hold up the RFC. It can always be added later.

d-v-b: > Also, I trust your statement here but I don't fully grok it. Can you construct me a pathological example with code in which the axis ordering is tczyx but the data is contiguous in t? A concrete example would help the conversation, I think.

this gist illustrates part of the point: https://gist.github.com/d-v-b/f53c0e3b6fb1fe2b1314500cc969f0e8

it shows that `zarr`, the python lib, can decode the exact same chunk bytes into C or F contiguous memory, depending on a runtime configuration flag. This shows that memory order of decoded chunks is an implementation degree of freedom.

and C / F order is just one point in a bigger space of memory layouts that are possible. I would like to support decoding into sparse arrays, which don't have a contiguous memory layout at all.

this can be a hard part of the spec to internalize (I know because it took me a while to have the "a ha" moment), because numpy-style "arrays ARE contiguous sequences of bytes in memory" thinking is so sticky. But for the Zarr format, the memory layout of decoded arrays is outside the spec. Implementations can use C, F, morton, sparse, unordered, whatever. So OME-Zarr can't make any assumptions or statements about the in-memory representation about arrays, unless you want to be stricter than the language in the Zarr spec.

lubianat: I think the authors have a valid concern, but I'd vote for keeping the SHOULD. The wording still allows other ordering, just requires that decision to not be taken without thinking about it. From [rfc2119](https://www.rfc-editor.org/info/rfc2119/): 

> SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

I am also hesitant to write SHOULDs in the documentation elsewhere without a clear reference point in the spec

lubianat (commented):
It is well formatted. If authors want to amend it in anyway, that is okay (the conversation is surely healthy).

PR #621, opened 2026-09-18
get version history to 0.6 submodule

jo-mueller:
Title. The version history currently comes from the 0.6dev branch, which I think is not the right place. The latest version history that's in sync with the 0.6 release comes from the 0.6 submodule.


lubianat (approved):
approving; it seems like both options are valid, though, so I am equally fine with either.

joshmoore (approved):

PR #622, opened 2026-09-18
Fix link to https://ome.github.io/ome2024-ngff-challenge/

will-moore:
Link fix to https://ome.github.io/ome2024-ngff-challenge/


lubianat (approved):
thanks for the catch!

